PR #1, opened 2019-01-26
Rewrite with_module_fallback without format macro

killercup:
Also swaps the if/else branches to reduce indentation (and make the
possibly hotter non-unknown case ore obvious).

### DIFF
--- a/src/collapse-perf.rs
+++ b/src/collapse-perf.rs
@@ -325,23 +325,34 @@ impl PerfState {
 // massage function name to be nicer
 // NOTE: ignoring https://github.com/jvm-profiling-tools/perf-map-agent/pull/35
 fn with_module_fallback(module: &str, rawfunc: &str, pc: &str, include_addrs: bool) -> String {
-    if rawfunc == "[unknown]" {
-        // try to use part of module name as function if unknown
-        let rawfunc = if module != "[unknown]" {
-            // use everything following last / of module as function name
-            &module[module.rfind('/').map(|i| i + 1).unwrap_or(0)..]
-        } else {
-            "unknown"
-        };
-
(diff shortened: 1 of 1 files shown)


killercup:
I think this was the last thing you talked about before we distracted you with other stuff :)

src/collapse-perf.rs
```diff
@@ -325,23 +325,34 @@ impl PerfState {
...
-        } else {
-            format!("[{}]", rawfunc)
-        }
+    // try to use part of module name as function if unknown
+    let rawfunc = if module != "[unknown]" {
+        // use everything following last / of module as function name
+        &module[module.rfind('/').map(|i| i + 1).unwrap_or(0)..]
+    } else {
+        "unknown"
+    };
```
jonhoo: How about:
How about, above:
```
if module != "[unknown]" {
    // ...
} else if !include_addrs {
    return rawfunc.to_string();
} else {
    "unknown"
}
```
to save us the string construction below for the `rawfunc == module == "[unknown]"` and `!include_addrs` common-case? Might not matter in terms of performance though...

killercup: Good idea! I might want to write it like this, though :)

```rust
let rawfunc = match (module, include_addrs) {
    ("[unknown]", true) => "unknown",
    ("[unknown]", false) => {
        // no need to process this further
        return rawfunc.to_string();
    }
    (module, _) => {
        // use everything following last / of module as function name
        &module[module.rfind('/').map(|i| i + 1).unwrap_or(0)..]
    }
};
```

jonhoo: I like that! Either order is fine I think.

jonhoo:
Thanks!

PR #6, opened 2019-01-27
Add regression tests

Byron-TW:
By running ./test.sh, one can now verify we produce the same results
as the upstream implementation.
It's still to be determined if this is what we want, but it's a good
way to know where we are at in terms of conformity and correctness.


jonhoo:
Ultimately, I'd prefer to see this as an integration test so that it's run with `cargo test`. You can probably do that just by writing a single test in that file that loops over all the files in the directory and calls `handle_file` on each one. You might have to modify `handle_file` slightly so that it also takes a `Write` though. Otherwise you wouldn't have a good way of looking at `inferno`'s output.

Another thing worth taking into account here is the fact that we diverge from `stackcollapse-perf` in our [treatment of `"` and `'`](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/stackcollapse-perf.pl#L301). Specifically, we don't strip those characters, whereas `stackcollapse-perf` does. So, the test, however it's written, will probably have to remove those character's from `inferno-collapse-perf`'s output before comparing it to the expected output.

Finally, I think I'd prefer to see `FlameGraph` just added directly as a git submodule instead of bringing in all of the `test/` files directly into this repository. What are your thoughts on that? That would also factor into #7, which could then use `stackcollapse-perf` directly from the submodule's directory.

jonhoo:
I think this will be superseded by #11, but thanks anyway!

PR #7, opened 2019-01-27
Add 'compare.sh' to pitch one against the other

Byron-TW:
Using hyperfine, it's easy to measure the 10x speedup, repeatedly, and maybe squeeze out some more one day :).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 **/*.rs.bk
 /perf*
 *.log
+/FlameGraph/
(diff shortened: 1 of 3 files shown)


jonhoo:
I wonder if it'd be a good idea to run it on more than one input file to measure things more reliably? Also, I would _love_ to see us use [`criterion`](https://github.com/bheisler/criterion.rs) for these benchmarks, as that would also give us performance results with regression tests over time! I'd be okay with merging without that though if you think that's too much work.

Also, I assume this requires #6?

Byron-TW:
It didn't cost me much time to set this up, but I thought it's a start already. Criterion would certainly be a way to go, too, but apparently wouldn't be used when comparing the new with the old implementation. The 10x speedup is probably just a start. 

#6 should not be required, even though both edit the README, which might conflict.

jonhoo:
I think this is a good thing to merge, though I think we should wait until #11 lands so that we can take advantage of the submodule it introduces.

Issue #2, opened 2019-01-27
Add test cases from FlameGraph


jonhoo:
The upstream FlameGraph repository has a number of [test cases](https://github.com/brendangregg/FlameGraph/tree/master/test) run by [this script](https://github.com/brendangregg/FlameGraph/blob/master/test.sh). It'd be good to integrate those as a test suite into this tool to make sure we retain compatibility with the original `stackcollapse` tool.

jonhoo:
Implemented by https://github.com/jonhoo/inferno/pull/11

PR #11, opened 2019-01-29
Add tests that use the FlameGraph test files

jasonrhansen:
The main functionality of collapse-perf has been split out into a library to
allow the tests to call handle_file() directly and get a result without
creating temporary files.

Flamegraph has been added as a git submodule so we can use their test files in
flamegraph/test. Not all tests are passing so those still need to be looked
into.


jasonrhansen:
#2

tests/collapse-perf-tests.rs
```diff
@@ -0,0 +1,222 @@
...
+    let mut expected = BufReader::new(File::open(expected_file)?);
+
+    result.set_position(0);
+
+    let mut buf = String::new();
+    for line in result.lines() {
+        // Strip out " and ' since perl version does.
+        let line = line?.replace("\"", "").replace("'", "");
+        expected.read_line(&mut buf)?;
+        assert_eq!(line, buf.trim_end());
```
jonhoo: It'd be awesome if this could also include the line number. I guess the file is visible from the test name?

jonhoo: Also, I wonder if we might want to use something like [`pretty_assertions`](https://github.com/colin-kiegel/rust-pretty-assertions) to make these things easier to debug.

Cargo.toml
```diff
@@ -22,6 +22,10 @@ structopt = "0.2"
 [profile.release]
 debug = true
 
+[lib]
+name = "inferno_collapse_perf"
```
jonhoo: Given that we'll also expand this repository to have other programs, I wonder if I'd rather use nested modules here:
```rust
inferno::collapse::perf
```

src/collapse-perf.rs
```diff
@@ -18,7 +17,7 @@ const TIDY_JAVA: bool = true;
         perf script -F comm,pid,tid,cpu,time,event,ip,sym,dso,trace
     If you save this output add --header on Linux >= 3.14 to include perf info."
 )]
-struct Opt {
+pub struct Opt {
```
jonhoo: Why does this need to be `pub`?

jonhoo:
This is so great! See the inline comments I've left. Can you give me some insight into which tests/patterns are failing?

src/collapse-perf.rs
jasonrhansen: It doesn't. I think I initially moved it to lib.rs and mad it public, but forgot to remove pub when I moved it back.

tests/collapse-perf-tests.rs
jasonrhansen: I agree with both suggestions. pretty_assertions looks awesome!

Cargo.toml
jasonrhansen: Agreed.

jasonrhansen:
5 more tests pass with 07c27d7.
67 tests pass, 11 fail.

jonhoo:
Thanks! I'll start integrating this with CI and do some clean-up, but this is a superb starting point!

jonhoo:
I believe I've fixed the remaining tests now. CI is running it [here](https://travis-ci.org/jonhoo/inferno/builds/485750104) as we speak!

PR #12, opened 2019-01-29
Fail test if result file has more lines than test output

jasonrhansen:


jonhoo:
Ah, good catch. I think we may also want to assert the opposite too though -- that our output isn't _longer_ than the expected output.

jasonrhansen:
If the output is longer than the expected output, the test will at least fail as it is. I'll go ahead and make it more explicit.

PR #13, opened 2019-01-29
Support Java inlining

jasonrhansen:


src/collapse/perf.rs
```diff
@@ -253,38 +253,48 @@ impl PerfState {
...
-            //
-            //     7f722d142778 Ljava/io/PrintStream;::print (/tmp/perf-19982.map)
-            if self.opt.annotate_kernel
-                && (module.starts_with('[') || module.ends_with("vmlinux"))
-                && module != "[unknown]"
-            {
-                func.push_str("_[k]");
+                // Annotations
+                //
+                // detect inlined from the inline Vec
```
jonhoo: I think we should be a bit more specific here. Perhaps even link to https://github.com/brendangregg/FlameGraph/pull/89 for context.

src/collapse/perf.rs
```diff
@@ -253,38 +253,48 @@ impl PerfState {
...
+                // detect kernel from the module name; eg, frames to parse include:
+                //
+                //     ffffffff8103ce3b native_safe_halt ([kernel.kallsyms])
+                //     8c3453 tcp_sendmsg (/lib/modules/4.3.0-rc1-virtual/build/vmlinux)
+                //     7d8 ipv4_conntrack_local+0x7f8f80b8 ([nf_conntrack_ipv4])
+                //
+                // detect jit from the module name; eg:
+                //
+                //     7f722d142778 Ljava/io/PrintStream;::print (/tmp/perf-19982.map)
+                if inline.len() > 0 {
```
```suggestion
                if !inline.is_empty() {
```

jonhoo:
I really wish we had a test-case for this... I'll ping https://github.com/brendangregg/FlameGraph/pull/89 and see if we may be able to dig one up!

src/collapse/perf.rs
```diff
@@ -253,38 +254,54 @@ impl PerfState {
...
+                    && module.ends_with(".map")
+                {
+                    func.push_str("_[j]"); // jitted
+                }
+
+                java_inline.push(func);
             }
 
-            self.stack.push_front(func);
+            for func in java_inline.into_iter().rev() {
```
jonhoo: How do you feel about making this:
```rust
while let Some(func) = java_inline.pop() {
```
I feel like it reads a little nicer?

Issue #5, opened 2019-01-27
Add support for un-inlining.


jonhoo:
`stackcollapse-perf` supports [calling out to `addr2line`](https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-perf.pl#L123L153) to determine the symbol to use for each program counter address ([as opposed](https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-perf.pl#L264L267) to using the symbol names that `perf script` produces). We should probably take advantage of the [`addr2line` crate](https://crates.io/crates/addr2line), which enables us to do un-inlining directly without executing an external process, which should significantly speed up processing. In particular because we can parse the debug symbols only once!

jonhoo:
Fixed in #14!

PR #14, opened 2019-02-01
Add un-inline support using addr2line

jasonrhansen:
This commit adds support for #5, and adds command line flags --inline and --context. 

The problem is finding a good way to test it.  The perl version has no tests for it. They have a comment that says "these are tricky since they use addr2line, whose output will vary based on the test
system's binaries and symbol tables." I agree that this makes it difficult. Any ideas on how to approach this would be helpful.


src/collapse/perf.rs
```diff
@@ -412,3 +560,129 @@ fn tidy_java(mut func: String) -> String {
...
+        let rest = &module[perf + 5..];
+        if let Some(map) = rest.find(".map") {
+            if rest[..map].chars().all(|c| c.is_ascii_digit()) {
+                return false;
+            }
+        }
+    }
+
+    // /kernel\./
+    if module.find("kernel.").is_some() {
```
jonhoo: `.contains()`?

src/collapse/perf.rs
```diff
@@ -44,11 +95,11 @@ pub fn handle_file<R: BufRead, W: Write>(opt: Options, mut reader: R, writer: W)
         if line.is_empty() {
             state.after_event();
         } else {
-            state.on_line(line.trim_end());
+            state.on_line(line.trim_end())?;
         }
     }
 
-    state.finish(writer)
+    state.finish(writer).map_err(|e| e.into())
```
```suggestion
    Ok(state.finish(writer)?)
```

src/collapse/perf.rs
```suggestion
    if module.contains("kernel.") {
```

src/collapse/perf.rs
```diff
@@ -412,3 +560,129 @@ fn tidy_java(mut func: String) -> String {
...
+    }
+
+    // /kernel\./
+    if module.find("kernel.").is_some() {
+        return false;
+    }
+
+    // /\[[^\]]+\]/
+    let mut start = 0;
+    while start < module.len() {
```
```suggestion
    while let Some(open) = module[start..].find('[') {
```

src/collapse/perf.rs
```diff
@@ -412,3 +560,129 @@ fn tidy_java(mut func: String) -> String {
...
+        if let Some(open) = module[start..].find('[') {
+            let rest = &module[open + 1..];
+            if let Some(close) = rest.find(']') {
+                if close > 1 && rest[..close].chars().all(|c| c != ']') {
```
jonhoo: Won't this necessarily be true since `.find` will find the _next_ occurrence of `]`?

src/collapse/perf.rs
```diff
@@ -412,3 +560,129 @@ fn tidy_java(mut func: String) -> String {
...
+            }
+        }
+    }
+
+    // /kernel\./
+    if module.find("kernel.").is_some() {
+        return false;
+    }
+
+    // /\[[^\]]+\]/
```
jonhoo: I wonder why the regex is written this way. It's _specifically_ looking for a string that _somewhere_ has a square-bracket enclosed string. There's no further requirement that the square brackets are balanced or anything. What is it getting at here? What is it trying to match against? I'd be curious to find out, because we may be able to do better than the search below.

src/collapse/perf.rs
```diff
@@ -412,3 +560,129 @@ fn tidy_java(mut func: String) -> String {
...
+    #[test]
+    fn test_is_rust_hash() {
+        assert!(is_rust_hash("h21F2A89C449"));
+        assert!(!is_rust_hash("21F2A89C449"));
+        assert!(!is_rust_hash("q21F2A89C449"));
+        assert!(!is_rust_hash("h21F2A89C44Z"));
+    }
+
+    #[test]
+    fn test_can_un_iline() {
```
jonhoo: I wish we had some real strings to use here that came from real profiles and that we know appear in the wild. That way we'd actually know what it's trying to match on...

src/collapse/perf.rs
```diff
@@ -324,6 +387,91 @@ impl PerfState {
...
+                    file.unwrap_or("??"),
+                    line.map(|l| l.to_string()).unwrap_or_else(|| "?".into())
+                );
+            }
+
+            total_length += function.len();
+            full_func.push(function);
+        }
+
+        let mut joined = String::with_capacity(total_length + full_func.len() - 1);
```
jonhoo: I think it would actually be better to _not_ join these here, and instead push them all directly onto `self.stack` in the right order. I don't think there's a good reason to join them separately here? Even though I know that's what the Perl version does.

src/collapse/perf.rs
```diff
@@ -245,12 +301,17 @@ impl PerfState {
                 }
             }
 
-            // TODO: show_inline
+            // --inline flag
+            if self.opt.show_inline && can_un_inline(module) {
+                let un_inlined = self.un_inline(pc, module)?;
```
jonhoo: I think we _probably_ don't want to exit the entire parsing just because we couldn't resolve a particular symbol. Instead, we probably just want to print an error and move on. That would also let us not propagate around this custom error value everywhere. Thoughts?

src/collapse/perf.rs
```diff
@@ -324,6 +387,91 @@ impl PerfState {
...
+            ))
+        })?;
+
+        let mut full_func = SmallVec::<[String; 1]>::new();
+        let mut total_length = 0;
+        while let Some(frame) = frames.next().unwrap() {
+            let mut function = frame
+                .function
+                .map(|func| {
+                    let name = remove_discriminator(
```
jonhoo: I think this would be easier to read as a series of `let name` statements:
```rust
let name = func.raw_name().map(Cow::from);
let name = name.unwrap_or_else(|_| Cow::from("??"));
let name = remove_discriminator(name);
let name = addr2line::demangle_auto(name, func.language).as_ref();
let name = strip_rust_hash(name).to_string();
```

src/collapse/perf.rs
```diff
@@ -324,6 +387,91 @@ impl PerfState {
...
+                        func.raw_name()
+                            .map(Cow::from)
+                            .unwrap_or_else(|_| Cow::from("??")),
+                    );
+                    strip_rust_hash(addr2line::demangle_auto(name, func.language).as_ref())
+                        .to_string()
```
jonhoo: It seems unfortunate for this code to throw away the `Cow<str>` with `.as_ref()` for `strip_rust_hash`, and then to turn it back into a `String` here, when the `Cow<str>` _could_ already be owned.

src/collapse/perf.rs
```diff
@@ -324,6 +387,91 @@ impl PerfState {
...
+                            .unwrap_or(None);
+                        (file, loc.line)
+                    }
+                    None => (None, None),
+                };
+                function = format!(
+                    "{}:{}:{}",
+                    function,
+                    file.unwrap_or("??"),
+                    line.map(|l| l.to_string()).unwrap_or_else(|| "?".into())
```
jonhoo: Why is the `to_string` necessary here?

src/collapse/perf.rs
```diff
@@ -324,6 +387,91 @@ impl PerfState {
...
+                            .as_ref()
+                            .map(|f| Path::new(f).file_name())
+                            .map(|f| f.map(OsStr::to_str))
+                            .unwrap_or(None)
+                            .unwrap_or(None);
+                        (file, loc.line)
+                    }
+                    None => (None, None),
+                };
+                function = format!(
```
jonhoo: I wonder if, instead of this `format!` with lots of `unwrap_or`s in it, we should just match on `file` and `line` and put the `?`s in the format string. Should avoid some allocations.

src/collapse/perf.rs
```diff
@@ -324,6 +387,91 @@ impl PerfState {
...
+                let loc = frame.location;
+                let (file, line) = match loc {
+                    Some(ref loc) => {
+                        let file = loc
+                            .file
+                            .as_ref()
+                            .map(|f| Path::new(f).file_name())
+                            .map(|f| f.map(OsStr::to_str))
+                            .unwrap_or(None)
+                            .unwrap_or(None);
```
jonhoo: Why all this song and dance to turn `loc.file` into a `Path`?

jonhoo:
This is a cool change, and pretty much how I expected it to turn out. Left a bunch of inline comments. Should also be far faster than `stackcollapse-perf`'s support for inlining, since it doesn't have to shell out _for every `$pc`_!

As for testing, I sadly don't have a good answer for that either. One way would be for us to include a binary in the repo that is a compiled C program that we also have profiles for. We'd just want to make sure the compiler does inline stuff in it, but then it should make for a decent, and consistent, test case. What do you think?

jonhoo:
Also, seeing all that inlining code _really_ makes me wish that `addr2line` had the code from the [`addr2line` example program](https://github.com/gimli-rs/addr2line/blob/master/examples/addr2line.rs) available as a function directly on the library (cc @fitzen @philipc @main--).

src/collapse/perf.rs
```diff
@@ -324,6 +387,91 @@ impl PerfState {
...
+                        func.raw_name()
+                            .map(Cow::from)
+                            .unwrap_or_else(|_| Cow::from("??")),
+                    );
+                    strip_rust_hash(addr2line::demangle_auto(name, func.language).as_ref())
+                        .to_string()
+                })
+                .unwrap_or_else(|| "??".into());
+
+            if self.opt.show_context {
```
jonhoo: The Perl code does some really funky stuff [around here](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/stackcollapse-perf.pl#L140L144) when `--context` is enabled. Any idea how that fits into here?

main--:
> Also, seeing all that inlining code _really_ makes me wish that `addr2line` had the code from the [`addr2line` example program](https://github.com/gimli-rs/addr2line/blob/master/examples/addr2line.rs) available as a function directly on the library

I'm not sure what you mean? From what I can see the example does only three notable things: mmap, string parsing and string printing. The mmap is is 2 lines of code plus a crate while the other two exist basically only to closely mimic binutils addr2line. It's unclear to me how this would help here.

jonhoo:
@main-- oh, sorry, I should have been a bit more specific. The `un_inline` function introduced in this PR has to:

 - open the file
 - pass it to `object::File::parse`
 - pass that to `Context::new`
 - call `Context::find_frames`
 - iterate over the frames (with an extra `unwrap` it seems)
 - tease out the function name from each frame using (somewhat simplified)
   ```rust
   addr2line::demangle_auto(frame.function.map(|func| func.raw_name()), func.language))
   ```

To me at least, that's very similar to the flow of [`examples/addr2line.rs`](https://github.com/gimli-rs/addr2line/blob/master/examples/addr2line.rs), and it seems unfortunate to have to repeat the whole boilerplate if this is a common use-case for `addr2line` (looking up function stack by address). In particular, I wonder whether `addr2line` could provide a convenience API for "create a new `Context` from a file" (maybe using `mmap` and `from_sections` internally if possible), and "iterate over the functions of an address" (including demangling the `raw_name`). Something like `Context::from_file(&File) -> Context` and `Context::resolve_iter(u64) -> impl Iterator<Item = Function>`, where `Function` is a struct that holds the demangled name, file, line number, etc.

src/collapse/perf.rs
```diff
@@ -324,6 +387,91 @@ impl PerfState {
...
+    fn un_inline(&mut self, addr: &str, module: &str) -> result::Result<String, Error> {
+        let addr = u64::from_str_radix(addr, 16).map_err(|e| {
+            Error::Parse(format!(
+                "unable to parse {} as hex address, caused by: {}",
+                addr, e
+            ))
+        })?;
+
+        if let Entry::Vacant(entry) = self.addr2line_contexts.entry(module.into()) {
+            let bytes = fs::read(module)?;
```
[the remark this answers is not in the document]
jasonrhansen: Sounds good to me. I just wanted to avoid unsafe.

src/collapse/perf.rs
jasonrhansen: The to_string is to get the type to match the unwrap_or_else so we can fall back to "?". I'm sure there's a nicer way to do this. What do you suggest?

jasonrhansen: Never mind, you answered it in your next comment.

src/collapse/perf.rs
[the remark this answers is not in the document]
jonhoo: Hmm, that's a good point, though I suspect the performance speed-up here is worth it. It's not immediately clear to me why `Mmap::map` is `unsafe` in the first place, and the docs don't specify why, but then I found https://github.com/danburkert/memmap-rs/pull/66. It seems like the concern is that the underlying file may be modified, which would cause undefined behavior :/ I still think this is the right thing to do, as we're not likely to run for very long, but perhaps document that just above the `unsafe` line?

src/collapse/perf.rs
jasonrhansen: Great idea!

src/collapse/perf.rs
jasonrhansen: I hate it too. First of all loc.file is an Option, so I have to deal with the None case. Then I want to get the file name without the full path so I do Path:::new(f).file_name(). But this returns Option<&OsStr>. And OsStr can't be formatted, so I call to_str, which also returns an Option since it might not be valid unicode. What's the simplest portable way to strip the path and leave only the file name?

src/collapse/perf.rs
jasonrhansen: Yes, I was mimicking the Perl version, but there's no reason to join here since they'll get joined later if we push them directly to the stack.

src/collapse/perf.rs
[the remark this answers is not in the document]
jasonrhansen: Agreed. It's worth it, especially for larger files.

src/collapse/perf.rs
```diff
@@ -412,3 +560,129 @@ fn tidy_java(mut func: String) -> String {
...
+            }
+        }
+    }
+    s
+}
+
+// Removes Rust hash from function name if it exists.
+// When demangling Rust names, addr2line in binutils removes the hash,
+// but the Rust addr2line crate does not. Calling this function
+// after demangling will make it match binutils.
```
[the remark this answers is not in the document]
jasonrhansen: I reported an issue to the addr2line crate.

src/collapse/perf.rs
jasonrhansen: I'm not really sure what they were going for. I wanted to at least match the Perl behavior for now, but I'd also like to find out exactly what we should be searching for. In our current test files these are the only two matches: [unknown] and [vdso]

jasonrhansen: Also [kernel.kallsyms] but that already gets matched earlier because it contains "kernel."

src/collapse/perf.rs
jonhoo: Just make sure the order stays the same!

src/collapse/perf.rs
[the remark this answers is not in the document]
jonhoo: In particular because `addr2line` only needs the debug symbols, not the rest of the binary. With `fs::read`, we'll end up reading in the entire file!

src/collapse/perf.rs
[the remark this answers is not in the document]
jonhoo: Can you link the issue here? I guess we could use `rustc_demangle` when `func.language == "rust"` or something?

src/collapse/perf.rs
jasonrhansen: I'll use real test strings taken from the FlameGraph test files.

src/collapse/perf.rs
jonhoo: Hmm.. How about:
```rust
let fname = f.and_then(|f| f.rsplitn(2, std::path::MAIN_SEPARATOR).next());
```

jonhoo:
Oh, @jasonrhansen, just making sure you saw my [comment about testing](https://github.com/jonhoo/inferno/pull/14#issuecomment-459612871).

src/collapse/perf.rs
jasonrhansen: I like that much more, but I wish there was something simpler to use in the standard library.

src/collapse/perf.rs
jasonrhansen: How do you want to handle errors as a whole for this program? The main function originally returned io::Result. I could switch back to that. Would you rather not have a custom error type for the program to be able to return other types of errors? I guess anything that's not an IO error can just be printed as you're suggesting for this.

jonhoo: Yeah, I think that's the behavior we want. Specifically, we want `collapse-perf` to try to produce output to the greatest extent that it can, and failing to resolve a particular address probably shouldn't count. Maybe we should fall back to `[unknown 0x....]` in that case? I believe the only errors that would _truly_ stop progress would be I/O errors, so yeah, I think we can stick with that.

src/collapse/perf.rs
jonhoo: Maybe file an issue? Or maybe even a PR!

src/collapse/perf.rs
jasonrhansen: Perfect, that makes sense.

jonhoo: Now, this _might_ mean that we want to keep track of warnings (instead just dumping them with `eprintln!`). Perhaps it's time to introduce [`log`](https://docs.rs/log/) into inferno?

src/collapse/perf.rs
jasonrhansen: When the -f option for addr2line is used the output alternates as follows:

FUNCTIONNAME
FILENAME:LINENO
FUNCTIONNAME
FILENAME:LINENO

When --context is enabled, the perl script uses both the FUNCTIONNAME line and the FILENAME:LINENO line. It saves the function name line in $one_item and gets the location info during the next iteration of the loop. When --context is disabled it only cares about every other line.

I believe my Rust version matches this behavior, but it loops on frames instead of lines. If --context is enabled it tries to include the filename and line number from frame.location. If --context is disabled it only includes the function name.

Does that make sense?

src/collapse/perf.rs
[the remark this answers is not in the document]
jasonrhansen: I think I'll remove strip_rust_hash and use rustc_demangle. If addr2line changes we can remove rustc_demangle.

src/collapse/perf.rs
jonhoo: Ah, I see, so we replicate this behavior through the `:`-separated `format!` we do to produce each name of the trace. That seems reasonable.

src/collapse/perf.rs
```diff
@@ -412,3 +560,129 @@ fn tidy_java(mut func: String) -> String {
...
     func
 }
+
+// Removes discriminator markers.
+// Equivalent to s/ \(discriminator \S+\)//
+fn remove_discriminator(s: Cow<str>) -> Cow<str> {
+    let disc = " (discriminator ";
+    if let Some(start) = s.find(disc) {
+        let rest = &s[start + disc.len()..];
+        if let Some(end) = rest.find(')') {
```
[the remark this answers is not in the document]
jasonrhansen: You're absolutely right. I'll fix the comment.

src/collapse/perf.rs
jasonrhansen: I'm removing strip_rust_hash so there's no longer a need to change it.

src/collapse/perf.rs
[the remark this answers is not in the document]
jonhoo: Does `rustc_demangle` work correctly on non-rust names though? I wonder whether we want to use `rustc_demangle` _only_ when `func.language` is Rust?

jonhoo:
Also, to be clear, it's fine if you make multiple commits, and then we just squash them all at the end. Instead of force-pushing all the time :)

jasonrhansen:
@jonhoo, I did see your comment about testing and I agree with the general approach. Are you suggesting we write our own C program to use as a test case? Do you have any ideas of something fairly simple we could use?

Also, sorry about the force-pushing. I'm about to push a large commit with all my changes. I probably should have broken it up into several commits.

jonhoo:
Yeah, I was thinking we'd write our own simple C program. Or a Rust program would probably also be totally fine. Something with no external dependencies like:
```rust
#[inline(always)]
fn count_to(mut n: usize) {
    while n != 0 {
        n -= 1'
    }
}

#[inline(always)]
fn double_inline() {
    count_to(1_000_000);
}

#[inline(never)]
fn not_inlined() {
    count_to(1_000_000);
}

fn main() {
    count_to(1_000_000);
    double_inline();
    not_inlined();
}
```
We'd do a `perf record --call-graph dwarf` of running that program, run `perf script`, and store the output of that somewhere in the repository (adjust the constants if the file is too large). Then for the test we'd compare `stackcollapse-perf` with us with and without `--inline` and `--context`.

jonhoo:
As for force-pushing, don't worry about it. Going forward though, individual commits are easier to review, and then we can just squash it all at the end if we want!

src/collapse/perf.rs
```diff
@@ -390,87 +357,84 @@ impl PerfState {
...
+                    let name = name.unwrap_or_else(|_| Cow::from("??"));
+                    let name = remove_discriminator(name);
+                    let name = match func.language {
+                        Some(gimli::DW_LANG_Rust) => {
+                            // Using rustc_demangle directly here since addr2line::demangle_auto
+                            // doesn't remove trailing hash value from Rust symbol names.
+                            // See https://github.com/gimli-rs/addr2line/issues/108
+                            format!("{:#}", rustc_demangle::demangle(&name))
+                        }
+                        lang => addr2line::demangle_auto(name, lang).to_string(),
```
jonhoo: Can we not keep using `Cow<str>` here?

src/collapse/perf.rs
```diff
@@ -1,4 +1,11 @@
+use addr2line;
+use memmap;
+use rustc_demangle;
```
jonhoo: I don't think you need these any more in the 2018 edition?

src/collapse/perf.rs
```diff
@@ -72,6 +84,10 @@ struct PerfState {
     /// Number of times each call stack has been seen.
     occurrences: HashMap<String, usize>,
 
+    /// Cached Contexts by module name for un_inline() since constructing them is somewhat costly.
+    addr2line_contexts:
+        HashMap<String, addr2line::Context<gimli::EndianRcSlice<gimli::RunTimeEndian>>>,
```
jonhoo: Watch https://github.com/gimli-rs/addr2line/issues/109 for whether we can simplify this

src/collapse/perf.rs
```diff
@@ -245,7 +262,20 @@ impl PerfState {
                 }
             }
 
-            // TODO: show_inline
+            // --inline flag
+            if self.opt.show_inline && can_un_inline(module) {
+                match self.un_inline(pc, module) {
```
jonhoo: How about:
```rust
if let Err(e) = self.un_inline(pc, module) {
    // ...
} else {
    return;
}
```

jonhoo: Actually, on second thought, I wonder if we should just log directly from inside `self.un_inline`, since it can probably provide more specific error messages. It also saves us from having to do an additional layer of error handling here.

src/collapse/perf.rs
```diff
@@ -324,6 +354,87 @@ impl PerfState {
...
+            // since we shouldn't run very long. The potential speed-up is worth it.
+            let map = unsafe {
+                memmap::Mmap::map(&file)
+                    .map_err(|_| format!("unable to memmap file: {}", module))?
+            };
+            let file = &object::File::parse(&*map)
+                .map_err(|_| format!("unable to parse file: {}", module))?;
+            entry.insert(addr2line::Context::new(file).map_err(|e| e.to_string())?);
+        };
+        let ctx = &self.addr2line_contexts[module];
```
jonhoo: It's a little unfortunate that we can't save the extra map lookup here due to the need to return early on error. If it weren't for that, we could just use `.entry().or_insert_with(|| {})`. I wonder if there's a way to get around that. It's like we want a `fallible_or_insert_with` :/ Even moving to a setup where we _don't_ return the error but just print it won't help, since we'll still have to return early.

jonhoo: Actually, here's a though: what if we make the map `HashMap<String, Option<Context>>`? That way, we can do:
```rust
let ctx = self.addr2line_contexts.entry(module.into()).or_insert_with(|| {
    let file = File::open(module).map_err(|e| {
        warn!("unable to open module file {} to resolve {}: {:?}", module, addr, e);
    }).ok()?;
    // ...
    addr2line::Context::new(file).map_err(|e| {
        warn!("could not parse debug symbols from module {}: {:?}", module, e);
    }).ok()
});
let ctx = if let Some(ctx) = ctx {
    ctx
} else {
    return;
};
```
This has the added advantage that we won't keep trying to resolve debug symbols for a module once we've tried to parse it once!
This

src/collapse/perf.rs
```diff
@@ -324,6 +354,87 @@ impl PerfState {
...
+                            format!("{:#}", rustc_demangle::demangle(&name))
+                        }
+                        lang => addr2line::demangle_auto(name, lang).to_string(),
+                    }
+                })
+                .unwrap_or_else(|| "??".into());
+
+            if self.opt.show_context {
+                let loc = frame.location;
+                let (file, line) = match loc {
```
jonhoo: Can we just `match` directly on `frame.location` here?

src/collapse/perf.rs
```diff
@@ -412,3 +523,85 @@ fn tidy_java(mut func: String) -> String {
...
+                ));
+            }
+        }
+    }
+    s
+}
+
+// Returns whether the module can be un-inlined.
+//
+// Equivalent to the following perl:
```
jonhoo: > *Roughly* equivalent to the following perl from `stackcollapse-perf`:

src/collapse/perf.rs
```diff
@@ -390,87 +357,84 @@ impl PerfState {
...
-                        .to_string()
+                    let name = func.raw_name().map(Cow::from);
+                    let name = name.unwrap_or_else(|_| Cow::from("??"));
+                    let name = remove_discriminator(name);
+                    let name = match func.language {
+                        Some(gimli::DW_LANG_Rust) => {
+                            // Using rustc_demangle directly here since addr2line::demangle_auto
+                            // doesn't remove trailing hash value from Rust symbol names.
+                            // See https://github.com/gimli-rs/addr2line/issues/108
+                            format!("{:#}", rustc_demangle::demangle(&name))
```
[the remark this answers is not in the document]
jasonrhansen: The Display implementation on Demangle only removes the trailing hash if alternate formatting is requested. I can't see any other way to do it with the API rustc_demangle exposes.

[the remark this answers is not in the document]
jonhoo: Ah, that's awkward... Seems there aren't any really good resolutions here then sadly :'( I suppose we'll stick with this for now. Maybe document exactly what you said above for why the `{:#}` is there?

jonhoo:
This is starting to look _really_ good @jasonrhansen! Just a few more comments on the latest changes :)

src/collapse/perf.rs
jasonrhansen: Cool, why doesn't the compiler warn about this?

jonhoo: Good question. Seems like a bug. Should probably say "unnecessary `use`". My guess is that this is leftover from the old 2015 behavior.

src/collapse/perf.rs
jasonrhansen: Looks like we can now. With the way I was getting the file name before, the compiler complained, but it's no longer an issue.

src/collapse/perf.rs
jasonrhansen: Can we do that when we need to format for the Rust case?

src/collapse/perf.rs
jasonrhansen: Yes, that's great! I really wanted to use or_insert_with, but couldn't find a nice way to make it work.

src/collapse/perf.rs
jasonrhansen: I added log and replaced all uses of `eprintln!`. I'll let you choose a logging implementation.

src/collapse/perf.rs
jonhoo: Yup -- just means that that will be a `Cow::Owned`!

src/collapse/perf.rs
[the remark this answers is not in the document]
jonhoo: Oh hey, https://github.com/gimli-rs/addr2line/issues/108 just landed, so we can get rid of this now! :D

src/collapse/perf.rs
```diff
@@ -261,16 +260,8 @@ impl PerfState {
...
-                        return;
-                    }
-                    Err(e) => {
-                        eprintln!(
-                            "Error attempting to un-inline {} in {}, caused by: {}",
-                            pc, module, e
-                        );
-                    }
+                if self.un_inline(pc, module) {
+                    return;
```
jonhoo: We should check what the Perl version does here -- whether it always returns when you have `--inline` specified (i.e., doesn't fall back to another type of resolution) or not.

src/collapse/perf.rs
```diff
@@ -353,33 +344,68 @@ impl PerfState {
...
+                        );
+                    })
+                    .ok()?;
+
+                // Using memmap is unsafe because the underlying file may be modified,
+                // which would cause undefined behavior. It's unlikely to be a problem in practice
+                // since we shouldn't run very long. The potential speed-up is worth it.
+                let map = unsafe {
+                    memmap::Mmap::map(&file)
+                        .map_err(|_| warn!("unable to memmap file: {}", module))
```
jonhoo: Let's also print the error here.

src/collapse/perf.rs
```diff
@@ -353,33 +344,68 @@ impl PerfState {
...
+                        .ok()?
+                };
+
+                let file = &object::File::parse(&*map)
+                    .map_err(|_| warn!("unable to parse file: {}", module))
```
jonhoo: And let's also print the error here.

src/collapse/perf.rs
```diff
@@ -353,33 +344,68 @@ impl PerfState {
...
-        let mut frames = ctx.find_frames(addr).map_err(|_| {
-            format!(
-                "unable to parse frames from module: {}, at address: {}",
-                module, addr
-            )
-        })?;
+        let addr = match u64::from_str_radix(addr, 16) {
+            Ok(addr) => addr,
+            Err(_) => {
+                warn!("unable to parse {} as hex address", addr);
```
jonhoo: Include the error.

src/collapse/perf.rs
```diff
@@ -353,33 +344,68 @@ impl PerfState {
...
+                return false;
+            }
+        };
+
+        let mut frames = match ctx.find_frames(addr) {
+            Ok(frames) => frames,
+            Err(_) => {
+                warn!(
```
jonhoo: Include the error.

jonhoo:
We've moved the discussion with @main-- to https://github.com/gimli-rs/addr2line/issues/110#issuecomment-459965304 so we can keep this PR to the point.

src/collapse/perf.rs
jasonrhansen: When I try this, the compiler complains. What can I do here?
```Rust
                .map(|func| {
                    let name = func.raw_name().map(Cow::from);
                    let name = name.unwrap_or_else(|_| Cow::from("??"));
                    let name = remove_discriminator(name);
                    match func.language {
                        Some(gimli::DW_LANG_Rust) => {
                            // Using rustc_demangle directly here since addr2line::demangle_auto
                            // doesn't remove trailing hash value from Rust symbol names.
                            // See https://github.com/gimli-rs/addr2line/issues/108
                            // Also the "{:#}" formatting is required to not include the
                            // trailing hash.
                            Cow::Owned(format!("{:#}", rustc_demangle::demangle(&name)))
                        }
                        lang => addr2line::demangle_auto(name, lang),
                    }
                })
                .unwrap_or_else(|| Cow::from("??"));
```

```
error[E0515]: cannot return value referencing function parameter `func`
   --> src/collapse/perf.rs:429:33
    |
417 |                     let name = func.raw_name().map(Cow::from);
    |                                ---- `func` is borrowed here
...
429 |                         lang => addr2line::demangle_auto(name, lang),
    |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function
```

jonhoo: I think it'll be easier to first simplify the code now that https://github.com/gimli-rs/addr2line/issues/108 has landed, and then try to figure it out?

jasonrhansen: Don't we need to wait for a new release of addr2line on crates.io before we can simplify that?

src/collapse/perf.rs
jasonrhansen: Yeah, the Perl version never falls back to another type of resolution, so we probably shouldn't either. The Perl version will still be different when addr2line returns an error message since the error will be included on the stack, where we log the error and nothing gets added to the stack.

src/collapse/perf.rs
jasonrhansen: I pushed a commit that points Cargo.toml at the revision of addr2line with the fix we need. Are you ok with that, or should we just wait for a new release?

Also, let me know if you're satisfied with how Cow is being used now.

jasonrhansen:
I finally started trying to test this, but I'm not sure how to proceed. I wrote the following simple program in C.

```C
inline void __attribute__((always_inline)) count_to(unsigned int n) {
    for(int i = 0; i < n; i++);
}

inline void __attribute__((always_inline)) double_inline() {
    count_to(1000000000);
}

static void __attribute__ ((noinline)) not_inlined() {
    count_to(1000000000);
}

int main() {
    count_to(1000000000);
    double_inline();
    not_inlined();
}
```
Then did the following:
```
gcc -g inline_test.c
perf record --call-graph dwarf ./a.out
perf script > perfoutput
```
The problem I'm having is the --inline option doesn't seem to work with stackcollapse-perf.pl, or our implementation.

`stackcollapse-perf.pl --inline perfoutput`:
```
a.out;??;??;?? 3767
a.out;??;??;??;?? 1930
a.out;??;??;??;??;?? 1
```
`cargo run --bin inferno-collapse-perf -- perfoutput --inline`:
```
a.out 5698
```
So then I decided to try addr2line directly and took the address from this line `55ca69b98131 not_inlined+0x18 (/home/jrhansen/projects/rust/inferno/tests/a.out)` in perfoutput to see if addr2line could tell me the file and function name.
I ran `addr2line -a -e ./a.out -i -f -s -C 55ca69b98131`
But this was the output:
```
0x000055ca69b98131
??
??:0
```
Next I ran `objdump -d a.out`. Here's the the section of the output for `not_inlined`:
```
0000000000001119 <not_inlined>:
    1119:       55                      push   %rbp
    111a:       48 89 e5                mov    %rsp,%rbp
    111d:       c7 45 f8 00 ca 9a 3b    movl   $0x3b9aca00,-0x8(%rbp)
    1124:       c7 45 fc 00 00 00 00    movl   $0x0,-0x4(%rbp)
    112b:       eb 04                   jmp    1131 <not_inlined+0x18>
    112d:       83 45 fc 01             addl   $0x1,-0x4(%rbp)
    1131:       8b 45 fc                mov    -0x4(%rbp),%eax
    1134:       39 45 f8                cmp    %eax,-0x8(%rbp)
    1137:       77 f4                   ja     112d <not_inlined+0x14>
    1139:       90                      nop
    113a:       5d                      pop    %rbp
    113b:       c3                      retq

```
Then I realized that addr2line is expecting the offset into the binary, so I ran `addr2line -a -e ./a.out -i -f -s -C 112b` and got:
```
0x000000000000112b
count_to
inline_test.c:3
not_inlined
inline_test.c:11
```
Much better! But now I don't know how to use addr2line with the addresses that are in the perf output since those aren't relative to the binary.

Any ideas?

src/collapse/perf.rs
```diff
@@ -259,7 +259,8 @@ impl PerfState {
             }
 
             // --inline flag
-            if self.opt.show_inline && self.un_inline(pc, module) {
+            if self.opt.show_inline && can_un_inline(module) {
```
jonhoo: This change doesn't seem right to me given what the commit message claims it does? Specifically, this will still fall back if we don't think we can un-inline? Is that intended behavior?

jonhoo:
So, the issue here is basically [this one](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=860394): ASLR + position-independent executables are the default, and that gives a random offset to all executables. I thought there was a way to work around this, but can't remember it off the top of my head. For testing, we could just compile with `-static -no-pie`, but we'll probably want a better solution long-term that doesn't require people to change the way they profile their apps (i.e., we don't want to require them to use something like `setarch -R` to run their programs).

jonhoo:
FWIW, I filed https://github.com/jonhoo/inferno/issues/31

src/collapse/perf.rs
jasonrhansen: Maybe the commit message just isn't clear?
This is what we're trying to match:
```
                 if ($show_inline == 1 && $mod !~ m/(perf-\d+.map|kernel\.|\[[^\]]+\])/) {
			unshift @stack, inline($pc, $mod);
			next;
		}
```
`can_inline(module)`  is meant to be roughly the same as `$mod !~ m/(perf-\d+.map|kernel\.|\[[^\]]+\])/`. So if it looks like something we can't un-inline we fall back to another resolution type. But if it looks like something we should be able to un-inline, but for whatever reason `un-inline` fails, we don't fall back.

jonhoo: Ah, yes, you're entirely right!

jasonrhansen:
Thanks for the suggestion to compile with `-static -no-pie`, @jonhoo. I'll just do that for our test binary for now. I agree we need a better long-term solution, but I want to at least be able to test the functionality we currently have.

jonhoo:
\o/ I think those two are the last two questions!

tests/data/results/perf-inline-counter-collapsed-inline+context.txt
```diff
@@ -0,0 +1,3 @@
+inline_counter;main:inline_counter.c:16;count_to:inline_counter.c:4 15
```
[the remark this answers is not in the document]
jasonrhansen: Yup

jonhoo:
\o/ Thank you so much @jasonrhansen!

PR #27, opened 2019-02-03
Discussion: should BufReader have a for_each_line() method?

alsuren:
I was watching your first stream in this series, and saw that you rejected BufReader::lines() because it would allocate a new string for each iteration of the loop.

If BufReader had a for_each_line() method (prototyped here as a plain old function) would you have picked it over lines()? If so, should I propose it for the standard library?

I've not profiled it yet (it's late, and I need to go to bed, but I saw that you'd just finished a stream, so I thought I'd strike while there are lots of people interested in the project). I also suspect that a PR is not the best place to discuss such things, so please close this PR when the discussion has run its course.

### DIFF
--- a/src/collapse/perf.rs
+++ b/src/collapse/perf.rs
@@ -26,18 +26,25 @@ pub struct Options {
     pub event_filter: Option<String>,
 }
 
-pub fn handle_file<R: BufRead, W: Write>(opt: Options, mut reader: R, writer: W) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


jonhoo:
Hmm, that's an interesting idea... I think that's probably something you could try to argue should be in the standard library, though I'm probably not the right person to evaluate the suggestion. It could very well be that the response would be that it should be done in a library instead. I think it sounds pretty reasonable though! There'll be some bikeshedding on the name I'd wager, and they'll probably ask you to change the signature to make the closure take an `&str`, but apart from that it seems pretty sensible.

Issue #20, opened 2019-02-02
Persistent color palettes


jonhoo:
In the original `flamegraph.pl`, there's an option ([`--cp`](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L147)) that enables "consistent palettes". When this option is on, the color used for every function in the current flame graph is stored in a file called `palette.map`. When the option is used while the file exists, the color palette is also read in from that file when the flame graph generator starts up, and functions get their colors from that file unless they were previously unknown.

Bonus points for compatibility with `palette.map` files generated by `flamegraph.pl`.

jonhoo:
Fixed in #32!

Issue #18, opened 2019-02-02
Add full color support


jonhoo:
Currently, inferno's flamegraph implementation just [hardcodes](https://github.com/jonhoo/inferno/blob/adf4e745dc93cdc5a23768ff5b807b8f17970585/src/flamegraph/mod.rs#L117) the color of all the frames in the flamegraph. That's makes the flamegraph significantly less useful, and also much less pretty. Instead, we should support `flamegraph.pl`'s [`--colors` option](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L143L145). It affects colors in a number of ways:

 - The background color [changes](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L225L234)
 - It colors [individual frames](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L1115) through [this serious function](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L344) (which should live in its own module)
 - Some of the color modes do extra funky color things like ["chain"](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L633L646)
 - When combined with the `--hash` flag, the [`namehash`](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L323) coloring function is used, which attempts to color similar "parts" of the stack similarly, and consistently across runs.

We should support all that! It's probably most a matter of porting the giant `color` function.

jonhoo:
Fixed in #32!

PR #32, opened 2019-02-04
Full color support

JordiChauzi:
Adds support for #20 and for most of #18.

Supports:
- `--cp` (with the `palette.map` file being compatible with the `flamegraph.pl` version)
- `--hash`
- `--colors COLORS`

The hack for [chain graphs](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L633L646) is not in this PR.
It requires (I think) to change the borrowed `input: &str` in the [frames ](https://github.com/jonhoo/inferno/blob/master/src/flamegraph/merge.rs#L77)function to an owned String, because we would need to modify some lines (the ones where we need to apply the hack). I was not so sure about making that change before any review.


jonhoo:
That's awesome! Give me a little while to walk through the patch and make comments as I go :)

src/bin/flamegraph.rs
```diff
@@ -2,18 +2,31 @@ use std::fs::File;
...
-use inferno::flamegraph::{self, Options};
+use inferno::flamegraph::{self, Options, color::Palette};
 
 #[derive(Debug, StructOpt)]
 #[structopt(name = "inferno-flamegraph", author = "")]
 struct Opt {
     /// collapsed perf output file, or STDIN if not specified
     infile: Option<String>,
+    /// set color palette. choices are: hot, mem, io, wakeup, chain, java, js, perl, red, green, blue, aqua, yellow, purple, orange
+    #[structopt(long = "colors", default_value = "hot")]
```
jonhoo: Let's also use [`possible_values`](https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.possible_values) here:
```rust
#[structopt(..., raw(possible_values = r#"&["hot","mem","io","wakeup","chain","java","js","perl","red","green","blue","aqua","yellow","purple","orange"]"#)]
```
That way the user will get a helpful error if they try to use an unrecognized value.

Also, maybe we should not list `chain` since it's not truly supported?

src/bin/flamegraph.rs
```diff
@@ -2,18 +2,31 @@ use std::fs::File;
...
+    colors: Palette,
+    /// colors are keyed by function name hash
+    #[structopt(long = "hash")]
+    hash: bool,
+    /// use consistent palette (palette.map)
```
jonhoo: I _think_ `flamegraph.pl` also allows for specifying the file name to use?

src/bin/flamegraph.rs
jonhoo: I'd also like us to add `short = "c"` since I think this is a pretty common option.

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+            "yellow" => Ok(Palette::Yellow),
+            "purple" => Ok(Palette::Purple),
+            "orange" => Ok(Palette::Orange),
+            unknown=> Err(format!("unknown color palette: {}", unknown))
+        }
+    }
+}
+
+fn namehash(name: &str) -> f32 {
+    let mut vector = 0f32;
```
jonhoo: I would just use `.0` (so `0.0`) for these and let the compiler choose the type. That way moving to something like `f64` later would be an easier change.

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+            "aqua" => Ok(Palette::Aqua),
+            "yellow" => Ok(Palette::Yellow),
+            "purple" => Ok(Palette::Purple),
+            "orange" => Ok(Palette::Orange),
+            unknown=> Err(format!("unknown color palette: {}", unknown))
+        }
+    }
+}
+
+fn namehash(name: &str) -> f32 {
```
jonhoo: Let's also bring over the comment from https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L324L326 to explain what's going on here.

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    let mut vector = 0f32;
+    let mut weight = 1f32;
+    let mut max = 1f32;
+    let mut modulo = 10;
+
+    let name = {
```
jonhoo: I'm not entirely sure how the code below matches
```perl
# if module name present, trunc to 1st char
$name =~ s/.(.*?)`//;
```
from the original perl code? A comment explaining the translation may be a good addition here.

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    };
+
+    for character in name.bytes() {
+        let i = (character % modulo) as f32;
+        vector += (i / ((modulo - 1) as f32)) * weight;
+        modulo += 1;
+        max += weight;
+        weight *= 0.70;
+
+        if modulo > 12 {
```
jonhoo: I only just now realized how _weird_ this code is. Unless I'm reading it wrong, it only every iterates over three characters? If so, couldn't we just use `name.bytes().take(3)`?

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+            break
+        }
+    }
+
+    (1f32 - vector / max)
+}
+
+fn handle_java_palette(s: &str) -> Palette {
```
jonhoo: I think I'd prefer to see these functions each in their own sub-module as a `resolve` function:
```rust
colors::java::resolve(name: &str) -> Palette
```

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    if s.ends_with("]") {
+        if let Some(ai) = s.rfind("_[") {
+            if s[ai..].len() == 4 {
+                let suffix_char = &s[ai+2..ai+3];
+                if suffix_char.starts_with("k") { return Palette::Orange }
```
jonhoo: Can't this just be `==`? In fact, maybe this sequence of `if` statements can just become a `match &s[ai+2..ai+3]`?

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    (1f32 - vector / max)
+}
+
+fn handle_java_palette(s: &str) -> Palette {
+    if s.ends_with("]") {
+        if let Some(ai) = s.rfind("_[") {
+            if s[ai..].len() == 4 {
+                let suffix_char = &s[ai+2..ai+3];
```
jonhoo: This line makes me suspect that you haven't run `rustfmt` on this code. It'd be handy if you could do that so that we don't get spurious changes as parts of later commits!

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+            return Palette::Aqua
+        }
+    } else if s.contains("::") {
+        return Palette::Yellow
+    } else if let Some(ai) = s.find("/") {
+        if (&s[ai..]).contains(".js") {
+            return Palette::Green
+        }
+    }
+
```
jonhoo: One way to avoid this gap is to make the `/` case the last one instead of the middle one, and then just have a `Palette::Red` at the very end of the function (no `else`).

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    } else if s == " " {
+        Palette::Green
+    } else if s.contains("_[k]") {
+        Palette::Orange
+    } else {
+        Palette::Red
+    }
+}
+
+fn handle_wakeup_palette(_s: &str) -> Palette {
```
jonhoo: We should probably inline this in the caller (assuming there's just one). If not, it should be in its own `mod` (though probably not its own file) for consistency with the others.

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    Palette::Aqua
+}
+
+fn handle_chain_palette(s: &str) -> Palette {
```
jonhoo: I'd cut this since we don't handle `chain` correctly anyway.

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+
+fn coefficients_for_palette(palette: &Palette, name: &str, v1: f32, v2: f32, v3: f32) ->
+    ((u8, u8, f32), (u8, u8, f32), (u8, u8, f32)) {
+    let real_palette = match palette {
+        Palette::Hot => return ((205, 50, v3), (0, 230, v1), (0, 55, v2)),
+        Palette::Mem => return ((0, 0, v3), (190, 50, v2), (0, 210, v1)),
+        Palette::Io => return ((80, 60, v1), (80, 60, v1), (190, 55, v2)),
+        Palette::Red => return ((200, 55, v1), (50, 80, v1), (50, 80, v1)),
+        Palette::Green => return ((50, 60, v1), (200, 55, v1), (50, 60, v1)),
+        Palette::Blue => return ((80, 60, v1), (80, 60, v1), (205, 60, v1)),
```
jonhoo: I think the last component here should be `(205, 50, v1)`?

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+        Palette::Aqua => return ((50, 60, v1), (165, 55, v1), (165, 55, v1)),
+        Palette::Orange => return ((190, 65, v1), (90, 65, v1), (0, 0, v1)),
+        Palette::Java => handle_java_palette(name),
+        Palette::Perl => handle_perl_palette(name),
+        Palette::Js => handle_js_palette(name),
+        Palette::Wakeup => handle_wakeup_palette(name),
+        Palette::Chain => handle_chain_palette(name),
+    };
+
+    coefficients_for_palette(&real_palette, name, v1, v2, v3)
```
jonhoo: This confused me for a second -- I think a comment describing what this is for (resolving palettes used by language palettes) would be helpful.

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+fn color_from_palette(palette: &Palette, name: &str, v1: f32, v2: f32, v3: f32) -> String {
+    let ((r_b, r_a, r_x),
+         (g_b, g_a, g_x),
+         (b_b, b_a, b_x)) = coefficients_for_palette(palette, name, v1, v2, v3);
+
+    let red = affine_transform(r_a, r_b, r_x);
+    let green = affine_transform(g_a, g_b, g_x);
+    let blue = affine_transform(b_a, b_b, b_x);
```
jonhoo: I wonder if we should just do:
```rust
macro_rules! t {
  ($a:expr, $b:expr, $x:expr) => $b + ($a as f32 * $x) as u8
}
```
And then use `t!` directly inline in `coefficients_for_palette`. That would obviate the need for the extra destructuring and values here.

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    format!("rgb({},{},{})", red, green, blue)
+}
+
+pub(super) fn color_map<'a>(palette: &Palette, hash: bool, name: &'a str, palette_map: &'a mut HashMap<String, String>) -> &'a str {
```
jonhoo: Does the palette map _need_ to hold `String`?

jonhoo: Also, can we make it hold the color directly instead of the string mapping of it (i.e., the `(r, g, b)`)

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+pub(super) fn color(palette: &Palette, hash: bool, name: &str) -> String {
+    let (v1, v2, v3) = if hash {
+        let name_hash = namehash(name);
+        let reverse_name_hash = namehash(&name.chars().rev().collect::<String>());
+
+        (name_hash, reverse_name_hash, reverse_name_hash)
+    } else {
+        (rand::random(), rand::random(), rand::random())
```
jonhoo: I wonder how expensive this is... Does it need to create a new random number generator each time? Given how much `color()` is called, would it make sense to make a `thread_rng` outside the outer loop and pass in a `&mut` to it here?

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    };
+
+    color_from_palette(palette, name, v1, v2, v3)
+}
+
+pub(super) fn get_background_colors_for(palette: &Palette) -> (&str, &str) {
```
jonhoo:
```suggestion
pub(super) fn bgcolor_for(palette: &Palette) -> (&str, &str) {
```

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    match palette {
+        Palette::Hot | Palette::Java | Palette::Js | Palette::Perl => YELLOW_GRADIENT,
+        Palette::Mem | Palette::Chain => BLUE_GRADIENT,
+        _ => GRAY_GRADIENT
+    }
+}
+
+pub(super) fn read_palette(file: &str) -> Result<HashMap<String, String>, io::Error> {
```
jonhoo:
```suggestion
pub(super) fn read_palette(file: &str) -> io::Result<HashMap<String, String>> {
```

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    if path.exists() {
+        let file = File::open(path)?;
+        let file = BufReader::new(file);
+
+        for line in file.lines() {
+            let line = line?;
+            let words = line.split("->").collect::<Vec<_>>();
+            map.insert(words[0].to_string(), words[1].to_string());
+        }
+    }
```
jonhoo: A comment about why the `else` case isn't an error would be good here.

src/flamegraph/color.rs
jonhoo: Thinking a little more, I wonder if we want to wrap the palette map into its own struct. That way we could move all of the `::load` and `.save` logic into an `impl` for that struct too. Maybe even in its own file?

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    Ok(map)
+}
+
+pub(super) fn write_palette(file: &str, palette_map: HashMap<String, String>) -> Result<(), io::Error> {
+    let mut file = OpenOptions::new().write(true).create(true).open(file)?;
+    let mut entries = palette_map.into_iter().collect::<Vec<_>>();
+    entries.sort_unstable();
```
jonhoo: Is it important that these are sorted?

jonhoo: I guess the Perl version [does sort](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L514), but I don't quite see why?

src/flamegraph/mod.rs
```diff
@@ -9,6 +9,7 @@ use std::io::prelude::*;
 
 mod merge;
 mod svg;
+pub mod color;
```
```suggestion
mod color;
pub use color::Palette;
```

src/flamegraph/merge.rs
```diff
@@ -74,7 +74,7 @@ fn flow<'a, LI, TI>(
     }
 }
 
-pub(super) fn frames<'a>(input: &'a str) -> (Vec<TimedFrame<'a>>, usize, usize) {
+pub(super) fn frames(input: &str) -> (Vec<TimedFrame>, usize, usize) {
```
jonhoo: This change seems unrelated to this PR?

jonhoo:
Phew, okay, made a pass! Overall it looks quite good :D One thing I would like to see is a test. In particular, let's include a rendered `svg` (in `tests/flamegraphs/`) from each of the `flamegraph/test/results/*-collapsed-all.txt`, and then add an integration test (`tests/flamegraph-regression.rs`) that runs `flamegraph` on each of those files and compares the output to the corresponding one in `tests/flamegraphs/`.

src/bin/flamegraph.rs
JordiChauzi: It does not: [pal_file](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L118) is initialized, and then directly used in [read_palette ](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L522)and [write_palette](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L513).
But we can add the feature as an optional argument for the `--cp` option ? Or a new option ?

src/flamegraph/color.rs
JordiChauzi: Yep, the code feels awkward currently.

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+        Palette::Red
+    }
+}
+
+fn handle_perl_palette(s: &str) -> Palette {
+    if s.contains("::") {
+        Palette::Yellow
+    } else if s.contains("Perl") || s.contains(".pl") {
+        Palette::Green
+    } else if s.ends_with("_[k]") {
```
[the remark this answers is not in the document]
JordiChauzi: These conditions are not exclusive in theory. In some weird cases it may change the color ?
I don't think it is that relevant though.

src/bin/flamegraph.rs
jonhoo: Ah, yes, looks like you're right! I think having `cp` optionally take an argument would be the best way to go, but sadly that [doesn't seem to be supported yet](https://github.com/clap-rs/clap/issues/1354), so let's stick with hard-coded `palette.map` for now!

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+        return Palette::Yellow
+    } else if let Some(ai) = s.find("/") {
+        if (&s[ai..]).contains(".js") {
+            return Palette::Green
+        }
+    }
+
+    if s.contains(":") {
+        Palette::Aqua
+    } else if s == " " {
```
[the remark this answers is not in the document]
JordiChauzi: `s = " "` will not match `s.is_empty()`, or am I missing something?

src/flamegraph/color.rs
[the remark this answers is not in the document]
jonhoo: That's a fair point. I think in the cases of overlap you would actually prefer to have kernel things still marked the "kernel color" though, so I'm okay with that change in behavior!

src/flamegraph/color.rs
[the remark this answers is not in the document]
jonhoo: Err, sorry, I meant `s.trim().is_empty()`!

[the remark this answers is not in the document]
JordiChauzi: Okay, we would match multiple white-spaces as "Green" in that case, but I don't if this is a possible input.

src/flamegraph/color.rs
JordiChauzi: If we store the RBG components, we need to recreate the `rgb(r, g, b)` String each time right?

src/flamegraph/color.rs
JordiChauzi: After reading the documentation, it seems like a good idea indeed.

src/flamegraph/color.rs
JordiChauzi: I don't know if other tools may use the `palette.map` file ?
Sorting is nice if we want to compare with the `flamegraph.pl` output.

src/flamegraph/color.rs
JordiChauzi: It may be clearer if we separate language palettes from "basic" palettes in the ` Palette` enum, with a function resolving only the language palette enum into the basic palette enum ?

src/flamegraph/color.rs
JordiChauzi: [Oops](https://github.com/jonhoo/inferno/pull/32/commits/d5000681e16a220684e72429b32d5115588c483a)

src/flamegraph/color.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/d5000681e16a220684e72429b32d5115588c483a)

src/flamegraph/color.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/d5000681e16a220684e72429b32d5115588c483a)

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+pub(super) fn color_map<'a>(palette: &Palette, hash: bool, name: &'a str, palette_map: &'a mut HashMap<String, String>) -> &'a str {
+    palette_map.entry(name.to_string()).or_insert_with(|| color(palette, hash, name))
+}
+
+pub(super) fn color(palette: &Palette, hash: bool, name: &str) -> String {
+    let (v1, v2, v3) = if hash {
+        let name_hash = namehash(name);
+        let reverse_name_hash = namehash(&name.chars().rev().collect::<String>());
```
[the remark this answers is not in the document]
JordiChauzi: The module name check seems useful only for the first call to `namehash`. Maybe simplify `name` first, and then have a simple `namehash` function that just take an `Iterator<Item=u8>` to iterate over three characters ?

src/flamegraph/color.rs
JordiChauzi: [Sorry about that](https://github.com/jonhoo/inferno/pull/32/commits/c70f667e0080b41b5fa57feab4ad34e193f8c665)

src/flamegraph/color.rs
jonhoo: I think I'd hoist the `ends_with("_[j]")` further up too, and have the `/` case be the very last one.

src/flamegraph/color.rs
jonhoo: Yeah, that's a good point. Let's keep it this way then.

jonhoo: Actually, can you put in a comment saying "We sort the palette because the Perl implementation does"?

src/flamegraph/merge.rs
jonhoo: Yes please, just to make it easier to review :)

src/flamegraph/color.rs
jonhoo: Ah, yes, I like that idea!

src/flamegraph/color.rs
jonhoo: Not entirely sure what you mean here. We (currently) need to produce one `String` for every frame we produce, pretty much no matter what. While we may be able to fix that with a clever stack-allocated buffer (#33), I don't think storing the three `u8`s in the map instead is worse? It just changes when that allocation happens (I think?). It either happens when you load the palette, or when you print it. And I think it'll be easier for us to get rid of the `String` allocation when writing out the color.

src/flamegraph/color.rs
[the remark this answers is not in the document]
jonhoo: Hmm, yeah, that could work. Maybe try it out and see if it reads any better?

jonhoo:
Oh, and it's perfectly okay for you to make many smaller commits. That makes it a little easier to review than trying to sort through one big one! `git add -p` is your friend. No worries if that's inconvenient though -- I'll make do :)

src/flamegraph/color.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/11ebe1160354954fa91ec232a0bd5cbbf702337a )

src/flamegraph/color.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/e25e449aa9208bf5b3715b5bfec33b3c5b01a031)

src/flamegraph/color.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/b0a98a8940220af42cdd3d9b2a74a506863cb179)

src/flamegraph/merge.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/2ee9ca77c198766e29ee067561699564bcce49a0)

src/flamegraph/svg.rs
```diff
@@ -44,7 +44,7 @@ where
     Ok(())
 }
 
-pub(super) fn write_prelude<W>(svg: &mut Writer<W>, imageheight: usize) -> quick_xml::Result<()>
+pub(super) fn write_prelude<W>(svg: &mut Writer<W>, imageheight: usize, bgcolor1: &str, bgcolor2: &str) -> quick_xml::Result<()>
```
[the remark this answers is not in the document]
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/d891ec2f0b9a24b6f37eade81880f1fd007071ff)
I put in `svg.rs`.

src/flamegraph/color.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/6251c0562419bbe326eb0950f5c168737fbedeac)

src/flamegraph/color.rs
```diff
@@ -17,14 +17,15 @@ const GRAY_GRADIENT: (&str, &str) = ("#f8f8f8", "#e8e8e8");
 
 #[derive(Debug, PartialEq)]
 pub enum Palette {
+    Basic(BasicPalette),
+    Multi(MultiPalette),
+}
+
+#[derive(Clone, Debug, PartialEq)]
```
jonhoo: I would also `derive(Copy)` on this.

src/flamegraph/color.rs
```diff
@@ -34,9 +35,18 @@ pub enum Palette {
     Orange,
 }
 
+#[derive(Debug, PartialEq)]
```
jonhoo: And also `derive(Copy, Clone)` on this.

src/flamegraph/color.rs
```diff
@@ -203,25 +213,27 @@ fn rgb_components_for_palette(
...
-        Palette::Purple => return (t!(190, 65, v1), t!(80, 60, v1), t!(190, 65, v1)),
-        Palette::Aqua => return (t!(50, 60, v1), t!(165, 55, v1), t!(165, 55, v1)),
-        Palette::Orange => return (t!(190, 65, v1), t!(90, 65, v1), t!(0, 0, v1)),
-        Palette::Java => handle_java_palette(name),
-        Palette::Perl => handle_perl_palette(name),
-        Palette::Js => handle_js_palette(name),
-        Palette::Wakeup => handle_wakeup_palette(name),
-        Palette::Chain => handle_chain_palette(name),
+    let basic_palette = match palette {
+        Palette::Basic(basic) => basic.to_owned(),
```
jonhoo: If you `derive(Copy)` for `BasicPalette`, you can just make this `=> basic,`

src/flamegraph/color.rs
JordiChauzi: Currently, when using `color_map` one `String` is allocated for each function name: two frames with the same function will borrow the same `String`.

src/flamegraph/color.rs
JordiChauzi: Do you prefer to add a subdirectory `colors`, and files `java.rs`, `js.rs`, `perl.rs`, etc... or submodules in the `color.rs` file ?
I find the subdirectory approach a little overkill for a single, relatively small, function per submodule.

src/flamegraph/color.rs
jonhoo: I think I'd still prefer that we don't allocate a string at all, which I think we'll be able to do with #33 :)

src/flamegraph/color.rs
jonhoo: I think what I'd do is make a file `palettes.rs`, and in `color.rs` do:
```rust
mod palettes;
```
And then in `palettes.rs` do:
```rust
mod java {
  fn resolve(name: &str) -> Palette {}
}
mod perl {
  fn resolve(name: &str) -> Palette {}
}
```
etc. That strikes a good middle-ground.

src/flamegraph/color.rs
JordiChauzi: [Ok](https://github.com/jonhoo/inferno/pull/32/commits/40386d59b8986d0089d0c998c339915be353062a)

src/flamegraph/color.rs
JordiChauzi: [Ok](https://github.com/jonhoo/inferno/pull/32/commits/40386d59b8986d0089d0c998c339915be353062a)

src/flamegraph/color.rs
JordiChauzi: [Ok](https://github.com/jonhoo/inferno/pull/32/commits/40386d59b8986d0089d0c998c339915be353062a)

src/flamegraph/svg.rs
```diff
@@ -16,6 +16,22 @@ pub(super) struct TextItem<'a, I> {
...
 }
 
+pub(super) struct StyleOptions<'a> {
+    imageheight: usize,
+    bgcolor1: &'a str,
+    bgcolor2: &'a str,
+}
+
+impl<'a> StyleOptions<'a> {
+    pub(super) fn new(imageheight: usize, bgcolor1: &'a str, bgcolor2: &'a str) -> Self {
```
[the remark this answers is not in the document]
JordiChauzi: [Ok](https://github.com/jonhoo/inferno/pull/32/commits/e221f9084f72d6732ebac6ba3d53a4d9a6c1693b)

src/flamegraph/color.rs
JordiChauzi: The `palettes` mod will be defined in `flamegraph/mod.rs`, not `color.rs` right ?

src/flamegraph/color.rs
JordiChauzi: And remove the MultiPalette::Chain entry totally?
`chain` is not in the `possible_values` of the `colors` option, I think it is sufficient for now.

src/flamegraph/color.rs
JordiChauzi: [Implemented here](https://github.com/jonhoo/inferno/pull/32/commits/ad146b62196810a94e99b97231c57fc463293b59)

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+        java_prefix.starts_with("sun/") {
+        Palette::Green
+    } else if s.contains("::") {
+        Palette::Yellow
+    } else {
+        Palette::Red
+    }
+}
+
+fn handle_perl_palette(s: &str) -> Palette {
```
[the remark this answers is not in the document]
JordiChauzi: [Done here](https://github.com/jonhoo/inferno/pull/32/commits/ad146b62196810a94e99b97231c57fc463293b59)

src/flamegraph/color.rs
```diff
@@ -0,0 +1,272 @@
...
+    } else if s.contains("Perl") || s.contains(".pl") {
+        Palette::Green
+    } else if s.ends_with("_[k]") {
+        Palette::Orange
+    } else {
+        Palette::Red
+    }
+}
+
+fn handle_js_palette(s: &str) -> Palette {
```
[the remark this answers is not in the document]
JordiChauzi: [Done here](https://github.com/jonhoo/inferno/pull/32/commits/ad146b62196810a94e99b97231c57fc463293b59)

src/flamegraph/color.rs
JordiChauzi: [Done here](https://github.com/jonhoo/inferno/pull/32/commits/ad146b62196810a94e99b97231c57fc463293b59)

src/flamegraph/color.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/22b135baa9b70711241cf473836b9d961be4dea6)

src/flamegraph/color.rs
jonhoo: Yeah, I would remove it totally so that we don't have dead code laying around :)

src/flamegraph/palettes.rs
```diff
@@ -0,0 +1,111 @@
+pub(super) mod java {
+    use crate::flamegraph::color::BasicPalette;
```
jonhoo: I _think_ it's possible to just put this `use` at the top of the file?

src/flamegraph/mod.rs
```diff
@@ -9,6 +9,7 @@ use std::io::prelude::*;
 
 mod color;
 mod merge;
+mod palettes;
```
jonhoo: I think we'll want this as a submodule of `color`. I agree it's a little annoying to add a subdirectory for color, but I do think the module hierarchy makes more sense that way. We'd then have:
```
src/flamegraph/
  color/mod.rs <- what is now in color.rs
  color/palettes.rs <- what is now in palettes.rs
```

JordiChauzi: Seems fine. We can add a `palette_map.rs` file in here if we move the future `PaletteMap` struct to its own file.

src/flamegraph/color.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/2c486fe79d8cc76cf5cf43ba0ff5479cb0006293)

src/flamegraph/mod.rs
JordiChauzi: [Moved](https://github.com/jonhoo/inferno/pull/32/commits/4a774c8463b75acace1f20018da4f3b1ba4dff0f)

jonhoo: Ah, yes, moving `PaletteMap` too seems like a great idea!

JordiChauzi: [Moved `palette_map` functions](https://github.com/jonhoo/inferno/pull/32/commits/1ea68876e45d4afefc6674dbf50c81ed4e96d764)

src/flamegraph/color.rs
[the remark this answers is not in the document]
JordiChauzi: Not the simplest [commit](https://github.com/jonhoo/inferno/pull/32/commits/7383d25e492288f23102cf8001455af6fac322ec), sorry.

src/flamegraph/color.rs
JordiChauzi: [Done here](https://github.com/jonhoo/inferno/pull/32/commits/7383d25e492288f23102cf8001455af6fac322ec).
I was not correctly handling some cases I think.
I think it is correct (and commented) now.

src/flamegraph/color.rs
JordiChauzi: So we need to:
- parse the rgb string when loading the palette from a file to extract the RGB components `(u8, u8, u8)`
- when rendering, for each frame, retrieve (or compute) the RGB components and format them into a string
- when saving the palette, for each function name, format the associated RGB components into a string

Just making sure that this is the direction that you want to take.

src/flamegraph/color/mod.rs
```diff
@@ -77,41 +72,83 @@ impl FromStr for Palette {
...
+    // If the computation finishes normally, we search for the first next '`'.
+    // After that, either we have found a '`' (end of prefix), and we need to compute the hash from there,
+    // or there is no '`' in the iterator and we have the hash computed!
+
+    match name.next() {
+        None => return namehash_variables.result(),
+        Some(first_char) => namehash_variables.update(first_char),
+    }
+
+    for character in name.by_ref() {
```
jonhoo: I guess we could just do:
```rust
while let Some(character) = name.next() {
```
which is a little easier to follow (I wasn't familiar with `by_ref()`)

jonhoo: Or alternatively `by_ref().take(3)` and get rid of the `modulo` check?

JordiChauzi: I think I prefer removing the `modulo` check, but I admit it is not easy to immediately know how much the iterator has been consumed with `by_ref`...

JordiChauzi: [Here](https://github.com/jonhoo/inferno/pull/32/commits/01dacc3086dfedad8083642cf4486d4aad186a07)

src/flamegraph/color.rs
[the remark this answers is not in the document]
jonhoo: Haha, yeah, that's quite the method. I think the comment helps though, thanks!

src/flamegraph/color.rs
jonhoo: - Yes
 - Sort of -- given #37, for each frame we actually just have to write the number to the stack, and don't need to allocate. That's the key for why I'd like to see this change!
 - Yes

src/flamegraph/color/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    fn result(&self) -> f32 {
+        (1.0 - self.vector / self.max)
+    }
+}
+
+/// Generate a vector hash for the name string, weighting early over
+/// later characters. We want to pick the same colors for function
+/// names across different flame graphs.
+fn namehash<I: Iterator<Item = u8>>(mut name: I) -> f32 {
```
jonhoo: A couple of unit tests for this function would probably be a good idea. Let's specifically double-check what the same perl code gives to make sure we're not messing anything up!

jonhoo:
Okay, so I _believe_ what's remaining for this PR is:

 - Merge in `master` for some mostly unrelated changes that have happened there
 - Change the `PaletteMap` to hold `(u8, u8, u8)`: https://github.com/jonhoo/inferno/pull/32#discussion_r254436744
 - Add unit tests for `namehash`: https://github.com/jonhoo/inferno/pull/32#discussion_r254437319
 - Add an integration test that compares the final SVG to a known-good one: https://github.com/jonhoo/inferno/pull/32#issuecomment-460395134. We can probably do that in a separate PR though if you'd rather not do that work here. I suspect we'll also want, as a part of that, to visually compare our output to the Perl version (the SVG contents won't actually be the same) as a sanity check that we indeed produce the same end result as it does.

src/flamegraph/color.rs
JordiChauzi: Done ([here](https://github.com/jonhoo/inferno/pull/32/commits/ff9625bae8e6f1d291d026fea6653dbc09afb0ba), and [here](https://github.com/jonhoo/inferno/pull/32/commits/1c21eca0e3ca0cf308a6b80f97eef497d74f8d3a))

src/flamegraph/color/mod.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/32/commits/3f11a0397b84fb20770cc66c7659bbac2cd919f7).
Which helped fix [this bug](https://github.com/jonhoo/inferno/pull/32/commits/eba2d52a6cae571ad94f6c9f69eb95ee1fd2e901).

This is not exactly the expected values in the Perl implementation, because perl floats are double-precision values.

JordiChauzi:
I am not sure how to write a robust integration test for that feature. Comparing the whole svg seems fragile at best, and looking for the correct info in the svg file does not feel feasible.
Maybe adding unit tests on the `rgb_components_for_palette` function can be an alternative ? Or do you want to compare the whole svg file ?

jonhoo:
Ah, I guess part of the problem is the randomness that's used to generate the colors, which would mean that we don't get the same colors each time, even for the same input. Unit tests for `rgb_components_for_palette` seems like a good alternative though -- I like that!

Perhaps we could also include a full SVG integration test for `-cp`? Specifically, run `flamegraph.pl` with `-cp`, and then run `inferno-flamegraph` with the same palette map and (manually) visually compare the two to see that they're the same. Then check in the SVG we generate + the palette map file, and then we add an integration test to see that that generated SVG doesn't change?

src/flamegraph/color/mod.rs
jonhoo: I'd be okay with moving our implementation to also use `f64` if you think that change is worthwhile for consistency with `flamegraph.pl`?

JordiChauzi:
> Ah, I guess part of the problem is the randomness that's used to generate the colors, which would mean that we don't get the same colors each time, even for the same input. Unit tests for `rgb_components_for_palette` seems like a good alternative though -- I like that!
> 
> Perhaps we could also include a full SVG integration test for `-cp`? Specifically, run `flamegraph.pl` with `-cp`, and then run `inferno-flamegraph` with the same palette map and (manually) visually compare the two to see that they're the same. Then check in the SVG we generate + the palette map file, and then we add an integration test to see that that generated SVG doesn't change?

Actually `--hash` is enough to make the output deterministic.
My worry is mostly about the potential unrelated changes to the SVG that would break these integration tests. It may be hard to "fix" these tests if they broke.

src/flamegraph/color/mod.rs
JordiChauzi: I don't think this is needed. The double precision is basically useless in this use case0

jonhoo:
Hmm, `--hash` will specifically enable the function name hashing though. I think we'd want tests for non-hashed function names too, but maybe unit tests of `rgb_components_for_palette` is good enough for that...

I'm not so worried about other changes to the SVG breaking the tests. In some sense, that's what I would expect them to do. Besides, fixing it is just a matter of updating the expected SVG file.

I suppose one way to fix this would be to turn the SVG into a PNG, and then compare the _visual_ output using [`magick compare`](https://www.imagemagick.org/script/compare.php). Specifically:
```console
$ magick compare -metric PSNR known-good.png new.png /dev/null
```
will exit with an exit code of 0 if the two images are visually the same. That way, we won't be subject to changes in the SVG encoding that don't affect visual appearance!

jonhoo:
@JordiChauzi let's leave the integration test for another PR and try to land this as-is instead :) It's an excellent and much-desired change! Can you take a look at the conflicts? I think they should be pretty easy to resolve.

Also, I just noticed that https://github.com/brendangregg/FlameGraph/pull/197 landed. Maybe we should wrap that up into this while we're at it too?

JordiChauzi:
clippy seems to be angry with the `from_sorted_lines` function: 
`warning: the function has a cyclomatic complexity of 29`

I am okay with splitting the function up into smaller ones. The attributes management seems like a good start ?

JordiChauzi:
While writing the integration test, I found one of the difference with `flamegraph` that we talked about before: 
`./flamegraph/flamegraph.pl flamegraph/test/results/perf-js-stacks-01-collapsed-all.txt --hash --colors js --bgcolors green`
 will not produce the same output as `inferno` with the same arguments.

[The difference in pictures](https://imgur.com/a/68DgJA8).

In `js::resolve`, we select the green palette if the name is composed of empty spaces. In `flamegraph.pl`, it is only if the name is composed of a single empty space.

Are we still sure that we want a different behavior ?

JordiChauzi:
I added an integration test. I have another test ready for the `consistent palette` feature, but because we cannot currently choose the path of the palette file, it is not possible to integrate it for now...

tests/flamegraph-nameattr.rs
```diff
@@ -15,6 +15,10 @@ fn flamegraph_nameattr() {
     let nameattr_file = "./tests/data/nameattr/nameattr.txt";
 
     let options = flamegraph::Options {
```
jonhoo: I think what we should probably do here is
```rust
let options = flamegraph::Options::default();
options.hash = true;
options....
```

src/flamegraph/mod.rs
```diff
@@ -52,6 +52,67 @@ macro_rules! args {
     }};
 }
 
+struct FrameAttributes<'a> {
```
jonhoo: Why have a type here that's separate from `FrameAttrs`?

src/flamegraph/mod.rs
```diff
@@ -302,7 +323,7 @@ where
         )?;
 
         buffer.clear();
-        if href.is_some() {
+        if href_is_some {
```
jonhoo: Can't this be `frame_attributes.href.is_some()`?

jonhoo:
Ah, good catch with the integration test! I think the test we _actually_ want is:
```rust
if !func.is_empty() && func.trim().is_empty() {
```

src/flamegraph/mod.rs
JordiChauzi: I ran into an issue with conflicting borrowings. I chose to reduce the needed lifetime of `frame_attributes` to resolve it, by computing `href.is_some()` earlier.

src/flamegraph/mod.rs
JordiChauzi: `FrameAttrs` owns its fields. We only need to have reference to the fields when drawing the frame. I didn't want to copy some `String` and `Vec`.

jonhoo: Wouldn't `Cow` let you merge these types then?

tests/data/colors/js-perl.svg
```diff
@@ -342,80 +342,68 @@
...
 <g id="frames">
 <g >
-<title>Stub:JSEntryStub (2 samples, 100.00%)</title><rect x="10.0" y="133" width="1180.0" height="15.0" fill="rgb(88,200,200)" rx="2" ry="2" />
-<text  x="13.00" y="143.5" >Stub:JSEntryStub</text>
-</g>
-<g >
-<title>uv__async_io (2 samples, 100.00%)</title><rect x="10.0" y="245" width="1180.0" height="15.0" fill="rgb(223,84,84)" rx="2" ry="2" />
-<text  x="13.00" y="255.5" >uv__async_io</text>
+<title>uv__io_poll (2 samples, 100.00%)</title><rect x="10.0" y="261" width="1180.0" height="15.0" fill="rgb(223,84,84)" rx="2" ry="2" />
+<text  x="13.00" y="271.5" >uv__io_poll</text>
```
jonhoo: I'm not sure I follow why this file had to change in this way as a result of this change?

JordiChauzi: I actually don't know. Same command line, same output (at least visually...)

jonhoo: That's pretty disturbing, because this changes locations, not just colors... Could just have been merging changes from `master` though I suppose.

src/flamegraph/mod.rs
JordiChauzi: Other issue: most `FrameAttrs` fields are optional, which would add useless `unwrap` for some values that we know have values (because we have a default value ready to be used)

jonhoo: Okay, I guess you've convinced me. It just still makes me sad :p

JordiChauzi: My goal here was to reduce the complexity of `from_sorted_lines` (at least for clippy). Adding a structure may not be ideal, I may try something else (at least to make clippy happy, we may need to open another issue to fix the problem once and for all).

tests/data/colors/js-perl.svg
JordiChauzi: It actually only change the svg element layout... I thought it could be because I'm juggling between multiple `flamegraph.pl`, but some options (`--bgcolors`) are not available on the version we point at in the `inferno` repo, so I could not have use the wrong version.

JordiChauzi: `flamegraph.pl` output does not seem stable ? I generated `js-perl.svg` again, and I have other diffs...

jonhoo: Ohh, I wonder if this is because it uses a map to keep attributes, and iteration over the map is probably non-deterministic :/

Separately, feel free to do update the `flamegraph` submodule!

jonhoo:
@JordiChauzi are there any further TODOs for this PR now?

JordiChauzi:
> @JordiChauzi are there any further TODOs for this PR now?

I don't think so !

We can always do the following tasks in other PRs:
- handle the `chain` palette
- add more code coverage
- split up the `from_sorted_lines` function to reduce complexity

jonhoo:
Looks excellent! Time to merge :D Thank you, and great work :tada:

Issue #28, opened 2019-02-03
Support working with multiple collapsed files


jonhoo:
The trick to merge frames we use in `flow` only works if the input lines are [sorted](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L607). This is generally the case when we're operating over a single file, but is _not_ true if the user provides multiple input files (which we should support), and we should cater to that case. Sorting inputs will also be required for #21. Note that it _is_ fine for the sort to be [`unstable`](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable).

PR #35, opened 2019-02-05
Support working with multiple collapsed files #28

jasonrhansen:


jasonrhansen:
What do you think of something like this for #28?

src/bin/flamegraph.rs
```diff
@@ -18,21 +21,21 @@ impl Into<Options> for Opt {
...
 
 fn main() -> quick_xml::Result<()> {
-    let (infile, options) = {
+    let (infiles, options) = {
         let opt = Opt::from_args();
-        (opt.infile.clone(), opt.into())
+        let infiles = opt
+            .infiles
+            .iter()
+            .map(|f| File::open(f).map_err(quick_xml::Error::Io))
```
jonhoo: I wonder if we can special-case `-` here somehow so that I could do:
```console
$ inferno-flamegraph foo.folded -
```

src/bin/flamegraph.rs
```diff
@@ -18,21 +21,21 @@ impl Into<Options> for Opt {
...
-        None => {
-            let stdin = io::stdin();
-            let r = BufReader::with_capacity(128 * 1024, stdin.lock());
-            flamegraph::from_reader(options, r, io::stdout().lock())
-        }
+    if infiles.is_empty() {
+        let stdin = io::stdin();
+        let r = BufReader::with_capacity(128 * 1024, stdin.lock());
+        flamegraph::from_readers(options, iter::once(r), io::stdout().lock())
+    } else {
```
jonhoo: Maybe special-case when `infiles.len() == 1`, since we then don't need to re-sort?

jonhoo:
Yeah, looks pretty reasonable! The collect + sort makes me a little sad, but :shrug:

jasonrhansen:
We can avoid the collect + sort when only one file is passed in, but can we avoid for multiple files?

src/bin/flamegraph.rs
jasonrhansen: Do you want to be able to handle `-` with files also passed in, or only by itself?

jonhoo:
No, I doubt it sadly. I was just indicating my sadness that that's the case :p

src/bin/flamegraph.rs
jasonrhansen: I just pushed a commit that handles `-` by itself.

jonhoo: I was thinking specifically when it is combined with other files. That might be tricky to pull off though sadly, since the `Read` is a different type. We could get around this by `Box`ing the `Read` we pass to `from_readers`, but that's a little sad too. Handling `-` on its own I don't think is worth it.

jasonrhansen: OK, I got that working.

PR #36, opened 2019-02-05
Use log crate in flamegraph module

jasonrhansen:


src/flamegraph/mod.rs
```diff
@@ -28,13 +28,13 @@ pub struct Options {}
...
     if ignored != 0 {
-        eprintln!("Ignored {} lines with invalid format", ignored);
+        warn!("Ignored {} lines with invalid format", ignored);
     }
 
     // let's start writing the svg!
     let mut svg = Writer::new(writer);
     if time == 0 {
-        eprintln!("ERROR: No stack counts found");
+        warn!("ERROR: No stack counts found");
```
jonhoo: I think this should be `error!`, and the "ERROR:" part of the message should be cut.

jasonrhansen: Yes, of course.

PR #37, opened 2019-02-06
Eliminate all(?) allocation from flamegraph loop

### LINKED ISSUE #33: Avoid `format!` in the inner loop.
Currently, we use `format!` in a number of places in order to pass appropriate values to `BytesStart::with_attributes`. That's really sad, because it means we're allocating tons of little `String` items. It'd be far better if we instead just had a single buffer that use `write!()` to write into, and then pulled out slices of to get the attribute substrings. I don't know if such a crate exists, but if it doesn't, that seems like a useful thing to write!

jonhoo:
Fixes #33.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -236,6 +236,7 @@ dependencies = [
  "pretty_toa 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "quick-xml 0.13.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "smallvec 0.6.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "str_stack 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "structopt 0.2.14 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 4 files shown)


jonhoo:
Ah, good catch! Fixed.

jasonrhansen:
It's not a big deal, but it looks like `StrStack::new` allocates the internal `ends` Vec initially with a capacity of 1 since `with_capacity` adds 1 to the passed in `strings` size. Do you think we should call `with_capacity` with the expected number of strings? Maybe a `bytes` capacity too?
```Rust
    pub fn new() -> StrStack {
        StrStack::with_capacity(0, 0)
    }

    /// Create a new StrStack with the given capacity.
    ///
    /// You will be able to push `bytes` bytes and create `strings` strings before reallocating.
    #[inline]
    pub fn with_capacity(bytes: usize, strings: usize) -> StrStack {
        let mut stack = StrStack {
            data: String::with_capacity(bytes),
            ends: Vec::with_capacity(strings+1)
        };
        // Yes, I know I don't need this. However, putting this here avoids checks later which
        // makes this much faster.
        stack.ends.push(0);
        stack
    }
```

jonhoo:
So, the reason I didn't explicitly do that was that those allocations will only happen on the very first iteration. Once we've passed that, since we're re-using the same `StrStack`, the memory used by the previous iteration will be re-used (`clear()` doesn't de-allocate). At least unless I'm missing something obvious?

Separately from that, it's actually a little tricky to anticipate how much capacity we'll need for `StrStack`. We could probably anticipate the length of `ends`, but the length of `bytes` is highly variable, and therefore we might as well just let it "figure itself out" as we go through the iterations.

jasonrhansen:
You're right. We could anticipate the length of `ends`, but it's not really worth it since that might change in the future and it won't really have an impact on performance.

PR #38, opened 2019-02-07
Replace pretty_toa with num-format.

### LINKED ISSUE #29: itoa with separators that avoids allocation
Hey Jon. Just watched your most recent stream. I wrote a crate recently that does itoa with thousands separators without allocation. It's not up on crates.io yet, but will be shortly. [num-format](https://github.com/bcmyers/num-format).

bcmyers:
Unlike pretty_toa, num-format can create
a &str representation of a number (with
thousands separators) without allocating.

Creating this kind of string representation from
a number only occurs at one place in inferno's
code; so replacing pretty_toa with num-format
is a tiny improvement, but it is an improvement
nontheless as it avoids an unnecessary allocation
(which occurs in a loop).

Fixes #29.


bcmyers:
Your Windows CI identified a small bug in num-format (thanks!). It was implementing a trait unconditionally on a unix-only type (instead of putting it behind a #[cfg(unix)] flag.  I’m publishing a bug fix update to num-format shortly and will update this code to use version 0.2.1 as opposed to 0.2.0 once it’s up. Then the Windows CI should pass.

bcmyers:
Alight ... So that wasn't my only Windows issue. `libc::lconv` is not available on Windows; so let me figure out the best way to update num-format to work on windows and then update this PR once that's finished. Will come back to you.

ErichDonGubler:
Pinging @bcmyers: any update?

bcmyers:
Need a couple more days (Friday maybe?), but have been actively working on 0.3.0 of num-format. The changes needed for a new version of num-format have nothing to do with the part that would be used in this code (they relate to asking linux, windows, bsd.. for locale information, which is not needed here), but nevertheless I want to get that right before putting up another version on crates.io. Once it's up, making this PR work will be trivial.

bcmyers:
I just published a new version of num-format on crates.io, which fixes the issues I was having on Windows. This PR now incorporates those changes. Let me know if you have any questions / issues.

src/flamegraph/mod.rs
```diff
@@ -121,7 +123,10 @@ where
         };
 
         let samples = frame.end_time - frame.start_time;
-        let samples_txt = samples.thousands_sep();
+
+        // add thousands separators to `samples`
+        let _ = samples_txt_buffer.write_formatted(&samples, &Locale::en);
```
jonhoo: Doesn't this buffer have to be cleared?

jonhoo:
That's an _awful_ lot of new dependencies just for number formatting :/ This pulls in `hashbrown`, `regex`, and `termcolor` to name a few that struck me as particularly odd. Any way we can avoid those?

bcmyers:
Yah, the dependency explosion is due to stuff needed to support asking your OS for its locale (which is not needed here). It's not as bad as it looks since only Windows is the problem (and when you build on Unix the Windows dependencies don't come in). That said, what I need to do is put all the OS stuff behind a feature flag or something; so if you don't need that functionality you don't bring in a bunch of unneeded dependencies. That would get rid of all these dependencies as it relates to inferno. Let me take a crack at that and get back to you. This change should be much shorter.

src/flamegraph/mod.rs
bcmyers: No. Every time you call `write_formattted` on a `num_format::Buffer` it essentially takes care of clearing.

bcmyers:
Dependencies cut down to just `arrayvec` and `itoa`.

src/flamegraph/mod.rs
jonhoo: Huh, I wouldn't have guessed from the name, but sounds good!

jonhoo: Should we `.expect()` here since the writer should never error?

bcmyers: Yah, I should probably add a note in the docs that that's the behavior. I said "essentially" takes clear of clearing, by the way, because it doesn't do the naive thing of clearing bytes to zero or anything like that. Underneath, `Buffer` contains an array and two indices, one indicating the beginning of valid data in the array and the second indicating the end of valid data. "Clearing" just means changing those indices to both point to the end of the array (or both to the beginning - it's the same). Anyway, `write_formatted` takes care to always reset those indices before writing bytes; so this code does the thing you want. Just figured you might appreciate a bit more commentary.

jonhoo: As long as it re-uses memory but not content, I'm happy :p

bcmyers: `write_formatted` doesn't return a `Result`; it returns the number of bytes written. The code is written like that because I usually turn on a lint that prevents me from not using the return value of functions. If you think that's ugly style, though; just let me know and we can change the line to just plain old `samples_txt_buffer.write_formatted(&samples, &Locale::en);`.

jonhoo: Nah, that's fine then!

Issue #24, opened 2019-02-03
Custom per-function name attributes


jonhoo:
The original flame graph supports the [`--nameattr` option](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L172), which lets you specify a file that [contains](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L209L219) "extra" attributes to use [for the SVG frames](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L1097L1103) of particular functions. This seems like a pretty neat feature that we should support. Be careful that those attributes are handled in accordance with [`group_start` and `group_end`](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L270L296) from `flamegraph.pl` though!

PR #39, opened 2019-02-08
Implement --nameattr option

jasonrhansen:
Implements #24


src/bin/flamegraph.rs
```diff
@@ -3,24 +3,38 @@ use std::io::{self, BufReader, Read};
...
     infiles: Vec<PathBuf>,
 }
 
 impl Into<Options> for Opt {
     fn into(self) -> Options {
-        Options {}
+        let func_frameattrs = match self.nameattr_file {
+            Some(file) => match FuncFrameAttrsMap::from_file(&file) {
+                Ok(n) => n,
+                Err(e) => panic!("Error reading {}: {:?}", file.display(), e),
```
jonhoo: I wonder whether panicking is the right thing to do here... Do you know what the Perl version does?

src/flamegraph/attrs.rs
```diff
@@ -0,0 +1,254 @@
...
+        let mut line = String::new();
+        loop {
+            line.clear();
+
+            if reader.read_line(&mut line)? == 0 {
+                break;
+            }
+
+            let mut line = line.trim().splitn(2, '\t');
+            if let Some(func) = line.next() {
```
jonhoo: To avoid the extreme indentation here, I think I would use:
```rust
let func = if let Some(f) = line.next() {
  f
} else {
  continue;
};
```
We could probably make that nicer with a `try .. catch` and `Option?` + `continue` on `None`, but that's a bigger transformation.

src/flamegraph/attrs.rs
```diff
@@ -0,0 +1,254 @@
...
+                if !func.is_empty() {
```
jonhoo:
```suggestion
                if func.is_empty() {
                    continue;
                }
```

src/flamegraph/attrs.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+struct AttrIter<'a> {
+    s: &'a str,
+}
+
+impl<'a> Iterator for AttrIter<'a> {
+    type Item = (String, String);
+
+    fn next(&mut self) -> Option<(String, String)> {
```
jonhoo: I think you want to take advantage of the fact that you can use `?` with `Option` to make this function nicer to read :)

jonhoo:
This looks pretty good! I think I'd like to see a test that actually passes in a `nameattr` file and checks the SVG output that those attributes are actually there, but that's my only major piece of feedback :)

src/bin/flamegraph.rs
jasonrhansen: I don't love panicking here either, but the Perl code will `die` if it can't open the file, or if it detects an invalid format.

```perl
        open my $attrfh, $nameattrfile or die "Can't read $nameattrfile: $!\n";
	while (<$attrfh>) {
		chomp;
		my ($funcname, $attrstr) = split /\t/, $_, 2;
		die "Invalid format in $nameattrfile" unless defined $attrstr;
		$nameattr{$funcname} = { map { split /=/, $_, 2 } split /\t/, $attrstr };
	}
```

src/flamegraph/attrs.rs
jasonrhansen: Yes, much better! I'll get a hang of this Rust thing eventually.

src/bin/flamegraph.rs
jonhoo: I guess what we really want is something like `TryInto` so that we could deal with the error more nicely in `main`. This seems fine for now though then!

src/flamegraph/attrs.rs
```diff
@@ -0,0 +1,248 @@
...
+pub struct AElementAttrs {
+    /// If set to None the `a` tag will not be added
+    pub href: Option<String>,
+    pub target: Option<String>,
+
+    /// Extra attributes to include
+    pub extra: HashMap<String, String>,
+}
+
+fn parse_extra_attrs(attrs: &mut HashMap<String, String>, s: &str) {
```
jonhoo: I think you can just do `attrs.extend(name_value_pairs(s))` here

src/flamegraph/attrs.rs
```diff
@@ -0,0 +1,248 @@
...
+    for (name, value) in name_value_pairs(s) {
+        attrs.insert(name, value);
+    }
+}
+
+fn name_value_pairs(s: &str) -> AttrIter {
```
jonhoo: I'd just remove this function. It's only called from one place, and that place might as well just directly create an `AttrIter`

src/flamegraph/attrs.rs
```diff
@@ -0,0 +1,248 @@
...
+struct AttrIter<'a> {
+    s: &'a str,
+}
+
+impl<'a> Iterator for AttrIter<'a> {
+    type Item = (String, String);
+
+    fn next(&mut self) -> Option<(String, String)> {
+        let equal = self.s.find('=')?;
+        let open_quote = self.s[equal..].find('"')?;
```
jonhoo: I don't think this is something the Perl version supports?

jonhoo: And the way this code is written, it looks like it _requires_ there to be quotes? Or am I missing something?

jonhoo: In fact, I think the Perl code specifically just has the key be everything up until the first `=`, and value be everything that follows until the next `\t` or the end of the string?

src/flamegraph/attrs.rs
```diff
@@ -30,38 +30,38 @@ impl FuncFrameAttrsMap {
...
+                f
+            } else {
+                continue;
+            };
+
+            if func.is_empty() {
+                continue;
+            }
+            let funcattrs = funcattr_map.0.entry(func.to_string()).or_default();
+            if let Some(namevals) = line.next() {
```
jonhoo: I would do the same thing here to reduce indentation.

src/flamegraph/attrs.rs
```diff
@@ -30,38 +30,38 @@ impl FuncFrameAttrsMap {
...
+                for nameval in namevals.split('\t') {
+                    let mut nameval = nameval.splitn(2, '=');
+                    if let Some(name) = nameval.next() {
```
jonhoo: And here. Maybe even write a small macro to do it for you:
```rust
macro_rules! unwrap_or_continue {
    ($e:expr) => {{ if let Some(x) = $e { x } else { continue; } }}
}
```
That way you could just write:
```rust
let func = unwrap_or_continue!(line.next());
```

jasonrhansen: Great idea!

src/flamegraph/attrs.rs
jasonrhansen: You're right that the Perl version doesn't actually parse the value for `g_extra` and `a_extra`. It does no checks to see if they are valid attributes and just includes the string as is in either the `g` or `a` tag. But we use quick_xml so we need to parse out the key-value pairs of the extra attributes so we can add them with `extend_attributes`.

Not having quotes on the extra attributes will cause issues in the Perl version since the attribute values in the SVG tags need to be quoted. That's why I require them in this version. 

I don't like the file format they chose because it's inconsistent. If you add quotes to the other attribute values, the quotes get included in the value so it's probably not what you want. But for extra attributes you need to have quotes or they won't be there on the tag. 

For example: if we want the `class` to be "inferno" we have to do `class=inferno` which becomes `class="inferno"` on the tag. But if we wan't to change the `stroke` of the `g` tag we have to do `g_extra=stroke="green"`.

jonhoo: Ah, I see what you mean... Hmm, that's unfortunate indeed. Here's what I would propose: make `"` quoting optionally available for all values. Specifically, do the following for all attributes (whether extra or non-extra):

 - `split('\t')` to get each `name=value` pair
 - `splitn(2, '=')` to get name and value
 - `trim()` both `name` and `value`
 - if `value.starts_with('"') && value.ends_with('"')`, then remove both of them

This isn't perfect, since people may start doing weird stuff like `\tname="val\tue"\t` and expect it to work correctly. Or something like `name="val\"ue"`. But at least it makes the handling consistent across all attribute types: the `"`s are always optional.

jasonrhansen: Sounds good to me.

jasonrhansen: Except that the extra attributes still need to be parsed a little differently since they can't be tab separated.

jonhoo: Ohh, I see, `g_extra` is actually of the form
```
g_extra=foo="bar" z="baz"
```
That's awkward.. Yeah, then I guess I would say that all attribute values should be allowed to be (but not required to be) `"`-enclosed. For `g_extra`, that would mean that these are all equivalent:
```
g_extra=foo=bar z="baz"
g_extra=foo="bar" z="baz"
g_extra="foo="bar" z="baz""
g_extra="foo=bar z="baz""
```
What an awful syntax :sweat_smile:

jasonrhansen: So I can't split on tabs for those, but I can't simply split on whitespace either because the values themselves could have spaces. What should we do about something like `g_extra=a= b  c d e = f`? Should this become `a="b  c d" e="f"`?

jonhoo: It's tricky, because in HTML, that would parse as `a="b" c d e="f"`, since HTML allows "[boolean attributes](https://w3c.github.io/html/infrastructure.html#sec-boolean-attributes)" whose value is implicitly set to the name of the attribute itself. People might expect the same here, even though XML [does not allow](https://www.w3.org/TR/2006/REC-xml11-20060816/#sec-starttags) that same construct.

I think what we should do is as follows: values are *either* `"`-enclosed, or the next non-whitespace sequence of characters. If a token does not match that, it is ignored and a warning is issued. So, parsing `g_extra=a= b  c d e = f` should give:
```
WARN extra attribute c has no value (did you mean to quote the value?)
WARN extra attribute d has no value (did you mean to quote the value?)
```
and the attributes `a="b"` and `e="f"`.

jasonrhansen: That makes sense.

jasonrhansen: I think I've got it working with your described behavior in the latest commit.

jonhoo: Yup, looks good to me!

jasonrhansen:
I would still like to add a test that checks the SVG output since we're currently only testing the parsing.

tests/flamegraph-nameattr.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use inferno::flamegraph;
+use std::fs::File;
+use std::io::{BufRead, BufReader, Cursor};
+use std::path::PathBuf;
+
+#[test]
+fn flamegraph_nameattr() {
+    let input_file = "./flamegraph/test/results/perf-cycles-instructions-01-collapsed-all.txt";
+    let expected_result_file = "./tests/data/nameattr/nameattr.svg";
```
jonhoo: I think you forgot to add this file? It could be that you need to change the line
```
*.svg
```
in `.gitignore` to read
```
/*.svg
```
instead first.

src/flamegraph/attrs.rs
```diff
@@ -113,7 +114,7 @@ pub struct GElementAttrs {
     pub onclick: Option<String>,
 
     /// Extra attributes to include
-    pub extra: HashMap<String, String>,
+    pub extra: IndexMap<String, String>,
```
jonhoo: This change is pretty unfortunate, but I see why you made it. Could we maybe instead just make this a `Vec<(String, String)>`? I don't think we actually _need_ it to be a map?

tests/flamegraph-nameattr.rs
jonhoo: We should manually verify that the svg matches what `flamegraph.pl` generates. The two won't be exactly similar due to newlines and rounding errors, but anything related to `nameattr` should be the same!

src/flamegraph/attrs.rs
jasonrhansen: You're right, a Vec would do just fine here.

jasonrhansen: You're right, a Vec would do just fine here.

tests/flamegraph-nameattr.rs
jasonrhansen: Oops!

jasonrhansen: I'll probably be able to get to this and the other fixes sometime tonight or tomorrow.

jasonrhansen: I'll probably be able to get to this and the other fixes sometime tonight or tomorrow.

jasonrhansen: Since we can't just simply check that our svg output exactly matches the Perl version, I'm trying to think of a nice way to do this that isn't overly complicated. Any suggestions?

jasonrhansen: We also end up with differences like this.
Our version: `onclick="console.log(&apos;main clicked&apos;)"`
Perl version: `onclick="console.log('main clicked')"`

jasonrhansen: Actually, I just read your comment again and realized that I missed the "manually" part. I thought you wanted our test to check for that, but now I don't think that's what you meant. I've already checked that they both do the same thing other than ours escaping the single quote.

jonhoo: Yup! I was specifically thinking that we'd manually check against the Perl version once, and then from that point forward just test our version against our version :p Have you checked that the JS works correctly even with that escaping? I'm actually not sure whether or not that'll do the right thing?

jasonrhansen: Yes, that JS works correctly in both versions.

jasonrhansen: Also, why won't Travis CI run? 
"Expected — Waiting for status to be reported"

jonhoo:
No idea why CI was sad. I've merged now and then I guess we should watch https://travis-ci.org/jonhoo/inferno/builds/491312148.

jonhoo:
Oh, and thanks! :D

PR #40, opened 2019-02-08
Refactor - assigning to unbound _ unnecessary

PrismaPhonic:
Removed unnecessary `let _`.  You can just call next, no need to declare with a let to an unbound `_`.


jonhoo:
Ah, yes, I suppose `Option<T>` isn't `#[must_use]` like `Result<T>` is.. Thanks!

PR #45, opened 2019-02-11
Fixes

jasonrhansen:
A few very minor changes.

### DIFF
--- a/tests/collapse-perf.rs
+++ b/tests/collapse-perf.rs
@@ -4,7 +4,7 @@ extern crate pretty_assertions;
 extern crate inferno;
 
 use inferno::collapse::perf::{handle_file, Options};
-use std::fs::File;
+use std::fs::{self, File};
 use std::io::{self, BufRead, BufReader, Cursor};
 use std::path::Path;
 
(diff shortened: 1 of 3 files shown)


tests/collapse-perf.rs
```diff
@@ -156,8 +156,8 @@ collapse_perf_tests! {
 
 fn test_collapse_perf(test_file: &str, expected_file: &str, options: Options) -> io::Result<()> {
     let r = BufReader::new(File::open(test_file)?);
-
-    let mut result = Cursor::new(Vec::with_capacity(expected_file.len()));
+    let expected_len = fs::metadata(expected_file)?.len() as usize;
+    let mut result = Cursor::new(Vec::with_capacity(expected_len));
```
jonhoo: That's a neat trick! I wonder if we also want to play the same trick in `bin/collapse-perf.rs`?

jasonrhansen: Sure, and maybe cap the buffer size to `128 * 1024` in case the file is really large?

```Rust
let len = fs::metadata(f)?.len() as usize;
let r = BufReader::with_capacity(std::cmp::min(len, 128 * 1024), File::open(f)?);
handle_file(options, r, io::stdout().lock())
```

jonhoo: Actually, in `bin/collapse-perf.rs` don't we need to have the entire file in memory at once? Maybe I'm misremembering? One alternative here would be to just [mmap](https://docs.rs/memmap/) the entire file :sweat_smile:

jasonrhansen: In flamegraph we read the input files into memory completely so we can sort the lines, but don't we only read the file a line at a time in collapse-perf?

jonhoo: Ah, yes, you're right. I had my thinking reversed!

jasonrhansen: So do you want to leave those as is with `BufReader`s, or should we use `mmap`?

jonhoo: I'd leave them as `BufReader`s for now. `mmap` comes with its own set of `unsafe` restrictions that I don't think are worth it.

PR #48, opened 2019-02-15
Implement "--inverted" flag

### LINKED ISSUE #22: Icicle graphs
The [`--inverted` flag](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L149), apart from [changing the title of the plot](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L205), changes the flame graph such that the entire plot is [up-side-down](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L1067L1068) (note that this also affects [the JavaScript](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L864)). We should probably support that flag too.

busarovalex:
Closes #22

### DIFF
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -14,9 +14,7 @@ struct Opt {
     #[structopt(long = "nameattr")]
     nameattr_file: Option<PathBuf>,
 
-    /// The `--inverted` flag.
-    /// Changes the flame graph such that the entire plot is up-side-down.
-    /// Also, changing the title of the plot.
+    /// Plot the flame graph up-side-down.
     #[structopt(short = "i", long = "inverted")]
     inverted: bool,
 
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -3,7 +3,7 @@ use std::io::{self, BufReader, Read};
 use std::path::PathBuf;
 use structopt::StructOpt;
 
-use inferno::flamegraph::{self, FuncFrameAttrsMap, Options};
+use inferno::flamegraph::{self, Direction, FuncFrameAttrsMap, Options};
 
 #[derive(Debug, StructOpt)]
 #[structopt(name = "inferno-flamegraph", author = "")]
@@ -13,6 +13,12 @@ struct Opt {
     /// then a sequence of tab separated name=value pairs.
     #[structopt(long = "nameattr")]
     nameattr_file: Option<PathBuf>,
+
+    /// The `--inverted` flag.
+    /// Changes the flame graph such that the entire plot is up-side-down.
+    /// Also, changing the title of the plot.
+    #[structopt(short = "i", long = "inverted")]
+    inverted: bool,
 
     /// Collapsed perf output files. With no INFILE, or INFILE is -, read STDIN.
     #[structopt(name = "INFILE", parse(from_os_str))]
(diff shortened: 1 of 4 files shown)


src/flamegraph/mod.rs
```diff
@@ -9,6 +9,7 @@ use quick_xml::{
     events::{BytesEnd, BytesStart, BytesText, Event},
     Writer,
 };
+use std::default::Default;
```
jonhoo: You don't need this `use`; `Default` is in the Rust prelude :)

src/flamegraph/mod.rs
```diff
@@ -305,3 +324,18 @@ fn deannotate(f: &str) -> &str {
     }
     f
 }
+
+impl Default for Direction {
```
jonhoo: Do we even need an `impl Default for Direction`? It doesn't seem to be used anywhere?

src/flamegraph/mod.rs
```diff
@@ -305,3 +324,18 @@ fn deannotate(f: &str) -> &str {
...
+    fn default() -> Self {
+        Direction::Straight
+    }
+}
+
+impl Direction {
```
jonhoo: I don't think `is_inverted` is used anywhere?

src/flamegraph/svg.rs
```diff
@@ -140,7 +146,7 @@ var searchcolor = 'rgb(230,0,230)';",
             size: super::FONTSIZE + 5,
             x: (super::IMAGEWIDTH / 2) as f64,
             y: (super::FONTSIZE * 2) as f64,
-            text: "Flame Graph".into(),
+            text: opt.title.clone().into(),
```
jonhoo: I don't think you need the `clone()` here? Specifically, I think `&str` should work just fine:
```rust
(&*opt.title).into()
```

jonhoo:
Looks good to me once the few comments above have been sorted out! Thanks :D

src/flamegraph/mod.rs
jasonrhansen: I think we need Direction to implement Default since Options does.

jonhoo: Given that we [directly construct `Options`](https://github.com/jonhoo/inferno/pull/48/files#diff-5b9f024b0146520495e6e19fa806f9daR47), why does `Options` have to `impl Default` then? :p

jasonrhansen: What about when someone wants to use it as a library? Wouldn't we want the ability to just use the default options?

jonhoo: Ah, hmm, I suppose that's true... In fact, arguably what we should do in `from_args` is construct a `default` and then only override with what the args say...

src/flamegraph/mod.rs
busarovalex: Done

src/flamegraph/mod.rs
busarovalex: Removed this method, added Copy,Clone,Eq,PartialEq to Direction.
In `svg.rs` used `opt.direction == Direction::Inverted`

src/flamegraph/svg.rs
busarovalex: Done

src/bin/flamegraph.rs
```diff
@@ -14,6 +14,12 @@ struct Opt {
     #[structopt(long = "nameattr")]
     nameattr_file: Option<PathBuf>,
 
+    /// The `--inverted` flag.
+    /// Changes the flame graph such that the entire plot is up-side-down.
+    /// Also, changing the title of the plot.
```
jonhoo: Actually, sorry, one more thing. Let's make this a little less noisy, and make it just:
```rust
/// Plot the flame graph up-side-down.
```

busarovalex: Done

jonhoo:
Great, thank you!

Issue #16, opened 2019-02-02
Make style settings for generated flamegraph configurable


jonhoo:
`flamegraph.pl` has [_a lot_](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L100L128) of configuration arguments, and many of them are there mainly to change the style of the resulting flame graph. We'll leave color schemes do a different issue, but I'm talking about things like

 - [x] `imagewidth`
 - [x] `frameheight`
 - [x] `fonttype`
 - [x] `fontsize`
 - [x] `fontwidth`
 - [x] `minwidth`
 - [x] `nametype`
 - [x] `countname`
 - [x] `bgcolors` (#32)
 - [x] `titletext`
 - [ ] `searchcolor`
 - [x] `notestext`
 - [x] `subtitletext`

Individually, these should all be pretty easy to add in support for, so we should do it! Even smaller PRs that just add one or two of these are warmly welcome! Try to stick to the [`flamegraph.pl` options](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L160L183) where possible.

jonhoo:
A bunch of these just landed in #52 thanks to @AnderEnder!

Issue #50, opened 2019-02-17
Port  Optimize FlameGraphs by utilizing CSS and modern JS from FlameGraph


jbcden:
This issue is to port https://github.com/brendangregg/FlameGraph/pull/189 to inferno. The original author has consented to us porting their work to this project: https://github.com/jonhoo/inferno/issues/26#issuecomment-464442290.

jonhoo:
Now that this has been merged upstream too, this should be even easier to land! :D

jonhoo:
I _really_ want to see this land. @jbcden any chance you might submit a PR for this?

jbcden:
@jonhoo I can try to get something put together this weekend. I do not think I will have time before then though; does that sound alright?

jonhoo:
@jbcden that sounds perfect!

jonhoo:
@jbcden gentle ping on this :)

jbcden:
@jonhoo Sorry about the delay! I had some personal stuff come up this weekend and have family coming to visit this week. I will try to open something if I get time, but I may not get to it for a couple of weeks. 😞

jonhoo:
@jbcden No problem at all! Our lives are far more important than some random piece of code on the internet. Enjoy your week+weekend! :D

jasonrhansen:
@jbcden Last night I was playing around with some of the changes that will be required for this. I could finish it up if you're not already working on it. I don't want to step on your toes. Let me know.

PR #52, opened 2019-02-19
style settings configuration

AnderEnder:
Work in progress: implementing some of arguments setting from #16


jonhoo:
Just a note that you should _probably_ do your work on top of @JordiChauzi's changes in https://github.com/jonhoo/inferno/pull/32, since they affect a lot of the style code!

AnderEnder:
@jonhoo, lets wait #32  to be merged then I will rebase and rework the code

This is a still experimental change, so I may to rethink the implementation, so any comments are welcome.

jonhoo:
Sounds like a good plan!

I just had a skim, and the move to `SVGWriter` is interesting.. I see why you did it, but I'm wondering if the types are actually used enough that it's worth the extra abstraction? I think this is particularly true if we land https://github.com/jonhoo/inferno/issues/50, since that also moves a bunch of user-defined values out of each element. I'll have to think a little more on it.

I think we'll also want to try to be a bit more helpful with descriptions of the flags than `flamegraph.pl` is. We don't need to stick to the very terse format that they use :) Also, I think `structopt`/`clap` will automatically mark a flag as optional if it's an `Option<T>`, so we don't also need to say that in the flag help text.

jonhoo:
@AnderEnder #32 has now landed!

src/bin/flamegraph.rs
```diff
@@ -31,15 +32,61 @@ struct Opt {
...
     #[structopt(long = "hash")]
     hash: bool,
+
     /// use consistent palette (palette.map)
     #[structopt(long = "cp")]
     cp: bool,
+
+    /// Change title text
+    #[structopt(name = "title", long = "title")]
+    title: String,
```
jonhoo: Shouldn't this also be an `Option<String>`, given that the user doesn't _have_ to give one?

src/bin/flamegraph.rs
```diff
@@ -31,15 +32,61 @@ struct Opt {
...
     /// colors are keyed by function name hash
     #[structopt(long = "hash")]
     hash: bool,
+
     /// use consistent palette (palette.map)
     #[structopt(long = "cp")]
     cp: bool,
+
+    /// Change title text
+    #[structopt(name = "title", long = "title")]
```
jonhoo: Is it necessary to give a `name` to all of these?

src/bin/flamegraph.rs
```diff
@@ -51,16 +98,18 @@ impl Into<Options> for Opt {
...
         } else {
             Direction::Straight
         };
-        let title = if self.inverted {
-            "Icicle Graph".to_string()
-        } else {
-            "Flame Graph".to_string()
+
+        let title = match self.title.as_str() {
+            "Flame Graph" if self.inverted => "Icicle Graph".to_string(),
```
jonhoo: Why would it ever be set to the string "Flame Graph"? That hasn't been declared anywhere else as far as I can see? Ideally I'd like it if the two "default" titles were both defined in the same place.

src/flamegraph/mod.rs
```diff
@@ -339,7 +371,8 @@ where
         };
         filled_rectangle(&mut svg, &mut buffer, &rect, color, &mut cache_rect)?;
 
-        let fitchars = (rect.width() as f64 / (FONTSIZE as f64 * FONTWIDTH)).trunc() as usize;
+        let fitchars =
+            ((x2 - x1) as f64 / (opt.font_size as f64 * opt.font_width)).trunc() as usize;
```
jonhoo: We should probably keep using `rect.width()` here.

src/flamegraph/mod.rs
```diff
@@ -49,6 +52,30 @@ pub enum Direction {
...
+            image_width: self.image_width,
+            frame_height: self.frame_height,
+            min_width: self.min_width,
+            font_type: self.font_type.clone(),
+            font_size: self.font_size,
+            name_type: self.name_type.clone(),
+            notes: self.notes.clone(),
+        }
+    }
+    pub fn ypad1(self: &Options) -> usize {
```
jonhoo: Do these need to be `pub`?

src/flamegraph/svg.rs
```diff
@@ -48,224 +48,270 @@ pub(super) struct StyleOptions<'a> {
...
-    svg.write_event(Event::Empty(
-        BytesStart::borrowed_name(b"stop").with_attributes(
-            iter::once(("stop-color", &*style_options.bgcolor2))
-                .chain(iter::once(("offset", "95%"))),
-        ),
-    ))?;
-    svg.write_event(Event::End(BytesEnd::borrowed(b"linearGradient")))?;
-    svg.write_event(Event::End(BytesEnd::borrowed(b"defs")))?;
+impl SVGWriter {
+    pub fn ypad2(self: &SVGWriter) -> usize {
```
jonhoo: It's a little unfortunate that we have to re-define this function (it's also defined further up in a different file).

src/flamegraph/svg.rs
```diff
@@ -48,224 +48,270 @@ pub(super) struct StyleOptions<'a> {
...
+                ("xmlns", "http://www.w3.org/2000/svg"),
+                ("xmlns:xlink", "http://www.w3.org/1999/xlink"),
+            ]),
+        ))?;
+        svg.write_event(Event::Comment(BytesText::from_plain_str(
+            "Flame graph stack visualization. \
+             See https://github.com/brendangregg/FlameGraph for latest version, \
+             and http://www.brendangregg.com/flamegraphs.html for examples.",
+        )))?;
+        svg.write_event(Event::Comment(BytesText::from_plain_str(
```
jonhoo: Shouldn't this be only if notes isn't empty? I don't remember what the Perl code does.

src/flamegraph/svg.rs
```diff
@@ -48,224 +48,270 @@ pub(super) struct StyleOptions<'a> {
...
+        ))?;
+        svg.write_event(Event::CData(BytesText::from_escaped_str(&format!(
+            "\
+    var nametype = '{}';
+    var fontsize = {};
+    var fontwidth = {};
+    var xpad = {};
+    var inverted = {};
+    var searchcolor = 'rgb(230,0,230)';",
+            self.name_type,
```
jonhoo: Here we probably have to do some escaping _just_ in case the font name has a `'` in it!

src/flamegraph/svg.rs
```diff
@@ -48,224 +48,270 @@ pub(super) struct StyleOptions<'a> {
...
+        let mut buf = StrStack::new();
+        self.write_str(
+            svg,
+            &mut buf,
+            TextItem {
+                color: "rgb(0, 0, 0)",
+                size: self.font_size + 5,
+                x: (self.image_width / 2) as f64,
+                y: (self.font_size * 2) as f64,
+                text: self.title.clone().into(),
```
jonhoo: I don't think the `clone().into()` is needed here? Can it just be `.to_string()` maybe?

src/flamegraph/svg.rs
```diff
@@ -48,224 +48,270 @@ pub(super) struct StyleOptions<'a> {
...
-                ("id", "search"),
-                ("onmouseover", "searchover()"),
-                ("onmouseout", "searchout()"),
-                ("onclick", "search_prompt()"),
-                ("style", "opacity:0.1;cursor:pointer"),
-            ],
-        },
-    )?;
+        match self.subtitle.clone() {
+            Some(subtitle) => self.write_str(
```
jonhoo: This should be:
```rust
if let Some(ref subtitle) = self.subtitle {
```

src/flamegraph/svg.rs
```diff
@@ -48,224 +48,270 @@ pub(super) struct StyleOptions<'a> {
...
+                svg,
+                &mut buf,
+                TextItem {
+                    color: "rgb(160, 160, 160)",
+                    size: self.font_size,
+                    x: (self.image_width / 2) as f64,
+                    y: (self.font_size * 4) as f64,
+                    text: subtitle.into(),
```
jonhoo: I don't think you need the `.into()` here

src/flamegraph/svg.rs
```diff
@@ -48,224 +48,270 @@ pub(super) struct StyleOptions<'a> {
...
 
-    svg.write_event(Event::Start(
-        BytesStart::borrowed_name(b"style").with_attributes(iter::once(("type", "text/css"))),
-    ))?;
-    svg.write_event(Event::Text(BytesText::from_plain_str(
-        ".func_g:hover { stroke:black; stroke-width:0.5; cursor:pointer; }",
-    )))?;
-    svg.write_event(Event::End(BytesEnd::borrowed(b"style")))?;
+    pub fn write_header<W>(
+        self: &SVGWriter,
```
jonhoo:
```suggestion
        &self,
```

jonhoo: Any particular reason why you didn't use the simpler form initially?

src/flamegraph/svg.rs
```diff
@@ -48,224 +48,270 @@ pub(super) struct StyleOptions<'a> {
...
+        });
+        text.push_attribute(Attribute {
+            key: b"y",
+            value: Cow::from(buf[y].as_bytes()),
+        });
+        text.push_attribute(Attribute {
+            key: b"font-size",
+            value: Cow::from(buf[fs].as_bytes()),
+        });
+        text.push_attribute(Attribute::from(("font-family", self.font_type.as_str())));
```
jonhoo:
```suggestion
        text.push_attribute(Attribute::from(("font-family", &self.font_type)));
```

src/bin/flamegraph.rs
AnderEnder: No, according to original implementation there is default, but I missed it.

src/bin/flamegraph.rs
AnderEnder: missed a default earlier

src/flamegraph/mod.rs
AnderEnder: missed during initial rebase.

jonhoo:
Also, just as a side note, now that we have comments pending, it'd be great if you can avoid force pushing. That way I can review just what has changed, rather than having to review the whole thing all over again :) We'll squash before merging anyway so the history remains clean!

src/flamegraph/svg.rs
AnderEnder: Perl code has default as empty string and use it in template without any condition.

https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L256

src/flamegraph/svg.rs
AnderEnder: Do we have a valid font with `'`?
Maybe it is better to validate it in arguments and to fail fast?

src/flamegraph/svg.rs
AnderEnder: Me as well, but compiler do not like the proposed either:
```
180 |                 text: &self.title,
    |                       ^^^^^^^^^^^ expected enum `flamegraph::svg::TextArgument`, found reference
    |
    = note: expected type `flamegraph::svg::TextArgument<'_>`
               found type `&std::string::String`
```
I decided to get back to optimisation later.

src/flamegraph/svg.rs
jonhoo: I don't _know_ of one, but that doesn't mean there aren't any :p
I don't think there's a good reason do _disallow_ such fonts? Probably better to just do
```rust
self.name_type.replace('\'', r#"\'"#)
```

jonhoo: Also, I just realized, this isn't the font name, but rather the "name" assigned to functions (basically the prefix `Function:`). There are _definitely_ cases where the user might use `'` in those.

src/flamegraph/svg.rs
AnderEnder: Unfortunately it does not work either:
```
error[E0308]: mismatched types
   --> src/flamegraph/svg.rs:195:27
    |
195 |                     text: subtitle,
    |                           ^^^^^^^^ expected enum `flamegraph::svg::TextArgument`, found reference
    |
    = note: expected type `flamegraph::svg::TextArgument<'_>`
               found type `&std::string::String`
```

jonhoo: Ah, yes, you might still need the `.into()` to make a `TextArgument` out of the `&str`, you're right!

src/flamegraph/svg.rs
AnderEnder:
```
   --> src/flamegraph/svg.rs:301:29
    |
301 |         text.push_attribute(Attribute::from(("font-family", &self.font_type)));
    |                             ^^^^^^^^^^^^^^^ the trait `std::convert::From<(&str, &std::string::String)>` is not implemented for `quick_xml::events::attributes::Attribute<'_>`
    |
    = help: the following implementations were found:
              <quick_xml::events::attributes::Attribute<'a> as std::convert::From<(&'a [u8], &'a [u8])>>
              <quick_xml::events::attributes::Attribute<'a> as std::convert::From<(&'a str, &'a str)>>
    = note: required by `std::convert::From::from`
```

src/flamegraph/svg.rs
jonhoo: Ah, what's going on here is that you need the `.into()` to get a `TextArgument`, but you _don't_ need the `clone`, since `TextArgument: From<&str>`: https://github.com/jonhoo/inferno/blob/c74e8b9dbd505fb5680212a17025e980ae331edb/src/flamegraph/svg.rs#L17

You may need to make Rust realize that you have a `&str` (not just a `&String`) by doing:
```rust
text: (&*self.title).into()
```

Alternatively you could add an `impl From<&String> for TextArgument` in `svg` :)

src/flamegraph/svg.rs
```diff
@@ -187,21 +183,20 @@ impl SVGWriter {
...
+            self.write_str(
                 svg,
                 &mut buf,
                 TextItem {
                     color: "rgb(160, 160, 160)",
                     size: self.font_size,
                     x: (self.image_width / 2) as f64,
                     y: (self.font_size * 4) as f64,
-                    text: subtitle.into(),
+                    text: subtitle.clone().into(),
```
jonhoo: See https://github.com/jonhoo/inferno/pull/52#discussion_r259893659. Specifically, this should work:
```rust
text: (&**subtitle).into()
```

src/flamegraph/svg.rs
AnderEnder: done moved all usage to this place only.

src/flamegraph/svg.rs
jonhoo: Ah, sorry, you need `&*self.font_type`. At that point it's also fine the way it was.

jonhoo:
Oh, now that we've had some other PRs land, there are also some conflicts with `master` that'll have to be resolved. I don't think any of them should be _too_ bad though. If you get stuck on any of them, please let me know and I'll take a look! :)

src/flamegraph/svg.rs
AnderEnder: done

src/flamegraph/mod.rs
AnderEnder: changed to pub(super)

src/bin/flamegraph.rs
AnderEnder: made a cleanup of names

src/flamegraph/svg.rs
AnderEnder: I think we need to have tests for this. It is already escaped in xml, so it should be escaped properly, probably replace is not a good option either.

src/flamegraph/svg.rs
AnderEnder: done

src/flamegraph/svg.rs
jonhoo: Ah, I'm not worried about the XML, but rather about the fact that we're creating a `'`-enclosed JavaScript string directly. It should be the case that `'` is the only character we _need_ to escape, though `\` may also be a good candidate (specifically by making it `\\`).

src/bin/flamegraph.rs
```diff
@@ -31,42 +32,109 @@ struct Opt {
...
     #[structopt(long = "hash")]
     hash: bool,
-    /// use consistent palette (palette.map)
+
+    /// Use consistent palette (palette.map)
     #[structopt(long = "cp")]
     cp: bool,
 
+    /// Change title text
+    #[structopt(long = "title", default_value = "Flame Chart")]
```
```suggestion
    #[structopt(long = "title", default_value = "Flame Graph")]
```

src/bin/flamegraph.rs
```diff
@@ -31,42 +32,109 @@ struct Opt {
...
+    count_name: String,
+
+    /// Name type label
+    #[structopt(long = "nametype", default_value = "Function:")]
+    name_type: String,
+
+    /// Set embedded notes in SVG
+    #[structopt(long = "notes", default_value = "")]
+    notes: String,
+
```
jonhoo: It's a little sad that we end up duplicating all the defaults between here and `Options::default()`. I don't have a good solution for that at the moment though. We'll probably have to introduce some consts somewhere. cc @jasonrhansen. It's fine for this PR though.

jonhoo:
@AnderEnder this is looking _really_ good now! Only the one type ("Chart" vs "Graph") and the JavaScript escaping left I think :)

AnderEnder:
and test cases :)

src/bin/flamegraph.rs
AnderEnder: I think something is possible in structopts, but it will introduce additional complexity which is not good either.

I would not worry because this looks like a strict line between cli options and library options.

tests/common.rs
```diff
@@ -0,0 +1,54 @@
...
+use pretty_assertions::assert_eq;
+use std::error::Error;
+use std::fs::{self, File};
+use std::io::{BufRead, BufReader, Cursor};
+
+pub fn generate_and_check(
+    input_file: &str,
+    expected_result_file: &str,
+    options: flamegraph::Options,
+) -> Result<(), Box<dyn Error>> {
```
jonhoo: I think we actually probably _want_ to use `unwrap` in tests, because it means we'll get a better read on exactly where the error occurred. Is there a particular reason you wanted to use `?` instead?

jasonrhansen:
@AnderEnder I reorganized and removed the duplication in the flamegraph tests in 59a9dacc6103a4f08dece2a54fedd012699fd5b3. It might be easier to revert your changes to the testing code, then merge `master` before adding any new tests.

tests/common.rs
AnderEnder: I found the message is not very useful for the context(git was cloned without submodules):
```
---- flamegraph_nameattr stdout ----
thread 'flamegraph_nameattr' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:997:5
```
Sure stack race can help, but real context is required for tests to understand the issue without stacktrace. I considered that `Result` in tests should help, but looks I was wrong.

In this case `expect` is more relevant than `unwrap`.  It should provide an additional context.

BTW, I deleted this commit.

src/flamegraph/svg.rs
```diff
@@ -118,16 +125,17 @@ where
...
-var nametype = 'Function:';
+var nametype = '{}';
 var fontsize = {};
 var fontwidth = {};
 var xpad = {};
 var inverted = {};
 var searchcolor = 'rgb(230,0,230)';",
-        super::FONTSIZE,
-        super::FONTWIDTH,
+        opt.name_type.replace("'", "\\'"),
```
jonhoo:
```suggestion
        opt.name_type.replace('\'', "\\'"),
```

jonhoo: Otherwise I suspect clippy will complain. Specifically, `replace` is faster if the pattern is a `char` than if it's a single-character `str`.

tests/flamegraph.rs
```diff
@@ -346,3 +345,169 @@ fn flamegraph_grey_frames() {
...
+    test_flamegraph(input_file, expected_result_file, options).unwrap();
+}
+
+#[test]
+fn flamegraph_font_type_with_quote() {
+    let input_file = "./tests/data/differential/perf-cycles-instructions-01-collapsed-all-diff.txt";
+    let expected_result_file = "./tests/data/options/font_type_with_quote.svg";
+
+    let options = flamegraph::Options {
+        font_type: "Andale Mono\"".to_owned(),
```
jonhoo: I think you want this to contain a single quote `'`, not a double quote `"`.

src/bin/flamegraph.rs
```diff
@@ -46,6 +48,50 @@ struct Opt {
     #[structopt(long = "cp")]
     cp: bool,
 
+    /// Change title text
+    #[structopt(long = "title", default_value = "Flame Graph")]
```
jonhoo: We should use `DEFAULT_TITLE` here too.

jonhoo:
We have a lot of repetition of the included JavaScript in all the tests now. I wonder if we want to have a separate test-only generator that `test_flamegraph` uses, and that _doesn't_ include the JS. That'll let us keep the fixture files much smaller. The JS is static anyway, so this shouldn't be too much of an issue, and it'll reduce false test failures when the JS changes. Thoughts?
/cc @jasonrhansen

jasonrhansen:
@jonhoo I agree that we shouldn't need to include the JavaScript in all of the tests since it's static. Do you think we should just add a `no-javascript` flag that `test_flamegraph` could set?

jonhoo:
Yeah, something like that. Probably make it `#[doc(hidden)]` since it's not something users should ever set. It has to be `pub` so that we can use it in integration tests sadly.

src/flamegraph/svg.rs
AnderEnder: clippy does not complain :)
But I found another case escaping case - `\`

tests/flamegraph.rs
AnderEnder: Font type  is used inside the xml tags:
```<text text-anchor="left" x="13.00" y="159.50" font-size="12" font-family="Verdana" fill="rgb(0, 0, 0)">__libc_start_main</text>```

So only it makes sense to check `"` for escaping.

Comment (jonhoo) [reply tests/flamegraph.rs]: Oh, yes, sorry, I got confused with `opt.name_type` that we're discussing above :)

File: src/bin/flamegraph.rs (hunk 20 above)
Comment (jasonrhansen) [reply src/bin/flamegraph.rs]: I think this should work:
```Rust
#[structopt(long = "title", raw(default_value = "DEFAULT_TITLE"))]
```

jonhoo: I think you have to use
```rust
#[structopt(long = "title", raw(default_value = DEFAULT_TITLE))]
```

jasonrhansen: It would be nice to have constants for all of the defaults, but the above trick will only work for `&str` constants. I don't know how we would use `f64` constants and others with `structopt`.

jonhoo: @jasonrhansen I think we should be able to use `raw` for all of them, no?

jasonrhansen: I just tried using `raw` with an `f64` constant and the compiler complained.

```
error[E0308]: mismatched types
 --> src/bin/flamegraph.rs:9:17
  |
9 | #[derive(Debug, StructOpt)]
  |                 ^^^^^^^^^ expected &str, found f64
  |
  = note: expected type `&str`
             found type `f64`
```

jonhoo: Hmm, that seems wrong. Maybe try filing an issue with structopt?

jasonrhansen: I think the problem is that `clap` itself requires a `&str`.
```Rust
pub fn default_value(self, val: &'a str) -> Self
```

jonhoo: Oooh, that's neat. I think I'd use a different field name than `name` though, since it's really something like `strval`. I'd expect `name` to be the name of the parameter.

Also keep in mind that `stringify!` doesn't quite do what you'd want, since it makes a string representation of the [tokens of the expression](https://doc.rust-lang.org/std/macro.stringify.html). For example, if you did
```rust
define!(PORT: u32 = 1024 + 16);
```
You'd end up with `default_value = "1024 + 16"`. We could perhaps limit it to take `$val:tt` instead (note the lack of a multiplier), which would only allow a single token, but I don't know if that's any better.

AnderEnder: IMHO, the solution should not overcomplicate the problem it solves, so it is better to leave this "duplication" as is.

jasonrhansen: I was thinking the same thing about the field being called `name`. `strval` makes more sense. And yes, we'll need to be careful to only use a single literal with the macro, but that's all we need anyway. It could definitely trip someone up though in the future. `$val::tt` would help, but wouldn't it still allow an expression in parentheses like this `(1024 + 16)`?

AnderEnder:
@jonhoo and @jasonrhansen
I found that writing tests(working with xml) and finding difference is not an easy and pleasant task, so I think about using prettified(human readable with indentation) xml for tests.

jasonrhansen:
@AnderEnder I agree that finding differences in the XML isn't nice since it's all on one line. For `collapse-perf` tests `pretty_assertions` works really well, but not for `flamegraph`. I wouldn't mind comparing prettified versions in tests. What do you think, @jonhoo?

src/bin/flamegraph.rs
jonhoo: @AnderEnder I actually quite worry about having mismatched defaults. They are _also_ being duplicated in docstrings, and being able to refer to public constants instead would be nice. We should do this as part of a separate PR though :)

@jasonrhansen I'm actually not sure; you might be right. But at least it stops the easiest to commit mistakes..

src/flamegraph/svg.rs
AnderEnder: Done with `enquote` crate.

jonhoo:
`pretty_assertions` plus prettified XML seems like a _great_ idea! I think we'll then _definitely_ want to strip out the JavaScript for tests though.

src/flamegraph/svg.rs
```diff
@@ -1,4 +1,5 @@
 use super::{Direction, Options};
+use enquote::enquote;
```
jonhoo: It's a little tempting to just bring in [the function](https://docs.rs/enquote/1.0.3/src/enquote/lib.rs.html#19-35) directly rather than adding an entire external dependency for something this simple..

jasonrhansen:
We can use `quick_xml::Writer::new_with_indent` instead of `quick_xml::Writer::new` to pretty format the XML. Is this something we want to do always, or only for tests? 

If we do it for tests only, should we add another flag for pretty-printed XML, or just add one flag for testing that both prettifies the XML and doesn't include the JavaScript?

The way I see it, a user would never want to generate a flame graph without JavaScript, but they might want to generate one with pretty XML. We could have the `no-javascript` flag use `#[doc(hidden)]`, but make the `pretty-xml` flag a documented feature.

What are your thoughts?

jonhoo:
`pretty-xml` seems like a good idea for a public flag, I agree! And `no-javascript` should be `#[doc(hidden)]`.

And yes, we want _that_ segment of JavaScript to be included, just not the giant JS file that's loaded with `include_str!`.

jasonrhansen:
So in the library we can use `#[doc(hidden)]` for `no-javascript` and in the CLI we can make the flag hidden like this.
```Rust
    #[structopt(raw(hidden = "true"), long = "no-javascript")]
    no_javascript: bool
```

jonhoo:
I don't think we need the flag to be present at all in the CLI. All the tests operate directly on the library, no?

jasonrhansen:
We don't necessarily need the flag on the CLI, but it would make it easier to create the test SVG files that our tests compare to. I suppose removing the JavaScript manually from the SVGs is easy enough though.

jonhoo:
Hmm, I suppose that's a good point. I think I'd be okay with a hidden CLI flag too then. Maybe with a help text that explains exactly what the intended use-case is.

jasonrhansen:
@AnderEnder I'm working on reorganizing the test data files. While I'm at it I can add the `no-javascript` and `pretty-xml` flags and modify all of the existing tests to use them. You could then update your tests to use those flags as well. I just want to make sure you're not already working on these specific changes. If you are, let me know.

jasonrhansen:
I've made those changes in #70.

jonhoo:
@AnderEnder a gentle ping on this -- it's a pretty wide change, and it will probably only get more painful to try to merge as time goes on. I think the only real remaining comment now is whether we simply import the `escape` function directly so we don't need the extra dependency? What do you think?

Also, I think @jasonrhansen's changes in #70 may let you simplify the tests somewhat, but he can advise on that better than I can.

jasonrhansen:
@AnderEnder #70 changes the flamegraph tests to not include JavaScript and to pretty-print the XML so it's easy to see diffs when there is a test failure. Because of this, you'll need to regenerate the test SVGs using the following flags `--no-javascript --pretty-xml`. 

The tests directory structure has also been altered. The files in `tests/data` are now organized into subdirectories for `flamegraph`, `collapse-perf`, etc. You'll want to put the test SVGs in a subdirectory inside `tests/data/flamegraph`.

jonhoo:
I just did the merge with `master`, and realized that there were a bunch of other smaller changes. I'll open PRs for each of those changes separately, and will force-push this soon to contain only the relevant changes.

src/bin/flamegraph.rs
jonhoo: Err, yeah, that doesn't work sadly:
```
error: proc-macro derive panicked
  --> src/bin/flamegraph.rs:11:17
   |
11 | #[derive(Debug, StructOpt)]
   |                 ^^^^^^^^^
   |
   = help: message: invalid structopt syntax: # [ structopt ( long = "title" , raw ( default_value = DEFAULT_TITLE ) ) ]

error: aborting due to previous error
```
Let's leave it for another PR.

jonhoo:
@jasonrhansen I believe this is now finally ready for review :tada:

src/bin/flamegraph.rs
jasonrhansen: `DEFAULT_TITLE` needs to be in quotes like this.
```
#[structopt(long = "title", raw(default_value = "DEFAULT_TITLE"))]
```

jonhoo: Won't that make the title quite literally "DEFAULT_TITLE"?

jasonrhansen: Not with `raw`.

jonhoo: Opened https://github.com/jonhoo/inferno/issues/91

jonhoo:
:+1: Filed as #91.

jonhoo:
:tada:

jonhoo:
Thanks for all your work on this @AnderEnder !

PR #53, opened 2019-02-20
Reuse Event::Start and BytesStart when drawing, saving allocations

### LINKED ISSUE #41: Share `BytesStart`s across frames in flamegraph
While #37 got us pretty far, there is still one source of allocation in the main flamegraph loop: the attributes on `BytesStart`. Specifically, when you add attributes to a `BytesStart`, it [allocates](https://docs.rs/quick-xml/0.13.2/src/quick_xml/events/mod.rs.html#164-172) a `Vec` to hold the underlying bytes if it doesn't already have a `Vec`. That's unfortunate since we create a new `BytesStart` for every `g`. If we instead created a single `BytesStart` that we re-used somehow (by resetting all the attributes), that'd be way more efficient. Sadly, that's not yet supported, but I've filed https://github.com/tafia/quick-xml/issues/148 which would enable us to do that!

meven:
Thanks @tafia for the release of [`quick-xml 1.3.3`](https://github.com/tafia/quick-xml/releases/tag/v0.13.3) and your suggestion

Fixes #41


src/flamegraph/mod.rs
```diff
@@ -103,6 +103,9 @@ where
     svg::write_header(&mut svg, imageheight)?;
     svg::write_prelude(&mut svg, imageheight, &opt)?;
 
+    // struct to reuse accross loops to avoid allocations
+    let event_start = &mut Event::Start({ BytesStart::owned_name("g") });
```
```suggestion
    let mut event_start = Event::Start({ BytesStart::owned_name("g") });
```

jonhoo: I'm surprised that that even works?

src/flamegraph/mod.rs
```diff
@@ -182,23 +185,27 @@ where
...
-        svg.write_event(Event::Start({
-            let mut g = BytesStart::borrowed_name(b"g").with_attributes(args!(
-                "class" => class
-            ));
-            if let Some(style) = style {
-                g.extend_attributes(std::iter::once(("style", style)));
-            }
+        if let Event::Start(ref mut g) = event_start {
+            // clear the BytesStart
+            g.clear_attributes();
```
jonhoo: You'll have to bump `quick-xml` to `0.13.3` in `Cargo.toml` for this to compile :)

meven:
`cargo test` reveals a bug, I am working on it.

src/flamegraph/mod.rs
meven: Yes `cargo test` runs. what is surprising ?

src/flamegraph/mod.rs
meven: No need, a new cargo build downloads quick-xml 0.13.3 as in Cargo.toml we have `quick-xml = 0.13`

meven:
> `cargo test` reveals a bug, I am working on it.

Well it is not a real bug, it is that the reference svg `./tests/data/nameattr/nameattr.svg` that we compare against a generated file does not correspond anymore. It is just the order of the attributes that changed between reference between reference and generated svg.

Should I edit the svg/xml or match the current reference file from the code ?
I like the code as it is as it is quite clean only one call to `g.extend_attributes(args!(`

src/flamegraph/mod.rs
meven: I should probably update the Cargo.lock file though.

jonhoo: I think you should explicitly set `0.13.3`, since we _require_ that version now.

jonhoo:
I would update the reference file with what is generated by your updated code :)

meven:
I have taken care of the review chances :)

jonhoo:
Thanks!

PR #56, opened 2019-02-22
add support for bpftrace

bcmyers:
Hopefully this does the trick. Git is still quite confusing to me. Thanks for the help!


src/bin/collapse-bpftrace.rs
```diff
@@ -0,0 +1,36 @@
...
+const READER_CAPACITY: usize = 128 * 1024;
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "inferno-collapse-bpftrace", author = "")]
+struct Opt {
+    /// bpftrace script output file, or STDIN if not specified
+    infile: Option<PathBuf>,
+}
+
+fn main() -> io::Result<()> {
```
jonhoo: I think it would be good if we could avoid all the duplication between this and the `collapse-perf.rs`..

src/collapse/bpftrace.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    #[test]
+    fn test_bpftrace() {
+        let root_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+        let data_dir = Path::new(&root_dir)
+            .join("tests")
+            .join("data")
+            .join("bpftrace");
+
+        for i in 1..NUMBER_OF_TEST_INPUTS + 1 {
```
jonhoo: I think I would probably instead just [walk the files in the directory](https://doc.rust-lang.org/std/fs/fn.read_dir.html).

bcmyers: Sounds like a good change to me. Will re-write to do just that.

src/bin/collapse-bpftrace.rs
bcmyers: Well, the scripts take different command line arguments (if we want to remain faithful to the perl), but I'll think about how to address this as well since it is sad to duplicate similar code.

bcmyers: I introduced an abstraction to get at this. Let me know if you think it was the wrong direction.

src/collapse/bpftrace.rs
bcmyers: The new version walks the directory so you don't have to hardcode the number of files. It's still based on pretty brittle file naming conventions, though, but I didn't see a great way around that. Let me know if you want to take it a different direction.

bcmyers:
I'll investigate why Windows is failing. I now have a Virtualbox setup for Windows after all that num-format system locale business...

src/collapse/mod.rs
```diff
@@ -1,2 +1,63 @@
...
+
+use std::io;
+use std::path::Path;
+use std::fs::File;
+
+const READER_CAPACITY: usize = 128 * 1024;
+
+/// Type that knows how to collapse formats from various frontends (e.g. bpftrace and perf)
+/// into output that can be subsequently ingested by `inferno-flamegraph`.
+pub struct Collapser {
```
jonhoo: I think this can probably be a trait instead:
```rust
trait Frontend {
    fn collapse<R: BufRead, W: Write>(r: R, w: W) -> io::Result<()>;
}
```
And then there'll be a free-standing function:
```rust
fn collapse_with<P, F>(infile: Option<P>, frontend: F) -> io::Result<()>
  where P: AsRef<Path>, F: Frontend {
    // ...
    frontend.handle(/* ... */)
    // ...
}
```
That way, each `main` just becomes something along the lines of:
```rust
collapser::collapse_with(infile.as_ref(), Perf::from(Opt::from_args()));
```
It'll require a little bit of re-organization in the perf module to have a `Perf` that implements `From<perf::Options>`, but that shouldn't be too bad.

What do you think?

src/collapse/bpftrace.rs
jonhoo: Here's what I think I'd do: name the input files `XXX.trace`, and the corresponding output file `XXX.orig.folded`. Then, have the code just iterate over all files `f` in `bpftrace/` whose `.extension()` is `trace`. For each file, compare the collapsed output to that contained in the file `f.file_stem().with_extension("orig.folded")`.

src/collapse/mod.rs
bcmyers: Yes. I like that. Wasn't sure how much you wanted me to mess with the existing perf code, but I like the idea of turning frontends into structs that all implement common behavior (i.e. a `Frontend` trait). Not sure `collapse_with` needs to be a standalone function (seems like it could be an autoimplemented method on the trait itself), but I'll play around with it.

bcmyers: I like this architecture, although we may want to think of better names for `collapse` and `collapse_with`... It's a little non-obvious what the difference between the two is. I put that off just to get your feedback on this general structure. 

Also - I'm pretty sure the Windows issue is due to git changing to `\r\n` line endings for our data files on Windows. Let me know if you think otherwise, but since bpftrace only runs on windows, I think we should be ingesting and outputing files with unix-style line endings always, even on Windows. I think the perl script is always outputing `\n` line-endings, but, as I mentioned before, I don't really know perl.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Denote all files that are truly binary and should not be modified.
+tests/data/bpftrace/* binary
```
jonhoo: While we're at it, maybe denote this for the other binary files too?

src/bin/collapse-perf.rs
```diff
@@ -53,39 +53,29 @@ struct Opt {
...
-            handle_file(options, r, io::stdout().lock())
-        }
-        None => {
-            let stdin = io::stdin();
-            let r = BufReader::with_capacity(128 * 1024, stdin.lock());
-            handle_file(options, r, io::stdout().lock())
-        }
-    }
+    let opt = Opt::from_args();
+    let (infile, options) = opt.into_parts();
```
jonhoo:
```suggestion
    let (infile, options) = Opt::from_args().into_parts();
```

src/collapse/bpftrace.rs
```diff
@@ -0,0 +1,216 @@
+use std::io;
+use std::iter::Peekable;
+use std::mem;
+
+use crate::collapse::Frontend;
```
jonhoo:
```suggestion
use super::Frontend;
```

src/collapse/mod.rs
bcmyers: Also - Another unrelated thought as I was looking at the perf code. We're essentially interning (I think that's the right word) the `addr2line::Context`s because they're expensive to construct. I know this is generally a bad thing, but what do you think about making that HashMap a global variable via lazy_static (would probably have to be protected by a mutex, which is the downside). Benefit would be two-fold: 

* We could derive `Clone`, `Debug`, `Eq`, `PartialEq`, and `Hash` (I believe) for `Perf`, which are not strictly necessary but generally good to provide on public types if we can (who knows how people will want to use your library).
* If for some crazy reason people use multiple instances of `Perf` in their program, you'll save on constructing `addr2line::Context`s.

src/collapse/bpftrace.rs
```diff
@@ -2,103 +2,123 @@ use std::io;
...
-    InStack(Vec<String>),
+use crate::collapse::Frontend;
+
+#[derive(Clone, Debug, Default, Eq, PartialEq, Hash)]
+pub struct Bpftrace {
+    state: State,
+}
+
+impl Bpftrace {
+    pub fn new() -> Self {
```
jonhoo: I'd probably cut this and just have `Default`

src/collapse/mod.rs
```diff
@@ -1,63 +1,39 @@
-pub mod bpftrace;
-pub mod perf;
+mod bpftrace;
+mod perf;
 
+pub use bpftrace::Bpftrace;
+pub use perf::{Perf, PerfOptions};
```
jonhoo: I think I prefer these to be accessed through the `mod`s (so leave them as `pub`) rather than re-exporting them here. That way we avoid the number of types swelling as we add more collapsers.

.gitattributes
bcmyers: 👍

src/bin/collapse-perf.rs
bcmyers: 👍

src/collapse/bpftrace.rs
bcmyers: 👍

src/collapse/bpftrace.rs
bcmyers: 👍

src/collapse/mod.rs
bcmyers: 👍

src/collapse/mod.rs
```diff
@@ -1,63 +1,39 @@
...
-pub struct Collapser {
-    frontend: Frontend,
-}
-
-impl Collapser {
-    /// Constructs a new `Collapser`.
-    pub fn new(frontend: Frontend) -> Self {
-        Collapser { frontend }
-    }
+pub trait Frontend {
```
jonhoo: I'd like to see documentation on this trait and its methods, since authors of new collapsers will have to implement it :)

src/collapse/perf.rs
```diff
@@ -5,11 +5,13 @@ use std::fs::File;
 use std::io;
 use std::io::prelude::*;
 
+use crate::collapse::Frontend;
```
```suggestion
use super::Frontend;
```

src/collapse/perf.rs
```diff
@@ -88,28 +72,54 @@ struct PerfState {
...
+        }
+        self.finish(writer)
+    }
 }
 
-impl From<Options> for PerfState {
-    fn from(opt: Options) -> Self {
-        PerfState {
+impl Perf {
+    pub fn new(options: PerfOptions) -> Self {
```
```suggestion
    pub fn from_options(options: PerfOptions) -> Self {
```

src/collapse/mod.rs
bcmyers: 👍

src/collapse/perf.rs
bcmyers: 👍

src/collapse/perf.rs
bcmyers: 👍

src/collapse/perf.rs
```diff
@@ -5,11 +5,13 @@ use std::fs::File;
...
 
+use crate::collapse::Frontend;
+
 const TIDY_GENERIC: bool = true;
 const TIDY_JAVA: bool = true;
 
-#[derive(Debug, Default)]
-pub struct Options {
+#[derive(Clone, Debug, Default)]
+pub struct PerfOptions {
```
jonhoo: I think this should continue to just be named `Options` since it'll only ever be accessed through the `perf` module.

bcmyers: 👍 - With the module change, makes sense!

src/collapse/perf.rs
jonhoo: And while we're at it, maybe rename the argument to just `opt` so you can use it directly in the constructor below.

bcmyers: 👍

src/collapse/mod.rs
jonhoo: > we may want to think of better names for `collapse` and `collapse_with`

Mmm, good point... How about `collapse` and `collapse_file`?

> since bpftrace only runs on windows

I'm confused? `bpftrace` is mostly a Linux tool as far as I'm aware? Is that what you meant?

> I think we should be ingesting and outputing files with unix-style line endings always, even on Windows.

Yes, I think I agree with this.

> We're essentially interning the `addr2line::Context`s because they're expensive to construct. [...] [W]hat do you think about making that HashMap a global variable via lazy_static.

Sadly this doesn't quite work, since you could have multiple files with the same module names but different `Context`s. If you share the map across multiple `Perf`s, that would mess up your inlining info. I'm not entirely sure that there's an advantage to deriving those types on `Perf`. It doesn't really make sense to do any of those operations on them I think, in part especially because they also need access to things like debug symbols!

bcmyers: "since bpftrace only runs on windows" - Yah, I meant linux.

bcmyers:
Think I got all the comments. Let me know what you think.

src/collapse/bpftrace.rs
```diff
@@ -1,20 +1,21 @@
+//! Module containting the **bpftrace** implementation of [`Frontend`].
+//!
+//! [`Frontend`]: trait.Frontend.html
```
jonhoo: At this point I've stopped including links like this since RFC 1964 [is stabilizing soon](https://github.com/rust-lang/rust/issues/43466) and will entirely obviate the need for them!

bcmyers: Yah - The current way you have to do it is totally broken. I'll pull all the internal links out. Leave the external links, though (e.g. for io::Error or for the webpages of bpftrace and perf)?

src/collapse/bpftrace.rs
```diff
@@ -1,20 +1,21 @@
+//! Module containting the **bpftrace** implementation of [`Frontend`].
```
jonhoo: I almost wonder if we should _not_ document the module at this point, because otherwise we might not realize that we need to expand the docs when we later add `#![deny(missing_docs)]` to the crate.

src/collapse/mod.rs
```diff
@@ -1,22 +1,46 @@
...
+pub mod bpftrace;
+pub mod perf;
 
 use std::fs::File;
 use std::io;
 use std::path::Path;
 
 const READER_CAPACITY: usize = 128 * 1024;
 
+/// Trait representing the ability to collapse the output of various profilers, such as [bpftrace]
```
jonhoo: Let's make this more active-sounding: "This trait represents ..."

src/collapse/bpftrace.rs
```diff
@@ -0,0 +1,213 @@
...
+        let mut number_buffer: [u8; 39] = unsafe { mem::uninitialized() };
+
+        // Iterate over each line
+        for line in reader.lines() {
+            let line = line?;
+
+            let mut is_beginning_of_line = true;
+
+            // Iterate over all characters in the line
+            let mut chars = line.chars().peekable();
```
jonhoo: Sorry, I only got around to reading the actual implementation now!

I think I'd like to see this be less character-oriented. Specifically, I think we can:

 - Use `==` for `@[`
 - Use `.starts_with("]: ")` for `]: 123901238`, and then use `rsplit_whitespace().next()` to get the number

I also think you're missing a `.trim()` for when you're extracting the function name for each stack?

src/collapse/bpftrace.rs
```diff
@@ -0,0 +1,213 @@
...
+            .join("tests")
+            .join("data")
+            .join("bpftrace");
+
+        for entry in fs::read_dir(&data_dir).unwrap() {
+            let path = entry.unwrap().path();
+            if let Some(extension) = path.extension() {
+                if extension.to_str().unwrap() == "trace" {
+                    let inpath = path;
+                    let outpath = inpath.with_extension("orig.folded");
```
jonhoo: I may be missing something here, but I thought the intention was that the `orig.folded` files were generated by the upstream `stackcollapse-bpftrace`? Looking at them more closely, this doesn't seem to be the case? I think we specifically want to compare against the "true" results from the Perl version to make sure we conform correctly :)

bcmyers: The `orig.folded` files are generated by the perl implementation (`stackcollapse-bpftrace.pl`). This is also relevant for your `trim()` comment above.

src/collapse/bpftrace.rs
bcmyers: 👍

src/collapse/mod.rs
bcmyers: 👍

src/collapse/bpftrace.rs
bcmyers: The advantage of a character implementation is where sure to look at each character only once. Wouldn't deviating from that make it less efficient (although perhaps simpler)? Happy to do whatever you'd like, of course.

src/collapse/bpftrace.rs
jonhoo: Wow... That's super weird.. I wonder how that works. Maybe they haven't noticed because the SVG collapses whitespace in strings? Judging from the example input they give at the top of `stackcollapse-bpftrace`, it looks like they just assume that there's no indentation in the stack lines.. Let me file an issue there and see what they say!

bcmyers: Even though the `orig.folded` files in the data directory *are* generated by the upstream perl implementation, we could if you'd like have the test run the perl script each time it runs. Let me know if you would prefer that. I originally wrote the tests that way, but then changed for no apparent reason. So just let me know which one you prefer.

jonhoo: Oooh, I like that. That'd be even better! It'd slow down the tests a little, but I think it's okay.

jonhoo: Also, I just filed https://github.com/brendangregg/FlameGraph/issues/201

src/collapse/bpftrace.rs
jonhoo: While it's true that it's a little more efficient, `==` and `.starts_with` are also pretty efficient, since they'll only look at at most as many characters as are in the pattern. Thus we'll only look at a very small number of characters twice! And it makes the code _much_ easier to follow (I think). So yes, if you have the time to change it, I'd appreciate that!

src/collapse/bpftrace.rs
bcmyers: 👍

src/collapse/bpftrace.rs
bcmyers: Begrudgingly 👍 :)

jonhoo:
Ah, @brendangregg just responded in https://github.com/brendangregg/FlameGraph/issues/201 and pointed out that they're actually adding direct support for the `perf` output format in `bpftrace` (in https://github.com/iovisor/bpftrace/pull/438), so maybe we simply _shouldn't_ make this change? I think it'd still be worthwhile to keep the trait changes though, as that'll make it much easier to adopt more formats later!

bcmyers:
Ok, so pull out all the bpftrace-related stuff, but keep our trait / architecture / documentation comments?

bcmyers:
I hope I just did that right. I still don't understand git very well.

jonhoo:
Looks perfect! I'm not sure you needed to do a force-push or rebase, but it's fine either way :) Let's see what Travis says, and then merge. Sorry for all the effort you had to go through :sweat_smile:

PR #58, opened 2019-02-22
Cache svg::Event across frames

meven:
Fixes for #41 
Based on https://github.com/JordiChauzi/inferno/pull/1

### DIFF
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -135,6 +135,9 @@ impl Rectangle {
     fn width(&self) -> usize {
         self.x2 - self.x1
     }
+    fn height(&self) -> usize {
+        self.y2 - self.y1
+    }
 }
 
 pub fn from_sorted_lines<'a, I, W>(opt: Options, lines: I, writer: W) -> quick_xml::Result<()>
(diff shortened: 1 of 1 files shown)


src/flamegraph/mod.rs
```diff
@@ -413,27 +437,30 @@ fn deannotate(f: &str) -> &str {
...
     color: (u8, u8, u8),
+    cache_rect: &mut Event,
 ) -> quick_xml::Result<usize> {
-    let x = write!(buffer, "{}", x1);
-    let y = write!(buffer, "{}", y1);
-    let width = write!(buffer, "{}", x2 - x1);
-    let height = write!(buffer, "{}", y2 - y1);
+    let x = write!(buffer, "{}", rect.x1);
+    let y = write!(buffer, "{}", rect.y1);
+    let width = write!(buffer, "{}", rect.x2 - rect.x1);
```
```suggestion
    let width = write!(buffer, "{}", rect.width());
```

src/flamegraph/mod.rs
```diff
@@ -413,27 +437,30 @@ fn deannotate(f: &str) -> &str {
...
+    let height = write!(buffer, "{}", rect.y2 - rect.y1);
```
jonhoo: Can we introduce `rect.height()` and use that here?

jonhoo:
This looks excellent once those two minor comments are addressed!

Issue #25, opened 2019-02-03
Flame graph differentials 


jonhoo:
In https://github.com/brendangregg/FlameGraph/commit/465ac0c46db04cba413ae593d473181d694ea681, @brendangregg added support for [differential flame graphs](http://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html), which can be really neat for profiling performance regressions. They have a couple of effects on [`flamegraph.pl`](https://github.com/brendangregg/FlameGraph/commit/465ac0c46db04cba413ae593d473181d694ea681#diff-ac3933459c240ef4c972dc5eeb7a15bf):

 - They require tracking "delta" values in both [`TimedFrame`](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L555L557) and [`Frame`](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L564L566) when we generate them in `flow`.
 - They mean that there may be an [_additional_ sample column](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L621L631) in each stack line.
 - They change the [status line](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L1089L1094) shown on hover for each frame.
 - And they change the [color palette](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L1111).

Issue #43, opened 2019-02-09
Upstream flamegraph doesn't discard fractional samples


zao:
I'm still watching part 2 but given the current state of the source code, I believe that you may have misinterpreted the part of the regex in upstream FlameGraph about removing the fractional part of the samples.

```perl
my($stack, $samples) = (/^(.*)\s+?(\d+(?:\.\d*)?)$/);
```

https://github.com/jonhoo/inferno/blob/ce1c41a650b8254027eda24461227beff4bb6913/src/flamegraph/merge.rs#L94-L98

My reading is that the non-capturing group is there to allow input in either integer or fractional format, but not in any way remove it from the surrounding capture. The non-capturing group is there to group up the separator and the fractions for the following `?`.

That is, it accepts sample values like `12`, `34.`, and `56.78`.

jonhoo:
Ah, yes, I think you're right! That would mean that the sample counts are all fractional (probably `f64`). Would you be willing to write up a PR with that change?

zao:
I don't think I've got the mental bandwidth currently to fully implement this as it propagates throughout the program when you make the samples floating point.

In particular, it affects the lookups of `TimedFrame` due to `f64` not having an ordering for `Eq` nor having a `Hash`.

I'd gladly leave it for another contributor or for a stream where you get to explain the Fun in why floats don't have decent orderings.

jonhoo:
One trick we could play here is to upsample by, say, 100x, and then keep using `usize` :thinking:

jasonrhansen:
As far as I can tell there is no reason `TimedFrame` needs to be `Eq` or `Hash`. `Frame` is used as a key in a `HashMap`, but it doesn't have any fields relating to samples. `TimedFrame` is only ever stored in a `Vec`. I think we should be able to use `f64` for `start_time`, `end_time`, and `delta`, or am I overlooking something?

jonhoo:
I did some digging on this in https://github.com/jonhoo/inferno/pull/64#issuecomment-467992884, and my conclusion from reading https://github.com/brendangregg/FlameGraph/pull/18 is that we should probably simply discard the fractional parts of samples, and add support for [`--factor`](https://github.com/brendangregg/FlameGraph/blob/1b1c6deede9c33c5134c920bdb7a44cc5528e9a7/flamegraph.pl#L113) instead.

jasonrhansen:
#65 adds --factor instead.

PR #60, opened 2019-02-25
Add support for differential flame graphs

jasonrhansen:
Implements #25

### DIFF
--- a/src/flamegraph/color/mod.rs
+++ b/src/flamegraph/color/mod.rs
@@ -255,6 +255,18 @@ pub(super) fn color(
     rgb_components_for_palette(palette, name, v1, v2, v3)
 }
 
+pub(super) fn color_scale(value: isize, max: usize) -> (u8, u8, u8) {
+    if value == 0 {
+        (255, 255, 255)
+    } else if value > 0 {
+        let c = (210 * (max as isize - value) / max as isize) as u8;
+        (255, c, c)
+    } else {
+        let c = (210 * (max as isize + value) / max as isize) as u8;
+        (c, c, 255)
+    }
+}
+
 fn default_bg_color_for(palette: Palette) -> BackgroundColor {
     match palette {
         Palette::Basic(BasicPalette::Mem) => BackgroundColor::Green,
--- a/src/flamegraph/color/mod.rs
+++ b/src/flamegraph/color/mod.rs
@@ -259,9 +259,13 @@ pub(super) fn color_scale(value: isize, max: usize) -> (u8, u8, u8) {
     if value == 0 {
         (255, 255, 255)
     } else if value > 0 {
+        // A positive value indicates _more_ samples,
+        // and hence more time spent, so we give it a red hue.
         let c = (210 * (max as isize - value) / max as isize) as u8;
         (255, c, c)
     } else {
+        // A negative value indicates _fewer_ samples,
+        // or a speed-up, so we give it a green hue.
         let c = (210 * (max as isize + value) / max as isize) as u8;
         (c, c, 255)
     }
--- a/src/flamegraph/merge.rs
+++ b/src/flamegraph/merge.rs
@@ -181,21 +181,17 @@ where
 
 // Parse and remove the number of samples from the end of a line.
 fn parse_nsamples(line: &mut &str) -> Option<usize> {
-    if let Some(samplesi) = line.rfind(' ') {
-        let mut samples = &line[(samplesi + 1)..];
-        // strip fractional part (if any);
-        // foobar 1.klwdjlakdj
-        // TODO: Properly handle fractional samples (see issue #43)
-        if let Some(doti) = samples.find('.') {
-            samples = &samples[..doti];
-        }
-        if let Ok(nsamples) = samples.parse::<usize>() {
-            // remove nsamples part we just parsed from line
-            *line = line[..samplesi].trim_end();
-            // give out the sample count
-            return Some(nsamples);
-        }
+    let samplesi = line.rfind(' ')?;
+    let mut samples = &line[(samplesi + 1)..];
+    // strip fractional part (if any);
+    // foobar 1.klwdjlakdj
+    // TODO: Properly handle fractional samples (see issue #43)
+    if let Some(doti) = samples.find('.') {
+        samples = &samples[..doti];
     }
 
-    None
+    let nsamples = samples.parse::<usize>().ok()?;
+    // remove nsamples part we just parsed from line
+    *line = line[..samplesi].trim_end();
+    Some(nsamples)
 }
--- a/src/flamegraph/merge.rs
+++ b/src/flamegraph/merge.rs
@@ -118,22 +118,21 @@ where
             continue;
         }
 
+        // Parse the number of samples for the purpose of computing overall time passed.
         // Usually there will only be one samples column at the end of a line,
-        // but for differentials there will be two. In the case there are two
-        // we compute the delta between them and use the second one as as the
-        // number of samples for this frame.
-        let nsamples = match (parse_nsamples(&mut line), parse_nsamples(&mut line)) {
-            (Some(s2), nsamples1) => {
-                delta = nsamples1.and_then(|s1| Some(s2 as isize - s1 as isize));
-                if let Some(delta) = delta {
-                    delta_max = std::cmp::max(delta.abs() as usize, delta_max);
-                }
-                s2
+        // but for differentials there will be two. When there are two we compute the
+        // delta between them and use the second one.
+        let nsamples = if let Some(s2) = parse_nsamples(&mut line) {
+            // See if there's also a differential column present
+            let nsamples1 = parse_nsamples(&mut line);
+            delta = nsamples1.and_then(|s1| Some(s2 as isize - s1 as isize));
+            if let Some(delta) = delta {
+                delta_max = std::cmp::max(delta.abs() as usize, delta_max);
             }
-            _ => {
-                ignored += 1;
-                continue;
-            }
+            s2
+        } else {
+            ignored += 1;
+            continue;
         };
 
         if line.is_empty() {
--- a/src/flamegraph/merge.rs
+++ b/src/flamegraph/merge.rs
@@ -124,7 +124,7 @@ where
         // number of samples for this frame.
         let nsamples = match (parse_nsamples(&mut line), parse_nsamples(&mut line)) {
             (Some(s2), nsamples1) => {
-                delta = nsamples1.map_or(None, |s1| Some(s2 as isize - s1 as isize));
+                delta = nsamples1.and_then(|s1| Some(s2 as isize - s1 as isize));
                 if let Some(delta) = delta {
                     delta_max = std::cmp::max(delta.abs() as usize, delta_max);
                 }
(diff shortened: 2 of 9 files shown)


jonhoo:
Looks like `clippy` is [complaining](https://travis-ci.org/jonhoo/inferno/jobs/497965463#L538).

src/flamegraph/color/mod.rs
```diff
@@ -255,6 +255,18 @@ pub(super) fn color(
     rgb_components_for_palette(palette, name, v1, v2, v3)
 }
 
+pub(super) fn color_scale(value: isize, max: usize) -> (u8, u8, u8) {
+    if value == 0 {
+        (255, 255, 255)
+    } else if value > 0 {
+        let c = (210 * (max as isize - value) / max as isize) as u8;
```
jonhoo: I'd like to see a comment here along the lines of:
```rust
// A positive value indicates _more_ samples, and hence more time spent, so we give it a red hue.
```

src/flamegraph/color/mod.rs
```diff
@@ -255,6 +255,18 @@ pub(super) fn color(
...
+        (255, c, c)
+    } else {
+        let c = (210 * (max as isize + value) / max as isize) as u8;
```
jonhoo: And here:
```rust
// A negative value indicates _fewer_ samples, or a speed-up, so we give it a green hue.
```

jasonrhansen:
The first `clippy` fix is simple, but it's complaining about the complexity of `from_sorted_lines`. How do you want to break that up?

src/flamegraph/merge.rs
```diff
@@ -50,31 +58,50 @@ fn flow<'a, LI, TI>(
...
         };
+
+        let is_last = this.peek().is_none();
+        let delta = match delta {
+            Some(_) if !is_last => Some(0),
+            d => d,
+        };
+        let frame_time = FrameTime {
+            start_time: time,
+            delta,
```
jonhoo: I find it a little weird that the Perl version [does a `+=` for `delta`](https://github.com/brendangregg/FlameGraph/blob/1b1c6deede9c33c5134c920bdb7a44cc5528e9a7/flamegraph.pl#L588)...

jasonrhansen: I also thought that was strange, so I did what made sense to me.

src/flamegraph/merge.rs
```diff
@@ -148,8 +172,29 @@ where
...
-    (frames, time, ignored)
+    (frames, time, ignored, delta_max)
+}
+
+// Parse and remove the number of samples from the end of a line.
+fn parse_nsamples(line: &mut &str) -> Option<usize> {
+    if let Some(samplesi) = line.rfind(' ') {
+        let mut samples = &line[(samplesi + 1)..];
+        // strip fractional part (if any);
+        // foobar 1.klwdjlakdj
```
jonhoo: While we're touching this code, would you mind adding a reference to https://github.com/jonhoo/inferno/issues/43 here?

src/flamegraph/merge.rs
```diff
@@ -148,8 +172,29 @@ where
...
         );
     }
 
-    (frames, time, ignored)
+    (frames, time, ignored, delta_max)
+}
+
+// Parse and remove the number of samples from the end of a line.
+fn parse_nsamples(line: &mut &str) -> Option<usize> {
+    if let Some(samplesi) = line.rfind(' ') {
```
jonhoo: You can take advantage of the fact that `Option` implements [`Try`](https://doc.rust-lang.org/std/ops/trait.Try.html) here! Try:
```rust
let samplesi = line.rfind(' ')?;
```
You can probably use that to simplify the code further down too!

src/flamegraph/merge.rs
```diff
@@ -148,8 +172,29 @@ where
...
+        let mut samples = &line[(samplesi + 1)..];
+        // strip fractional part (if any);
+        // foobar 1.klwdjlakdj
+        if let Some(doti) = samples.find('.') {
+            samples = &samples[..doti];
+        }
+        if let Ok(nsamples) = samples.parse::<usize>() {
```
jonhoo: This can also become an `Option` using `samples.parse::<usize>().ok()`, at which point you can use `?` for this too!

src/flamegraph/merge.rs
```diff
@@ -83,34 +110,30 @@ where
...
-                    nsamples
-                }
-                Err(_) => {
-                    ignored += 1;
-                    continue;
+        // Usually there will only be one samples column at the end of a line,
+        // but for differentials there will be two. In the case there are two
+        // we compute the delta between them and use the second one as as the
+        // number of samples for this frame.
+        let nsamples = match (parse_nsamples(&mut line), parse_nsamples(&mut line)) {
```
jonhoo: I think this would be more readable as:
```rust
let nsamples = if let Some(s2) = parse_nsamples(&mut line) {
    let nsamples1 = parse_nsamples(&mut line);
    // ...
} else {
    ignored += 1;
    continue;
}
```

src/flamegraph/merge.rs
```diff
@@ -83,34 +110,30 @@ where
...
+            (Some(s2), nsamples1) => {
+                delta = nsamples1.map_or(None, |s1| Some(s2 as isize - s1 as isize));
```
jonhoo: Why isn't this just a `map`?

src/flamegraph/merge.rs
```diff
@@ -83,34 +110,30 @@ where
...
+                if let Some(delta) = delta {
```
jonhoo: Given that we have to do this `if let Some` anyway, I think I'd prefer to get rid of the `map` on the line above and instead just do the computation of `delta` inside the `if` too.

src/flamegraph/merge.rs
```diff
@@ -83,34 +110,30 @@ where
...
-                    // give out the sample count
-                    nsamples
-                }
-                Err(_) => {
-                    ignored += 1;
-                    continue;
+        // Usually there will only be one samples column at the end of a line,
+        // but for differentials there will be two. In the case there are two
+        // we compute the delta between them and use the second one as as the
+        // number of samples for this frame.
```
jonhoo: maybe add to this "for the purposes of computing overall time passed", since as far as I can tell that's the only place the sample count is used.

src/flamegraph/mod.rs
```diff
@@ -41,6 +41,22 @@ pub struct Options {
...
+    fn default() -> Self {
+        Options {
+            colors: Default::default(),
+            bgcolors: Default::default(),
+            hash: Default::default(),
+            consistent_palette: Default::default(),
+            func_frameattrs: Default::default(),
+            direction: Default::default(),
+            title: "Flame Graph".to_string(),
+            negate_differentials: Default::default(),
```
jonhoo: I think you can just use:
```rust
Options {
    title: "Flame Graph".to_string(),
    ..Default::default(),
}
```

src/flamegraph/mod.rs
```diff
@@ -254,15 +270,37 @@ where
...
-                "{} ({} samples, {:.2}%)",
-                deannotate(&frame.location.function),
-                samples_txt,
-                pct
-            )
+            match frame.delta {
+                None => write!(
+                    buffer,
+                    "{} ({} samples, {:.2}%)",
+                    deannotate(&frame.location.function),
```
jonhoo: Maybe hoist out `deannotate(&frame.location.function)` since it's used by all the arms?

src/flamegraph/mod.rs
```diff
@@ -254,15 +270,37 @@ where
...
+                    samples_txt,
+                    pct
+                ),
+                // Special case delta == 0 so we don't format percentage with a + sign.
```
jonhoo: Bah, that's annoying. But seems reasonable.

src/flamegraph/mod.rs
jasonrhansen: I actually did it that way in the tests. I have no idea why I didn't do that here.

src/flamegraph/merge.rs
jonhoo: Maybe leave a comment with a link to the Perl version?

jonhoo:
Re `from_sorted_lines`, ultimately I want the code that draws each frame to be in its own function, but #52 already makes some headroom we can build on there. For the time being, just add
```rust
```

jonhoo:
Also, awesome work! :heart:

src/flamegraph/mod.rs
jasonrhansen: OK, now I know why I did that. Using Default::default() here would cause an infinite recursion.

src/flamegraph/mod.rs
jasonrhansen: I wish there was a way to format with + and -, except when 0.

src/flamegraph/mod.rs
jonhoo: Ohh, yeah, I guess that makes sense. Too bad..

Maybe move the ones you are actually setting a value for further up, and all the `Default::default()`s to the bottom?

src/flamegraph/merge.rs
```diff
@@ -181,21 +181,17 @@ where
...
-        }
-        if let Ok(nsamples) = samples.parse::<usize>() {
-            // remove nsamples part we just parsed from line
-            *line = line[..samplesi].trim_end();
-            // give out the sample count
-            return Some(nsamples);
-        }
+    let samplesi = line.rfind(' ')?;
+    let mut samples = &line[(samplesi + 1)..];
+    // strip fractional part (if any);
```
jonhoo: Really nitpicky, but can we add an empty line above here?

src/flamegraph/merge.rs
```diff
@@ -125,9 +125,9 @@ where
         let nsamples = if let Some(s2) = parse_nsamples(&mut line) {
             // See if there's also a differential column present
             let nsamples1 = parse_nsamples(&mut line);
-            delta = nsamples1.and_then(|s1| Some(s2 as isize - s1 as isize));
-            if let Some(delta) = delta {
-                delta_max = std::cmp::max(delta.abs() as usize, delta_max);
+            if let Some(s1) = nsamples1 {
```
jonhoo: At this point you might as well get rid of the `nsamples1` binding too.

src/flamegraph/merge.rs
```diff
@@ -125,9 +125,9 @@ where
         let nsamples = if let Some(s2) = parse_nsamples(&mut line) {
```
jonhoo: I wonder if we can give this a more descriptive name than `s2`?

src/flamegraph/merge.rs
jasonrhansen: I'm ok with being nitpicky.

src/flamegraph/merge.rs
jasonrhansen: How about we change `s2` to just `samples`. Then if there's a differential column, we can use something like `samples_first` instead of `s1`. Or do you have a better suggestion?

jonhoo: I like that! Maybe `original_samples` for `s1`?

jasonrhansen: Sure

Issue #30, opened 2019-02-03
Tidy up library interface for flamegraph (and collapse?)


jonhoo:
If we want other projects, like @koute's [nperf](https://github.com/nokia/nperf), to start using inferno to draw their flame graphs (and perhaps even to do the collapsing), we need to provide an external interface that isn't just a purely line-based textual one like the ones used internally between `perf script`, `stackcollapse-perf`, and `flamegraph.pl`. Specifically, we'll want an interface that is typed, and where all (most?) of the parsing is done by the caller. I'm imagining something like:

```rust
mod flamegraph {
  struct StackCount<S> {
    stack: S,
    count: usize,
  }
    
  fn from_samples<I, S, W>(o: Opts, samples: I, writer: W) -> 
    where
      I: IntoIterator,
      I::Item: Into<StackCount<S>>,
      S: Iterator<Item = &str>,
      W: Write
  {
  }
}
```
`from_samples` would require that the input is given in sorted order (see #28), and then write a flame graph output to `writer`. The existing code that does parsing would call from_samples with an iterator created from the lines iterator, mapped something like this:
```rust
lines.map(|line| {
  let (stack, nsamples) = ...;
  StackCount {
    count: nsamples,
    stack: stack.split(';'),
  }
})
```
@koute: what do you think of an interface roughly like the above? You think that's an iterator that `nperf` could easily produce? What would be a good interface for `collapse` (if that's even something you might consider using)?

koute:
Sounds good to me; I should be able to produce such iterator easily.

I wonder though if in the long run an interface which has collapsing integrated wouldn't be better - the extra integration could probably me made more performant, and it would be potentially simpler to use.

Something like this maybe:

```rust
struct FrameBuilder<T> {}
impl FlameBuilder<T> where T: PartialEq + Eq + PartialOrd + Ord + ... {
    fn new() -> Self {}
    fn add_stack_trace<I>(&mut self, I)
        where I: IntoIterator<Item = T> {}
    fn generate<F>(self, output: F, write_frame: W) -> Result<()>
        where F: Write, W: FnMut(&T, &mut F) -> Result<()> {}
}
```

The `write_frame` is there since the `T`s might not be strings, e.g. in my case they wouln't be printable by themselves as they contain IDs to interned strings, hence the customizable writer.

jonhoo:
[...]

PR #61, opened 2019-02-25
Give docs a proper update

jonhoo:
I've made a pass on pretty much all the docs, as well as the names of a few types, to try to clean up the externally-visible API. It'd be awesome if someone could give it all a read through and see that a) the structure of the API makes sense, b) I'm not saying anything totally false, and c) there aren't any gaping holes that we should fill immediately.

I have included any example code somewhat intentionally since I'm guessing the external APIs are likely to change a decent amount with #30.

### DIFF
--- a/src/flamegraph/attrs.rs
+++ b/src/flamegraph/attrs.rs
@@ -18,17 +18,19 @@ macro_rules! unwrap_or_continue {
 pub struct FuncFrameAttrsMap(HashMap<String, FrameAttrs>);
 
 impl FuncFrameAttrsMap {
-    /// Parse a FuncFrameAttrsMap from a file.
-    /// Each line should be a function name followed by a tab,
-    /// then a sequence of tab separated name=value pairs.
+    /// Parse frame attribtues from a file.
+    ///
+    /// Each line should consist of a function name, a tab (`\t`), and then a sequence of
+    /// tab-separated `name=value` pairs.
     pub fn from_file(path: &PathBuf) -> io::Result<FuncFrameAttrsMap> {
         let file = BufReader::new(File::open(path)?);
         FuncFrameAttrsMap::from_reader(file)
     }
 
-    /// Parse a FuncFrameAttrsMap from a reader.
-    /// Each line should be a function name followed by a tab,
-    /// then a sequence of tab separated name=value pairs.
+    /// Parse frame attribtues from a `BufRead`.
+    ///
+    /// Each line should consist of a function name, a tab (`\t`), and then a sequence of
+    /// tab-separated `name=value` pairs.
     pub fn from_reader<R: BufRead>(mut reader: R) -> io::Result<FuncFrameAttrsMap> {
         let mut funcattr_map = FuncFrameAttrsMap::default();
         let mut line = String::new();
@@ -77,56 +79,56 @@ impl FuncFrameAttrsMap {
     }
 
     /// Return FrameAttrs for the given function name if it exists
-    pub fn frameattrs_for_func(&self, func: &str) -> Option<&FrameAttrs> {
+    pub(super) fn frameattrs_for_func(&self, func: &str) -> Option<&FrameAttrs> {
         self.0.get(func)
     }
 }
 
 /// Attributes to set on the SVG elements of a frame
 #[derive(PartialEq, Eq, Debug, Default)]
-pub struct FrameAttrs {
+pub(super) struct FrameAttrs {
     /// The text to include in the `title` element.
     /// If set to None, the title is dynamically generated based on the function name.
-    pub title: Option<String>,
+    pub(super) title: Option<String>,
 
-    pub g: GElementAttrs,
-    pub a: AElementAttrs,
+    pub(super) g: GElementAttrs,
+    pub(super) a: AElementAttrs,
 }
 
 /// Attributes to set on the SVG `g` element.
 /// Any of them set to `None` will get the default value.
 #[derive(PartialEq, Eq, Debug, Default)]
-pub struct GElementAttrs {
+pub(super) struct GElementAttrs {
     /// Defaults to "func_g"
-    pub class: Option<String>,
+    pub(super) class: Option<String>,
 
     /// Will not be included if None
-    pub style: Option<String>,
+    pub(super) style: Option<String>,
 
     /// Defaults to "s(this)"
-    pub onmouseover: Option<String>,
+    pub(super) onmouseover: Option<String>,
 
     /// Defaults to "c()"
-    pub onmouseout: Option<String>,
+    pub(super) onmouseout: Option<String>,
 
     /// Defaults to "zoom(this)"
-    pub onclick: Option<String>,
+    pub(super) onclick: Option<String>,
 
     /// Extra attributes to include
-    pub extra: Vec<(String, String)>,
+    pub(super) extra: Vec<(String, String)>,
 }
 
 /// Attributes to set on the SVG `a` element
 #[derive(PartialEq, Eq, Debug, Default)]
-pub struct AElementAttrs {
+pub(super) struct AElementAttrs {
     /// If set to None the `a` tag will not be added
-    pub href: Option<String>,
+    pub(super) href: Option<String>,
 
     /// Defaults to "_top"
-    pub target: Option<String>,
+    pub(super) target: Option<String>,
 
     /// Extra attributes to include
-    pub extra: Vec<(String, String)>,
+    pub(super) extra: Vec<(String, String)>,
 }
 
 fn parse_extra_attrs(attrs: &mut Vec<(String, String)>, s: &str) {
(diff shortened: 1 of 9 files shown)


src/flamegraph/attrs.rs
```diff
@@ -18,17 +18,19 @@ macro_rules! unwrap_or_continue {
 pub struct FuncFrameAttrsMap(HashMap<String, FrameAttrs>);
 
 impl FuncFrameAttrsMap {
-    /// Parse a FuncFrameAttrsMap from a file.
-    /// Each line should be a function name followed by a tab,
-    /// then a sequence of tab separated name=value pairs.
+    /// Parse frame attribtues from a file.
```
jasonrhansen: Typo: attribtues

src/flamegraph/attrs.rs
```diff
@@ -18,17 +18,19 @@ macro_rules! unwrap_or_continue {
...
+    /// tab-separated `name=value` pairs.
     pub fn from_file(path: &PathBuf) -> io::Result<FuncFrameAttrsMap> {
         let file = BufReader::new(File::open(path)?);
         FuncFrameAttrsMap::from_reader(file)
     }
 
-    /// Parse a FuncFrameAttrsMap from a reader.
-    /// Each line should be a function name followed by a tab,
-    /// then a sequence of tab separated name=value pairs.
+    /// Parse frame attribtues from a `BufRead`.
```
jasonrhansen: Same typo

src/flamegraph/mod.rs
```diff
@@ -32,23 +28,67 @@ const XPAD: usize = 10; // pad lefm and right
...
+    /// Choose names based on the hashes of function names.
+    ///
+    /// This will cause similar functions to be colored similarly.
     pub hash: bool,
+
+    /// Store the choice of color for each function so that later invocations use the same colors.
+    ///
+    /// With this option enabled, a file called `palette.map` will be created the first time a
+    /// flame graph is generated, and the (random) color chosen for each function will be written
+    /// into it. On subsequent invocations, functions that already have a colored registered in
```
jasonrhansen: Change "colored" to "color".

src/flamegraph/mod.rs
```diff
@@ -410,6 +463,12 @@ where
     from_sorted_lines(opt, input.lines(), writer)
 }
 
+/// Produce a flame graph from a set of reader that contain folded stack lines.
```
jasonrhansen: Change to "set of readers"

src/lib.rs
```diff
@@ -1,9 +1,109 @@
...
+//! spending its time by looking at which functions appear in many of the traces. To ease this
+//! analysis, we want to "collapse" the stack traces so if a particular trace occurs more than
+//! once, we instead just keep it _once_ along with a count of how many times we've seen it. This
+//! is what the various collapsing tools do! You'll sometimes see the resulting tuples of stack +
+//! count called a "folded stack trace".
+//!
+//! Since profiling tools produce stack traces in a myriad of different formats, and the flame
+//! graph plotter expects input in a particular folded stack trace format, each profiler needs a
+//! separate collapse implementation. While the original Perl implementation supports _lots_ of
+//! profilers, Inferno currenly only supports one: the widely used [`perf`] tool (specifically the
```
jasonrhansen: Change "currenly" to "currently".

jonhoo:
@jasonrhansen great, thanks for the review! I'll leave this open for a bit longer in case @JordiChauzi also has a chance to take a look, otherwise I'll merge towards the end of today EST.

src/flamegraph/color/mod.rs
```diff
@@ -16,49 +16,88 @@ const BLUE_GRADIENT: (&str, &str) = ("#eeeeee", "#e0e0ff");
...
+/// different function names (kernel functions, JIT functions, etc.).
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum MultiPalette {
+    /// Use Java semantics to color frames.
     Java,
+    /// Use JavaScript semantics to color frames.
     Js,
+    /// Use Perl semantics to color frames.
     Perl,
+    /// TODO: this seems to just be [`BasicPalette::Aqua`]?
```
JordiChauzi: The difference between `BasicPalette::Aqua` and `MultiPalette::Wakeup` is the default background gradient (`Grey` for `Aqua`, `Blue` for `Wakeup`).

src/flamegraph/mod.rs
```diff
@@ -32,23 +28,67 @@ const XPAD: usize = 10; // pad lefm and right
...
+
+    /// Choose names based on the hashes of function names.
+    ///
+    /// This will cause similar functions to be colored similarly.
     pub hash: bool,
+
+    /// Store the choice of color for each function so that later invocations use the same colors.
+    ///
+    /// With this option enabled, a file called `palette.map` will be created the first time a
+    /// flame graph is generated, and the (random) color chosen for each function will be written
```
JordiChauzi: Do we need to precise random ? With `--hash` the color is always deterministic.

src/flamegraph/mod.rs
```diff
@@ -32,23 +28,67 @@ const XPAD: usize = 10; // pad lefm and right
...
+    /// The title for the flame chart.
     pub title: String,
 }
 
+/// The direction the plot should grow.
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub enum Direction {
+    /// Stacks grow from the bottom to the top.
+    ///
+    /// `main` will be at the bottom.
```
JordiChauzi: `main` might be too language specific, and even for language with `main`, it is possible to have a stack with no `main` (launching a `Thread` in java for example). Maybe change to `entry point` ?

JordiChauzi:
For different structures implementing `Default`, you precised the default value in the doc of the structure. Hopefully the value will not change overtime, but is it not redundant ?

jonhoo:
I think it's useful to spell out the defaults in the docs since they're not otherwise visible anywhere beyond digging into the code. It's true that we'd have to make sure update the docs if the defaults change, but at the same time I think these defaults should also be relatively unlikely to change :)

jonhoo:
Thanks for the review from both of you! I'll merge once CI passes!

jonhoo:
@jasonrhansen I added some more docs following #60 in 3e706ea6ee8ceaca22f716e0bec311020d1fb0e6. Can you take a quick look?

src/flamegraph/color/mod.rs
```diff
@@ -16,49 +16,98 @@ const BLUE_GRADIENT: (&str, &str) = ("#eeeeee", "#e0e0ff");
...
+}
+
+/// A plain color palette in which the color is independent of the function name.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum BasicPalette {
+    /// A palette in which colors are chosen randomly from a red-yellow spectrum.
     Hot,
+    /// A palette in which colors are chosen randomly from a green-blue spectrum.
+    ///
+    /// Defaults to
```
JordiChauzi: Is this line correct ?

jonhoo: Nope! Good catch!

src/flamegraph/color/mod.rs
```diff
@@ -16,49 +16,98 @@ const BLUE_GRADIENT: (&str, &str) = ("#eeeeee", "#e0e0ff");
...
     Multi(MultiPalette),
 }
 
+impl Default for Palette {
+    fn default() -> Self {
+        Palette::Basic(BasicPalette::Hot)
+    }
+}
+
+/// A plain color palette in which the color is independent of the function name.
```
JordiChauzi: Actually, the function name may or may not be used to compute the exact color used (depending on the `--hash` option)

jasonrhansen:
Thanks for adding docs for differentials! I made "Flame Graph" the default for the title in `Options`. We should probably document that now.

src/flamegraph/color/mod.rs
```diff
@@ -16,49 +16,98 @@ const BLUE_GRADIENT: (&str, &str) = ("#eeeeee", "#e0e0ff");
...
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum BasicPalette {
+    /// A palette in which colors are chosen randomly from a red-yellow spectrum.
```
JordiChauzi: `randomly` <- here again the random aspect is not bound to the palette (only the options `--hash` and/or `--cp`)

src/flamegraph/color/mod.rs
jonhoo: Good point. I've tried to address this better in 0676aef07fcb81e210bb143f9586619f14f734c3. What do you think?

jonhoo:
@jasonrhansen done in 924421bff10b4bb0e743bea9543e632e60547ea4

src/flamegraph/color/mod.rs
JordiChauzi: Seems good!

Issue #19, opened 2019-02-02
Support arbitrary scaling factor


jonhoo:
The [`--factor` option](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L114) in `flamegraph.pl` lets you arbitrarily [multiply all sample counts](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L1071) by a constant, probably to show "true counts" if you've undersampled. We should document and adopt that option.

jasonrhansen:
I did an initial implementation of this and realized that the output sometimes differs between my version and the Perl version because of differences in the way floating point numbers are rounded when formatting to a string compared to calling `round` on an `f64`.

`712.5f64.round() as usize` results in `713`, but `print!("{:.0}", 712.5);` prints `712`.

The Perl version formats the samples like this:
```Perl
my $samples = sprintf "%.0f", ($etime - $stime) * $factor;
```
But our version is using `num_format`, which doesn't support formatting floating point values. Currently I'm doing this before formatting:
```Rust
let samples = ((frame.end_time - frame.start_time) as f64 * opt.factor).round() as usize;
```

Do we want to come up with a way to get the sample counts to match the Perl version in every case, or should we not worry about it?

jonhoo:
I think we specifically need the result to _not_ be a `usize`, since `factor` may be something like `0.01`. @bcmyers any plans to add support for floating point numbers to `num-format`?

jasonrhansen:
The Perl version formats the samples in the end without the fractional part. Are you saying you want to diverge from this behavior? If so, how many decimal places would we want to show?

jonhoo:
Ohh, you're entirely right! I misread that part of the Perl code. I _think_ what we want here is `f64::floor` or `f64::trunc` (not sure how they're different), both of which would exactly match the Perl version. Let's continue discussion in #65.

bcmyers:
Yes - I had always planned to add floating point support to num-format, but it's a bit more complicated than integers. I'd like to read and (at least partially) understand some of the following kind of stuff before I get started (just a sample of various stuff on the topic that's out there):

- https://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf
- https://github.com/miloyip/dtoa-benchmark
- https://github.com/dtolnay/dtoa
[...]

PR #65, opened 2019-02-28
Implement --factor option

jasonrhansen:
Implements #19 with a difference from the Perl implementation in how sample counts get rounded, as explained in my comment on that issue. We need to decide if this is acceptable or not.

I've also combined all of the flamegraph tests into one file since they all duplicated the main testing code.


src/bin/flamegraph.rs
```diff
@@ -66,6 +70,9 @@ impl Into<Options> for Opt {
             options.title = "Icicle Graph".to_string();
         }
         options.negate_differentials = self.negate;
+        if let Some(factor) = self.factor {
+            options.factor = factor;
+        }
```
jonhoo: Can't this just set `options.factor = self.factor`?

src/flamegraph/merge.rs
```diff
@@ -184,10 +184,17 @@ fn parse_nsamples(line: &mut &str) -> Option<usize> {
...
     // foobar 1.klwdjlakdj
-    // TODO: Properly handle fractional samples (see issue #43)
+    //
+    // The Perl version keeps the fractional part but this can be problematic
+    // because of cumulative floating point errors. Instead we recommend to
+    // use the --factor option. See https://github.com/brendangregg/FlameGraph/pull/18
     if let Some(doti) = samples.find('.') {
+        // Warn if we're stripping a non-zero fractional part.
+        if !samples[doti + 1..].chars().all(|c| c == '0') {
+            warn!("Fractional part of {} stripped. If you need to retain the extra precision you can scale up the sample data and use the --factor option to scale it back down.", samples);
```
jonhoo: I wonder if we want something in place to ensure that we only print this warning _once_.

jonhoo:
Looks good beyond my comment in https://github.com/jonhoo/inferno/issues/19#issuecomment-468304554. Specifically, I think it's not actually okay for us to round and truncate the result to an integer -- it may be fractional after multiplying by the factor.

src/bin/flamegraph.rs
jasonrhansen: No, because `options.factor` is `f64` and `self.factor` is `Option<f64>`. `option.factor` gets set to `1.0` in the `Default` implementation and we only override it here if `--factor` was passed on the command line.

jonhoo:
Thinking some more, I think we specifically want `f64::trunc`, since `f64::floor` would give the next _lower_ negative number. Not that we expect any of these to be negative, but :man_shrugging:

src/flamegraph/merge.rs
jasonrhansen: Yes, good point. If there's one fractional sample, there's probably many. No need to repeat the same warning message over and over.

This brings up another point. We currently don't include a `log` implementation in our binaries so nothing happens with these messages. Do you want me to add one? Do you have a preferred one?

src/bin/flamegraph.rs
jonhoo: Hmm.. Why do we need it to be `Option<f64>` anywhere then? We already have a `default_value` set for it, right?

src/flamegraph/merge.rs
jonhoo: I don't have a preferred one, but you're right that adding a `log` impl to the binary is probably a good idea (same with `collapse`).

jasonrhansen:
Using `f64::trunc` doesn't match the Perl version either. It's more complicated than that because of the weirdness of formatting floating point numbers. `round` actually works most of the time, but it's in the middle at `.5` where we have issues.

`printf "%.0f", 1.5` outputs `2`
`printf "%.0f", 2.5` outputs `2`
`printf "%.0f", 3.5` outputs `4`

If we round we get `2`, `3`, `4`.
If we truncate we get `1`, `2`, `3`.

If we didn't care about thousands separators we could just format it the same way as the Perl version.

src/bin/flamegraph.rs
jasonrhansen: We want the default to be `1.0` when using as a library, not just when using the CLI, so I have the default on the library's `Options` struct similar to the `title` field.  But I think you're right that we should use `default_value` on the `structopt` field. This just means that we have the default value defined in two places.

jonhoo:
Wow, that's bizarre. Is that the case in Rust too? That makes no sense to me. But if that's indeed the case, then yes, I think `round()` is fine!

jasonrhansen:
Rust formats exactly the same way when using `"{:.0}"`. I'll keep using `round`, but add a comment about how this slightly differs from the Perl version.

src/flamegraph/merge.rs
jasonrhansen: OK, I'll add a `log` impl in a separate PR.

jonhoo:
Sounds good! I wonder where that behavior comes from...

src/flamegraph/merge.rs
```diff
@@ -176,11 +180,15 @@ where
         );
     }
 
+    if stripped_fractional_samples {
+        warn!("The fractional parts of sample counts have been stripped. If you need to retain the extra precision you can scale up the sample data and use the --factor option to scale it back down.");
```
jonhoo: I think it's better for this warning to be printed the first time we see it rather than at the very end. Otherwise I might not realize until the program has completed that I need to re-run.

jonhoo:
:tada: :heart:

PR #66, opened 2019-02-28
Add env_logger to binaries

jasonrhansen:
We were already using the `log` crate in our libraries, but we weren't using a log impl in our binaries (inferno-collapse-perf and inferno-flamegraph). This commit uses env_logger to output "warn" or higher logging levels to stderr by default.


jonhoo:
We should probably also add a `-v` flag which lowers that as well.

jasonrhansen:
Do you want to include all log priorities with `-v`? We could also include `-q` flag to silence all log output.

jonhoo:
@jasonrhansen let's follow convention and have each `-v` increase logging by one "level" (using [`multiple`](https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.multiple)). `-q` as quiet also seems like a good idea!

jonhoo:
Looks great!

Issue #63, opened 2019-02-25
Improve test coverage for codebase


jonhoo:
The [code coverage](https://codecov.io/gh/jonhoo/inferno/tree/master/src) of our tests is good, but not as good as we'd like it to be. In particular, I see some obvious candidates for new tests.

In `perf` stack collapsing ([see coverage report](https://codecov.io/gh/jonhoo/inferno/src/master/src/collapse/perf.rs)):

 - [ ] Collapsing a `perf` trace with an [empty line](https://github.com/jonhoo/inferno/blob/4863e01504e795726f31e751eadbe66f9337faad/src/collapse/perf.rs#L192)
 - [ ] Collapsing a `perf` trace with a ["weird" event line](https://github.com/jonhoo/inferno/blob/4863e01504e795726f31e751eadbe66f9337faad/src/collapse/perf.rs#L354)
 - [ ] Collapsing a `perf` trace with a ["process name"](https://github.com/jonhoo/inferno/blob/4863e01504e795726f31e751eadbe66f9337faad/src/collapse/perf.rs#L302)
 - [ ] Collapsing a `perf` trace with an [inline Java function](https://github.com/jonhoo/inferno/blob/4863e01504e795726f31e751eadbe66f9337faad/src/collapse/perf.rs#L334)
 - [ ] Collapsing a `perf` trace with a ["weird" stack line](https://github.com/jonhoo/inferno/blob/4863e01504e795726f31e751eadbe66f9337faad/src/collapse/perf.rs#L354)
 - [ ] Using `--uninline` on a binary [without debug symbols](https://github.com/jonhoo/inferno/blob/4863e01504e795726f31e751eadbe66f9337faad/src/collapse/perf.rs#L424)
 - [ ] Using `--uninline` on a frame with an [invalid address](https://github.com/jonhoo/inferno/blob/4863e01504e795726f31e751eadbe66f9337faad/src/collapse/perf.rs#L440)
 - [ ] Using `--uninline` on a frame with a [non-existing address](https://github.com/jonhoo/inferno/blob/4863e01504e795726f31e751eadbe66f9337faad/src/collapse/perf.rs#L449)
 - [ ] Using `--uninline` on a frame where the debug symbols are incomplete (see the various `??` cases in [this loop](https://github.com/jonhoo/inferno/blob/4863e01504e795726f31e751eadbe66f9337faad/src/collapse/perf.rs#L457))
[...]

PR #67, opened 2019-03-01
Add additional flamegraph tests

jasonrhansen:
Includes tests for the following from #63:
- Use a palette file
- Ignore bad lines
- Plot an empty trace
- Plot a file where frames get pruned
- Plot an inverted flame graph
- Include grey frames
- Test a non-sorted multi-reader input


tests/flamegraph.rs
```diff
@@ -1,34 +1,65 @@
...
+    options: Options,
+) -> quick_xml::Result<()> {
     let r = File::open(input_file).unwrap();
     let expected_len = fs::metadata(expected_result_file).unwrap().len() as usize;
     let mut result = Cursor::new(Vec::with_capacity(expected_len));
-    flamegraph::from_reader(options, r, &mut result).unwrap();
-    let mut expected = BufReader::new(File::open(expected_result_file).unwrap());
+    let return_value = flamegraph::from_reader(options, r, &mut result);
+    let expected = BufReader::new(File::open(expected_result_file).unwrap());
+    result.set_position(0);
```
jonhoo: I don't think you explicitly have to set the position to 0?

tests/flamegraph.rs
```diff
@@ -116,9 +167,128 @@ fn flamegraph_nameattr() {
...
+
+#[test]
+fn flamegraph_should_warn_about_fractional_samples() {
+    test_flamegraph_logs(
+        "./tests/data/fractional-samples/fractional.txt",
+        |captured_logs| {
+            assert_eq!(captured_logs.len(), 1);
+            assert!(captured_logs[0].body.contains("fractional sample"));
+            assert_eq!(captured_logs[0].level, Level::Warn);
+        },
```
jonhoo: We may want to make this test a _little_ more liberal in that it should test that there is exactly one occurrence of that warning (as opposed to _also_ requiring that there are no other outputs).

tests/flamegraph.rs
```diff
@@ -116,9 +167,128 @@ fn flamegraph_nameattr() {
...
+    );
+}
+
+#[test]
+fn flamegraph_should_not_warn_about_zero_fractional_samples() {
+    test_flamegraph_logs(
+        "./tests/data/fractional-samples/zero-fractionals.txt",
+        |captured_logs| {
+            assert!(captured_logs.is_empty());
+        },
```
jonhoo: Here too, I think we probably want to be a little more liberal. Specifically, check that there are no messages about fractional numbers.

tests/flamegraph.rs
```diff
@@ -116,9 +167,128 @@ fn flamegraph_nameattr() {
...
+
+    test_flamegraph(input_file, expected_result_file, options).unwrap();
+}
+
+#[test]
+fn flamegraph_should_warn_about_bad_input_lines() {
+    test_flamegraph_logs("./tests/data/bad-lines/bad-lines.txt", |captured_logs| {
+        assert_eq!(captured_logs.len(), 1);
+        assert!(captured_logs[0].body.contains("invalid format"));
+        assert_eq!(captured_logs[0].level, Level::Warn);
```
jonhoo: Don't we also print something along the lines of "N lines ignored"? I guess that's only in collapsing.
In any case, I think we should be liberal here too.

tests/flamegraph.rs
```diff
@@ -116,9 +167,128 @@ fn flamegraph_nameattr() {
...
+    });
+}
+
+#[test]
+fn flamegraph_should_warn_about_empty_input() {
+    test_flamegraph_logs("./tests/data/empty/empty.txt", |captured_logs| {
+        assert_eq!(captured_logs.len(), 1);
+        assert_eq!(captured_logs[0].body, "No stack counts found");
+        assert_eq!(captured_logs[0].level, Level::Error);
```
jonhoo: Same here.

jonhoo:
Looks good overall! Thanks for putting in the work :D
Left a few inline comments but they should be pretty minor!

tests/flamegraph.rs
jasonrhansen: We're using `result` as both a writer and reader. It gets passed into `flamegraph::from_reader` to write the result into, which advances the cursor. Then when we're comparing we need to reset the cursor so we start reading from the beginning.

PR #68, opened 2019-03-01
Fix fractional samples false positive

jasonrhansen:
While testing inferno-flamegraph on some Rust profiles I noticed I was getting some false positives for the fractional samples warning. It was because of lines like this.
```
rustc;<alloc::vec::Vec<T> as alloc::vec::SpecExtend<T, I>>::from_iter;memcpy@GLIBC_2.2.5 3
```
The `3` gets correctly parsed as a sample, but then we check for another sample in case we're creating a differential flame graph. It falsely detects `I>>::from_iter;memcpy@GLIBC_2.2.5` as a fractional sample since it ends in `.5`.

This change makes sure that the part before the `.` is all digits.

### DIFF
--- a/src/flamegraph/merge.rs
+++ b/src/flamegraph/merge.rs
@@ -195,11 +195,15 @@ fn parse_nsamples(line: &mut &str, stripped_fractional_samples: &mut bool) -> Op
     // because of cumulative floating point errors. Instead we recommend to
     // use the --factor option. See https://github.com/brendangregg/FlameGraph/pull/18
     if let Some(doti) = samples.find('.') {
-        // Warn if we're stripping a non-zero fractional part, but only the first time.
-        if !*stripped_fractional_samples
-            && samples[..doti].chars().all(|c| c.is_digit(10))
-            && !samples[doti + 1..].chars().all(|c| c == '0')
+        if !samples[..doti]
+            .chars()
+            .chain(samples[doti + 1..].chars())
+            .all(|c| c.is_digit(10))
         {
+            return None;
+        }
+        // Warn if we're stripping a non-zero fractional part, but only the first time.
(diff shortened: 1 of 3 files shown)


jonhoo:
Hmm, the implemented solution doesn't seem quite right? Specifically, won't this still end up parsing the function name as a sample count, it just won't emit the warning? I think the right thing to do here is to only parse the line as having a single sample count?

jasonrhansen:
I only changed the warning logic. The function will still try to parse as a `usize` below and fail. It will return `None` and `line` will remain unmodified.
```Rust
    let nsamples = samples.parse::<usize>().ok()?;
    // remove nsamples part we just parsed from line
    *line = line[..samplesi].trim_end();
```

jasonrhansen:
One thing this implementation doesn't check is whether the part after the `.` is all digits. If it saw something like `123.foo`, it would strip off the `.foo` and parse `123` as a usize. This probably wouldn't come up in practice since I don't know of any language that allows function names to start with a digit. We could change that section to something like the following if you'd like.
```Rust
    if let Some(doti) = samples.find('.') {
        if !samples[..doti]
            .chars()
            .chain(samples[doti + 1..].chars())
            .all(|c| c.is_digit(10))
        {
            return None;
        }
        // Warn if we're stripping a non-zero fractional part, but only the first time.
        if !*stripped_fractional_samples && !samples[doti + 1..].chars().all(|c| c == '0') {
            *stripped_fractional_samples = true;
            warn!("The input data has fractional sample counts that will be truncated to integers. If you need to retain the extra precision you can scale up the sample data and use the --factor option to scale it back down.");
        }
        samples = &samples[..doti];
    }
```

jonhoo:
Ah, yes, of course, you're entirely right. I interpreted the change too locally. I do think I'd like to see a change like the one you proposed above though, just to be rigorous about it.

PR #70, opened 2019-03-07
Refactor tests

jasonrhansen:


src/bin/flamegraph.rs
```diff
@@ -61,6 +61,17 @@ struct Opt {
     /// Verbose logging mode (-v, -vv, -vvv)
     #[structopt(short = "v", long = "verbose", parse(from_occurrences))]
     verbose: usize,
+
+    /// Pretty print XML with new lines and indentation.
```
```suggestion
    /// Pretty print XML with newlines and indentation.
```

tests/data/collapse-perf/inline-counter.txt
```diff
@@ -0,0 +1,243 @@
+inline-counter 26037  8801.193547:     250000 cpu-clock:uhH:
```
jonhoo: I'm not entirely sure why this file seems to only have been added now?

jonhoo:
LGTM subject to those two minor comments.

tests/data/collapse-perf/inline-counter.txt
jasonrhansen: I moved the file and had to update the path to the binary everywhere it was referenced.

jonhoo: Ah, I just missed its removal is all!

PR #76, opened 2019-03-09
Add perf stack collapsing tests

jasonrhansen:


jonhoo:
Excellent! I assume you've checked the folded output against the Perl version too?

jonhoo:
Oh, relatedly, I just found [`example-perf-stacks.txt.gz`](https://github.com/brendangregg/FlameGraph/blob/master/example-perf-stacks.txt.gz). We may want to add that as a test, and include the generated flamegraph in the README, since it'll (presumably) match the one on `flamegraph.pl`!

jasonrhansen:
Yes, I've checked these against the output of the Perl version.

I agree that we should use `example-perf-stacks-txt.gz` as an additional test. We could use it as a test for both perf collapsing and flamegraph generation. I can do that in a separate PR.

Issue #73, opened 2019-03-08
When generating a flamegraph programmatically the `flamegraph::Options::consistent_palette` option shouldn't require filesystem access


koute:
There should be an API which allows the user to enable the `consistent_palette` option and maintain the palette map themselves in memory without touching the filesystem.

jonhoo:
@JordiChauzi this is probably your domain. Any ideas for how we might do this? Maybe the trick is going to be to have the CLI frontend open the file, and then have the `Options` struct to `flamegraph` just take a `Box<Read>`?

koute:
I haven't looked at the code, but a quick suggestion - passing a `String`/`Vec<u8>` to `Options` and letting the user extract it back somehow would probably be the simplest. (As I assume there also needs to be a way to read this back so that it can be saved somewhere by the user?) Or perhaps create a dedicated opaque `PaletteMap` struct, and pass that to the `Options`? (That way `inferno` won't have to continuously reparse the palette map on every invocation of `from_sorted_lines.)

jonhoo:
I think `R: Read` covers that use-case. Specifically, you could give in something like `Cursor<&[u8]>` and it should just work. Wrapping that in an opaque `PaletteMap` struct might be a good idea too to avoid re-parsing. Where it gets a little complicated is that we also want to _store_ the palette map at the end of execution (since we may have added a bunch of entries to it). Perhaps what we should do here is have `PaletteMap` be a _trait_, and then `impl` it for `PathBuf` along with a map like today, but also let the user provide their own impl. Specifically, it'd need
```rust
trait PaletteMap {
    fn lookup(&self, func: &str) -> (u8, u8, u8);
    fn remember(&mut self, func: &str, color: (u8, u8, u8));
    fn save(self);
}
```

koute:
> I think `R: Read` covers that use-case. Specifically, you could give in something like `Cursor<&[u8]>` and it should just work.

But then `inferno` would have to parse it every time `from_sorted_lines` gets called, no? With a dedicated struct you could just call `PaletteMap::from_bytes` or `PaletteMap::from_stream` to parse it once and create a `PaletteMap` instance, use it multiple times when calling `flamegraph::from_sorted_lines` (pass it as `Option<&mut PaletteMap>`?), and then call `PaletteMap::write_into_stream` to serialize it back somewhere.

jonhoo:
Yeah, you're right, hence my updated suggestion of having `PaletteMap` being a trait instead. How does that sound?

[...]

PR #77, opened 2019-03-09
Provide a public interface for PaletteMap

JordiChauzi:
cf #73 (and [this message in particular](https://github.com/jonhoo/inferno/issues/73#issuecomment-471023113))

Do we want to provide a custom type alias for  `(u8, u8, u8)` ? Or use a "standard" RGB color structure ?

I would have preferred not to add `PaletteMap` as a function parameter of `from_sorted_lines`, but the `PaletteMap` cannot outlive the `input: String` declared in `from_reader` & `from_readers`, so it seemed that it could not be stored in `Options`. If there is a better solution, I am all ears.


src/flamegraph/color/palette_map.rs
```diff
@@ -5,67 +5,96 @@ use std::fs::OpenOptions;
...
+    ///
+    /// Each line should follow the format: NAME->rgb(RED, GREEN, BLUE)
+    /// where NAME is the function name, and RED, GREEN, BLUE integer values between 0 and 255
+    /// included.
+    ///
+    /// This function will return an [`std::io::Error`] if the input is not correctly formatted.
+    pub fn from_stream(reader: &mut dyn io::Read) -> io::Result<Self> {
         let mut map = HashMap::default();
-        let path = Path::new(file);
+        let reader = BufReader::new(reader);
```
jonhoo: Maybe we should instead just take a `&mut dyn io::BufRead` directly?

src/flamegraph/color/palette_map.rs
```diff
@@ -5,67 +5,96 @@ use std::fs::OpenOptions;
...
-                    return Err(quick_xml::Error::UnexpectedToken(line));
-                }
-
-                let rgb_color = parse_rgb_string(color)
-                    .ok_or_else(|| quick_xml::Error::UnexpectedToken(color.to_string()))?;
-                map.insert(Cow::from(name.to_string()), rgb_color);
-            }
+        for line in reader.lines() {
+            let line = line?;
+            let (name, color) = parse_line(&line)?;
```
jonhoo: Would it be better to warn and skip rather than return if we encounter an ill-formatted line?

src/flamegraph/color/palette_map.rs
```diff
@@ -5,67 +5,96 @@ use std::fs::OpenOptions;
...
 
         Ok(PaletteMap(map))
     }
 
-    pub fn save(self, file: &str) -> Result<(), io::Error> {
-        let mut file = OpenOptions::new().write(true).create(true).open(file)?;
-        let mut entries = self.0.into_iter().collect::<Vec<_>>();
+    /// Writes the palette map using the given writer.
+    /// The output content will follow the same format described in [from_stream()]
+    /// The name/color pairs will be sorted, based on the name lexicographic order.
```
jonhoo:
```suggestion
    /// The name/color pairs will be sorted by name in lexicographic order.
```

src/flamegraph/color/palette_map.rs
```diff
@@ -5,67 +5,96 @@ use std::fs::OpenOptions;
...
         }
 
         Ok(())
     }
 
+    /// Utility function to load a palette map from a file.
+    ///
+    /// The file content should follow the format described in [from_stream()].
+    ///
+    /// If the file does not exist, and empty palette map is returned.
```
jonhoo: I wonder if we should make the method name reflect this fact somehow?

src/flamegraph/color/palette_map.rs
```diff
@@ -5,67 +5,96 @@ use std::fs::OpenOptions;
...
+    ///
+    /// The file content will follow the format described in [from_stream()].
+    pub fn save_to_file(&self, path: &dyn AsRef<Path>) -> io::Result<()> {
+        let mut file = OpenOptions::new().write(true).create(true).open(path)?;
+        self.to_stream(&mut file)
+    }
+
+    /// Returns the color value corresponding to the given function name if it is present.
+    /// Otherwise compute the color, and insert the new function name/color in the map.
     pub fn find_color_for<F: FnMut(&'a str) -> (u8, u8, u8)>(
```
jonhoo: Should this be `pub(crate)`?

src/flamegraph/color/palette_map.rs
```diff
@@ -100,3 +153,53 @@ fn parse_rgb_string(s: &str) -> Option<(u8, u8, u8)> {
...
+        assert_eq!(palette.get("baz"), Some((255, 0, 255)));
+
+        let mut vec = palette.iter().collect::<Vec<_>>();
+        vec.sort_unstable();
+        let mut iter = vec.iter();
+
+        assert_eq!(iter.next(), Some(&("bar", (50, 0, 60))));
+        assert_eq!(iter.next(), Some(&("baz", (255, 0, 255))));
+        assert_eq!(iter.next(), Some(&("foo", (128, 128, 128))));
+        assert_eq!(iter.next(), None);
```
jonhoo: I wonder if we also want a basic test for `from_stream` and `to_stream` here. Specifically, something like "store the map to a `Vec<u8>`, load that again, and see that the same entries are present.

src/flamegraph/mod.rs
```diff
@@ -250,17 +252,16 @@ impl Rectangle {
 ///
 /// [differential flame graph]: http://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html
 #[allow(clippy::cyclomatic_complexity)]
-pub fn from_sorted_lines<'a, I, W>(opt: Options, lines: I, writer: W) -> quick_xml::Result<()>
+pub fn from_sorted_lines<'a, I, W>(
+    opt: Options,
+    lines: I,
+    writer: W,
+    mut palette_map: Option<&mut color::PaletteMap<'a>>,
```
jonhoo: Ouch, yeah, that's pretty unfortunate... What do we think the downside is of just making the key of `PaletteMap` `String` is?

src/flamegraph/mod.rs
```diff
@@ -576,7 +579,40 @@ where
 
     let mut lines: Vec<&str> = input.lines().collect();
     lines.sort_unstable();
-    from_sorted_lines(opt, lines, writer)
```
jonhoo: Can this just call `from_reader` at this point? Or can we factor out a function that does the below so we don't end up repeating it? I think that'll also let us bring `fetch_consistent_palette_if_needed` and `save_consistent_palette_if_needed` back inline to said function.

src/flamegraph/color/palette_map.rs
```diff
@@ -5,67 +5,96 @@ use std::fs::OpenOptions;
...
-pub(crate) struct PaletteMap<'a>(HashMap<Cow<'a, str>, (u8, u8, u8)>);
+/// Mapping of the association between a function name and the color used when drawing information
+/// from this function.
+#[derive(Default)]
+pub struct PaletteMap<'a>(HashMap<Cow<'a, str>, (u8, u8, u8)>);
 
 impl<'a> PaletteMap<'a> {
-    pub fn load(file: &str) -> quick_xml::Result<Self> {
+    /// Returns the color value corresponding to the given function name.
+    pub fn get(&self, func: &str) -> Option<(u8, u8, u8)> {
```
jonhoo: I agree with you -- let's use a standard crate for this color type!

src/flamegraph/color/palette_map.rs
JordiChauzi: I find it difficult to handle errors if we just print a warning. But the alternative, if we want to let the user choose what to do, is to accumulate the errors and return something like `(Self, Iterator<Item=io::Error>)` ? It may be a little too complex for what we want to do here.

src/flamegraph/color/palette_map.rs
JordiChauzi: I am not sure of the implications of such a change. Do we do not want to provide the user with the most generic interface ? Or is there an advantage for the user to provide a different `BufRead` implementation that the one we use ?

src/flamegraph/mod.rs
JordiChauzi: Well, we already allocate `String` when reading function names from the palette file. So only new functions names that are not already in the `PaletteMap` after reading the file are just borrowed. If the main use case is to use `PaletteMap` when calling `flamegraph` repetitively on inputs from different executions of the same program, we would lose virtually nothing (apart from the first iteration).

src/flamegraph/mod.rs
JordiChauzi: Actually, if we change `PaletteMap` to have `String` keys, and thus be stored in `Options`, we may want to fill the `PaletteMap` one level higher in `main` ? This would return `from_reader`/`from_readers` to a simpler form.

I am not sure how this would impact integration tests however. We would probably have to do some additional work in the test itself because `from_reader` is our main interface, right ?

src/flamegraph/color/palette_map.rs
jonhoo: It means that if they already have something that implements `BufRead` (like `Cursor` or a `BufReader`) we don't add our own buffer on top of that. In some sense, switching to `BufRead` gives the user _more_ choice. They can always wrap it in a `BufReader`, and we don't particularly care!

src/flamegraph/color/palette_map.rs
jonhoo: I was more thinking "be liberal in what you accept" here. If a file has one line with some garbage in it, it seems unfortunate to fail on the entire file?

src/flamegraph/mod.rs
jonhoo: Okay, I think I'm onboard with switching to `String` then :)

src/flamegraph/mod.rs
jonhoo: Yeah, I think `from_reader` is the primary interface, at least for now. But that takes an `Options`, right?

jasonrhansen:
In my opinion we should rename `from_stream` and `to_stream` to `from_reader` and `to_writer`. What do you think @JordiChauzi and @jonhoo?

src/flamegraph/color/palette_map.rs
```diff
@@ -1,84 +1,136 @@
...
+        let mut file = OpenOptions::new().write(true).create(true).open(path)?;
+        self.to_stream(&mut file)
+    }
+
+    /// Returns the color value corresponding to the given function name if it is present.
+    /// Otherwise compute the color, and insert the new function name/color in the map.
+    pub(crate) fn find_color_for<S: ToString, F: FnMut(&S) -> Color>(
         &mut self,
-        name: &'a str,
+        name: &S,
```
jonhoo: Hmm, this doesn't seem right. This will cause us to double-allocate every string we look up colors for, no? I wonder if it'll be better here to take `&str` and do the `.get` and `.insert` calls separately?

jonhoo:
@jasonrhansen yeah, I agree with you -- that fits better with what other crates do.

src/flamegraph/color/palette_map.rs
```diff
@@ -1,84 +1,136 @@
...
+    ///
+    /// The reader should provide name/color pairs as text input, each pair separated by a line
+    /// separator.
+    ///
+    /// Each line should follow the format: NAME->rgb(RED, GREEN, BLUE)
+    /// where NAME is the function name, and RED, GREEN, BLUE integer values between 0 and 255
+    /// included.
+    ///
+    /// This function will return an [`std::io::Error`] if the input is not correctly formatted.
+    pub fn from_stream(reader: &mut dyn io::BufRead) -> io::Result<Self> {
```
jonhoo: Let's make this `from_reader` to be more consistent with other crates.

src/flamegraph/color/palette_map.rs
```diff
@@ -1,84 +1,136 @@
...
         Ok(PaletteMap(map))
     }
 
-    pub fn save(self, file: &str) -> Result<(), io::Error> {
-        let mut file = OpenOptions::new().write(true).create(true).open(file)?;
-        let mut entries = self.0.into_iter().collect::<Vec<_>>();
+    /// Writes the palette map using the given writer.
+    /// The output content will follow the same format described in [from_stream()]
+    /// The name/color pairs will be sorted by name in lexicographic order.
+    pub fn to_stream(&self, writer: &mut dyn io::Write) -> io::Result<()> {
```
jonhoo: Let's make this `to_writer` to be more consistent with other crates.

src/flamegraph/color/palette_map.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/77/commits/f99e779f3d876c95dac2e022186a1f4d89cc7cd3)

src/flamegraph/color/palette_map.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/77/commits/14ddd2402e973a7b993511e49488ab75438bc0ad)

src/flamegraph/mod.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/77/commits/6a5462050c0d2a8e9d396e35aa1e66a550bc16b2)

src/flamegraph/color/palette_map.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/77/commits/37c047b620b0328b42202161e093087eabaf805c)

src/flamegraph/color/palette_map.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/77/commits/574a80ba07903eb71fc0fded6759e77cee247afc)

src/flamegraph/color/palette_map.rs
JordiChauzi: [Done](https://github.com/jonhoo/inferno/pull/77/commits/574a80ba07903eb71fc0fded6759e77cee247afc)

src/flamegraph/color/palette_map.rs
```diff
@@ -9,7 +9,7 @@ use std::str::FromStr;
 
 /// Mapping of the association between a function name and the color used when drawing information
 /// from this function.
-#[derive(Default)]
+#[derive(Debug, Default)]
```
jonhoo: Let's also add `Clone` to this. Seems pretty reasonable that that might be something someone wants to do. Perhaps also `Eq, PartialEq`.

src/flamegraph/color/palette_map.rs
```diff
@@ -36,15 +36,17 @@ impl PaletteMap {
...
     pub fn from_reader(reader: &mut dyn io::BufRead) -> io::Result<Self> {
         let mut map = HashMap::default();
 
         for line in reader.lines() {
             let line = line?;
-            let (name, color) = parse_line(&line)?;
-            map.insert(name.to_string(), color);
+            if let Ok((name, color)) = parse_line(&line) {
+                map.insert(name.to_string(), color);
+            }
```
jonhoo: We probably want to log a warning otherwise.

PR #78, opened 2019-03-09
Add simple dtrace collapse

Licenser:
Add a minimal collapser for dtrace


jonhoo:
So, just to check, this is effectively a port of [`stackcollapse.pl`](https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse.pl), right?

Licenser:
It's output compatible but code wise more a port form the perf.rs to dtrace.rs (with a buch of deletion).

I'm a bit confused on how the ci works so. It looks in the `flamegrapm` (or `dtrace`) folder but the data lives under `tests/data/dtrace`.

src/bin/collapse-dtrace.rs
```diff
@@ -0,0 +1,70 @@
+use env_logger::Env;
+use std::io;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+use inferno::collapse::dtrace::{Folder, Options};
+use inferno::collapse::Collapse;
+
+#[derive(Debug, StructOpt)]
+#[structopt(
```
jonhoo: The values here don't look quite right. You probably to add a doc-comment to match [upstream](https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse.pl#L3L8). If you want to be particularly helpful, you'd probably want to include some [DTrace tips](https://github.com/brendangregg/FlameGraph#dtrace) in `after_help`; if not, just delete it!

Licenser: 👍

src/bin/collapse-dtrace.rs
```diff
@@ -0,0 +1,70 @@
...
+[1] perf script must emit both PID and TIDs for these to work; eg, Linux < 4.1:
+        perf script -f comm,pid,tid,cpu,time,event,ip,sym,dso,trace
+    for Linux >= 4.1:
+        perf script -F comm,pid,tid,cpu,time,event,ip,sym,dso,trace
+    If you save this output add --header on Linux >= 3.14 to include perf info."
+)]
+struct Opt {
+    /// include raw addresses where symbols can't be found
+    #[structopt(long = "addrs")]
+    include_addrs: bool,
```
jonhoo: I think this is called `includeoffset` [upstream](https://github.com/brendangregg/FlameGraph/blob/1b1c6deede9c33c5134c920bdb7a44cc5528e9a7/stackcollapse.pl#L54), but given that it's not a command-line argument there, we should feel free to call it whatever we want to here. So let's stick with `addrs` for consistency with `collapse-perf` :)

src/bin/collapse-dtrace.rs
```diff
@@ -0,0 +1,70 @@
...
+    author = "",
+    after_help = "\
+[1] perf script must emit both PID and TIDs for these to work; eg, Linux < 4.1:
+        perf script -f comm,pid,tid,cpu,time,event,ip,sym,dso,trace
+    for Linux >= 4.1:
+        perf script -F comm,pid,tid,cpu,time,event,ip,sym,dso,trace
+    If you save this output add --header on Linux >= 3.14 to include perf info."
+)]
+struct Opt {
+    /// include raw addresses where symbols can't be found
```
jonhoo: Upstream says that this does:
```
/// include function offset (except leafs)
```

src/bin/collapse-dtrace.rs
```diff
@@ -0,0 +1,70 @@
...
+    #[structopt(long = "addrs")]
+    include_addrs: bool,
+
+    /// annotate kernel functions with a _[k]
+    #[structopt(long = "kernel")]
+    annotate_kernel: bool,
```
jonhoo: This doesn't existing in [upstream](https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse.pl)?

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+use smallvec::SmallVec;
+use std::collections::{HashMap, VecDeque};
+use std::io;
+use std::io::prelude::*;
+
+/// Settings that change how frames are named from the incoming stack traces.
+///
+/// All options default to off.
+#[derive(Clone, Debug, Default)]
+pub struct Options {
```
jonhoo: See comments above about these options.

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub include_addrs: bool,
+
+    /// Annotate kernel functions with a `_[k]` suffix.
+    pub annotate_kernel: bool,
+}
+
+/// A stack collapser for the output of `perf script`.
+///
+/// To construct one, either use `perf::Folder::default()` or create an [`Options`] and use
+/// `perf::Folder::from(options)`.
```
jonhoo: This doc-comment doesn't look right?

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Function entries on the stack in this entry thus far.
+    stack: VecDeque<String>,
+    count: usize,
+    /// Number of times each call stack has been seen.
+    occurrences: HashMap<String, usize>,
+
+    /// Current comm name.
+    ///
+    /// Called pname after original stackcollapse-perf source.
+    pname: Option<String>,
```
jonhoo: I don't think this is set anywhere, so it should probably be removed?

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+///
+/// To construct one, either use `perf::Folder::default()` or create an [`Options`] and use
+/// `perf::Folder::from(options)`.
+#[derive(Default)]
+pub struct Folder {
+    /// All lines until the next empty line are stack lines.
+    in_event: bool,
+
+    /// Skip all stack lines in this event.
+    skip_stack: bool,
```
jonhoo: I don't think this is used?

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+    count: usize,
+    /// Number of times each call stack has been seen.
+    occurrences: HashMap<String, usize>,
+
+    /// Current comm name.
+    ///
+    /// Called pname after original stackcollapse-perf source.
+    pname: Option<String>,
+
+    //    event_filtering: EventFilterState,
```
jonhoo: This comment seems misleading?

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+impl Collapse for Folder {
+    fn collapse<R, W>(&mut self, mut reader: R, writer: W) -> io::Result<()>
+    where
+        R: BufRead,
+        W: Write,
+    {
+        let mut line = String::new();
+        //TODO: clean this up, we skip the first 4 lines:
```
jonhoo: Looks like upstream defines this [as a constant](https://github.com/brendangregg/FlameGraph/blob/9d7c1b1233839de1f842135be84e54f2470adb42/stackcollapse.pl#L53) and then just skips that many lines. We should probably do something similar.

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+        //
+        reader.read_line(&mut line)?;
+        loop {
+            line.clear();
+
+            if reader.read_line(&mut line)? == 0 {
+                break;
+            }
+
+            if line.starts_with('#') {
```
jonhoo: Does the DTrace format even support comments? If not, we should remove this -- it's better to give the user an error if we see something in the file that shouldn't be there :)

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+            count: 0,
+            occurrences: HashMap::default(),
+            pname: None,
+            opt,
+        }
+    }
+}
+
+impl Folder {
+    fn on_line(&mut self, line: &str) {
```
jonhoo: This should probably just be folded into `on_stack_line`.

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+        self.in_event = true;
+        self.on_stack_line(line)
+    }
+
+    fn stack_line_parts(line: &str) -> Option<(&str, &str)> {
+        let mut line = line.splitn(2, '+');
```
jonhoo: I think you want `rsplitn` here, since we want the [last](https://github.com/brendangregg/FlameGraph/blob/9d7c1b1233839de1f842135be84e54f2470adb42/stackcollapse.pl#L74) `+`.

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+            if line.starts_with('#') {
+                continue;
+            }
+
+            let line = line.trim_end();
+            if line.is_empty() {
+                self.after_event();
+            } else {
+                self.on_line(line);
+            }
```
jonhoo: I think I'd change this slightly since the DTrace format seems to be:
```
stack line
stack line
count

stack line
stack line
count
```
I'd do:
```rust
if line.is_empty() {
    continue;
} else if let Ok(count) = line.parse::<usize>() {
    self.on_stack_end(count);
} else {
    self.on_stack_line(line);
}
```
That way you also don't need to remember how many lines to skip at the head of the file. Just ignore any lines up to the first empty line, and from that on do normal processing. Any and all empty lines will just be skipped.

src/collapse/dtrace.rs
jonhoo: See also the comment below about simply ignoring empty lines.

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+    //     genunix`syscall_mstate+0x5d
+    //     unix`sys_syscall+0x10e
+    //       1
+    fn on_stack_line(&mut self, line: &str) {
+        let line = line.trim_start();
+        if let Ok(count) = line.parse::<usize>() {
+            self.count = count;
+        } else {
+            if let Some((func, addr)) = Folder::stack_line_parts(line) {
+                self.stack.push_front(func.to_string());
```
jonhoo: This doesn't seem quite right.
First, upstream [only](https://github.com/brendangregg/FlameGraph/blob/9d7c1b1233839de1f842135be84e54f2470adb42/stackcollapse.pl#L85) removes the address part if `includeoffset` isn't set.
Second, it seems to do some [additional post-processing](https://github.com/brendangregg/FlameGraph/blob/9d7c1b1233839de1f842135be84e54f2470adb42/stackcollapse.pl#L87L100) on the function name that we probably also want to do. You should feel free to ignore the [Java un-inlining](https://github.com/brendangregg/FlameGraph/blob/9d7c1b1233839de1f842135be84e54f2470adb42/stackcollapse.pl#L93) for the time being if you want to, though it's [not too bad](https://github.com/jonhoo/inferno/blob/b7605ef69cc53e07612e098264b3bd6499ab5877/src/collapse/perf.rs#L275L281).

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+
+            // add the other stack entries (if any)
+            for e in self.stack.drain(..) {
+                if first {
+                    first = false
+                } else {
+                    stack_str.push_str(";");
+                }
+                stack_str.push_str(&e);
+            }
```
jonhoo: Do we need to also "[trim leaf offsets](https://github.com/brendangregg/FlameGraph/blob/9d7c1b1233839de1f842135be84e54f2470adb42/stackcollapse.pl#L73L74)" here?

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,219 @@
...
+        for key in keys {
+            writeln!(writer, "{} {}", key, self.occurrences[key])?;
+        }
+        Ok(())
+    }
+}
+
+// massage function name to be nicer
+// NOTE: ignoring https://github.com/jvm-profiling-tools/perf-map-agent/pull/35
+fn with_module_fallback(module: &str, func: &str, pc: &str, include_addrs: bool) -> String {
```
jonhoo: I don't think this function is currently used?

src/collapse/mod.rs
```diff
@@ -1,3 +1,10 @@
+/// Stack collapsing for the output of [`dtrace`](...).
```
jonhoo: We should probably have a link for this. [Joyent's page](https://www.joyent.com/dtrace) is probably the right one to use. We could also be helpful and include a link to @brendangregg's [dtrace tools page](http://www.brendangregg.com/dtrace.html).

tests/collapse-dtrace.rs
```diff
@@ -0,0 +1,183 @@
...
+extern crate inferno;
+
+use inferno::collapse::dtrace::{Folder, Options};
+use inferno::collapse::Collapse;
+use log::Level;
+use std::fs::{self, File};
+use std::io::{self, BufRead, BufReader, Cursor};
+use std::path::Path;
+
+fn test_collapse_perf(test_file: &str, expected_file: &str, options: Options) -> io::Result<()> {
```
jonhoo: We should replace `perf` with `dtrace` everywhere in this file.

tests/collapse-dtrace.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    if expected.read_line(&mut buf).unwrap() > 0 {
+        panic!(
+            "\n{} has more lines than output, beginning at line: {}",
+            expected_file, line_num
+        )
+    }
+
+    return_value
+}
```
jonhoo: This function can probably be shared across _all_ our collapsers given that they all have a shared trait (`Collapse`). We should take advantage of that and factor this function out. /cc @jasonrhansen

tests/collapse-dtrace.rs
```diff
@@ -0,0 +1,183 @@
...
+
+// Create tests for test files in $dir. The test files are used as input
+// and the results are compared to result files in the results sub directory.
+// The test and result file names are derived from $name.
+// The part after the double underscore are underscore separated options.
+// For example, perf_cycles_instructions_01_pid will use the following:
+//     test file: perf-cycles-instructions-01.txt
+//     result file: perf-cycles-instructions-01-collapsed-pid.txt
+//     flag: pid
+macro_rules! collapse_perf_tests_inner {
```
jonhoo: This can probably also be shared across all the collapse tests, right @jasonrhansen ?

tests/collapse-dtrace.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+collapse_perf_tests_upstream! {
+    collapse_dtrace_example_mini__all,
+
+    collapse_dtrace_example__all
+}
+
+/*
```
jonhoo: Just delete these :)

jonhoo:
This adds a bunch of empty files to `tests/data/dtrace/test/results/`, which doesn't seem right? It suggests to me that there's some stuff working. Ideally, all the things in `results` should be what's produced by the Perl version so that we can compare our output against that!

jonhoo:
Ah, the `flamegraph` [here](https://github.com/jonhoo/inferno/blob/b7605ef69cc53e07612e098264b3bd6499ab5877/tests/collapse-perf.rs#L127) refers to the git submodule in the root of the repository that is a checkout of the upstream flamegraph repository! You probably don't want to use `collapse_perf_tests_upstream` or `collapse_perf_tests_inner`, and instead just call `test_collapse_dtrace_logs` directly [like this](https://github.com/jonhoo/inferno/blob/b7605ef69cc53e07612e098264b3bd6499ab5877/tests/collapse-perf.rs#L226).

tests/collapse-dtrace.rs
Licenser: 👍

tests/collapse-dtrace.rs
Licenser: 👍

src/collapse/mod.rs
Licenser: 👍

src/collapse/dtrace.rs
Licenser: 👍

src/collapse/dtrace.rs
Licenser: this is the address removeal

src/bin/collapse-dtrace.rs
Licenser: 👍 removed it as it's not passable to upstream.

src/bin/collapse-dtrace.rs
Licenser: see above

src/bin/collapse-dtrace.rs
Licenser: :+1: removed since it's not in upstream.

src/collapse/dtrace.rs
Licenser: 👍

src/collapse/dtrace.rs
Licenser: 👍

src/collapse/dtrace.rs
Licenser: 👍 removed

src/collapse/dtrace.rs
Licenser: 👍 removed

src/collapse/dtrace.rs
Licenser: 👍 clutter, removed

src/collapse/dtrace.rs
Licenser: that would lead to some warnings about odd stack lines at the beginning.

src/collapse/dtrace.rs
Licenser: 👍

src/collapse/dtrace.rs
Licenser: 👍

Licenser:
Ah now the tests make sense! I've copied in the test from upstream into a static test the empty files should be gone!

src/collapse/dtrace.rs
Licenser: I updated this to skip a given number of lines.

src/collapse/dtrace.rs
Licenser: 👍 removed

src/collapse/dtrace.rs
Licenser: 👍

tests/collapse-dtrace.rs
Licenser: I removed this copy as it's no longer required.

src/collapse/dtrace.rs
jonhoo: Ah, no, I'm proposing that you _first_ skip leading non-empty lines (and don't warn for them). Only once you've hit the first empty line do you start "normal" line processing (including warning for malformed lines).

src/bin/collapse-dtrace.rs
```diff
@@ -0,0 +1,56 @@
...
+use std::io;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+use inferno::collapse::dtrace::{Folder, Options};
+use inferno::collapse::Collapse;
+
+#[derive(Debug, StructOpt)]
+#[structopt(
+    name = "inferno-collapse-perf",
```
jonhoo:
```suggestion
    name = "inferno-collapse-dtrace",
```

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,163 @@
...
+    static ref CPP_REGEX: Regex = Regex::new("(?P<n>::.*)[(<].*").unwrap();
+}
+
+/// Settings that change how frames are named from the incoming stack traces.
+///
+/// All options default to off.
+#[derive(Clone, Debug, Default)]
+pub struct Options {}
+
+/// A stack collapser for the output of dtrace `ustace()`.
```
jonhoo:
```suggestion
/// A stack collapser for the output of dtrace `ustrace()`.
```

jonhoo: Or is it `ustack`?

src/collapse/dtrace.rs
jonhoo: I think the strategy [below](https://github.com/jonhoo/inferno/pull/78/commits/bdf75bfe0183a32ec8d6d2196e70b18a7d3512e3#r264005789) is probably a better solution.

src/collapse/dtrace.rs
Licenser: yes sorry, my typo-foo is strong :/

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,163 @@
...
+///
+/// To construct one, either use `dtrace::Folder::default()` or create an [`Options`] and use
+/// `dtrace::Folder::from(options)`.
+#[derive(Default)]
+pub struct Folder {
+    /// All lines until the next empty line are stack lines.
+    in_event: bool,
```
jonhoo: Do we even need this any more?

Licenser: It's part of the after_event loop to prevent printing empty lines in case of a mal-formated (i.e. empty) file.

src/collapse/dtrace.rs
```diff
@@ -126,30 +107,23 @@ impl Folder {
         if let Ok(count) = line.parse::<usize>() {
             self.count = count;
         } else {
-            if let Some((func, addr)) = Folder::stack_line_parts(line) {
-                self.stack.push_front(func.to_string());
+            if let Some((mut funcs, addr)) = Folder::stack_line_parts(line) {
```
jonhoo: I think you probably want to call `dejavaify` out here instead, because otherwise you are allocating a `Vec` _on every line_. Take a look at what the [perf code does](https://github.com/jonhoo/inferno/blob/b7605ef69cc53e07612e098264b3bd6499ab5877/src/collapse/perf.rs#L259).

jonhoo: Alternatively, you could also pass in `&mut self.stack` to `stack_line_parts`

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,163 @@
...
+    //     genunix`gethrtime_unscaled+0xa
+    //     genunix`syscall_mstate+0x5d
+    //     unix`sys_syscall+0x10e
+    //       1
+    fn on_stack_line(&mut self, line: &str) {
+        let line = line.trim_start();
+        if let Ok(count) = line.parse::<usize>() {
+            self.count = count;
+        } else {
+            if let Some((mut funcs, _addr)) = Folder::stack_line_parts(line) {
```
jonhoo: This still _always_ ignores the `addr`, which I don't think is what [upstream does](https://github.com/brendangregg/FlameGraph/blob/9d7c1b1233839de1f842135be84e54f2470adb42/stackcollapse.pl#L85).

src/collapse/dtrace.rs
jonhoo: I'm not entirely sure what you mean? Upstream removes the address in _two_ places: [here](https://github.com/brendangregg/FlameGraph/blob/1b1c6deede9c33c5134c920bdb7a44cc5528e9a7/stackcollapse.pl#L73L74) and [here](https://github.com/brendangregg/FlameGraph/blob/1b1c6deede9c33c5134c920bdb7a44cc5528e9a7/stackcollapse.pl#L85). How come we only have to do it in one place?

src/collapse/dtrace.rs
```diff
@@ -27,7 +32,7 @@ pub struct Folder {
     /// Number of times each call stack has been seen.
     occurrences: HashMap<String, usize>,
 
-    opt: Options,
+    _opt: Options, // to keep the interface consistant
```
jonhoo: I think we probably want to keep the `includeoffset` option that upstream has. We don't have to expose it as a CLI argument, but internally it seems like a good idea!

src/collapse/dtrace.rs
```diff
@@ -1,11 +1,16 @@
...
+use lazy_static::lazy_static;
+use regex::Regex;
 use std::collections::{HashMap, VecDeque};
 use std::io;
 use std::io::prelude::*;
 
 const HEADER_LINES: u8 = 3;
 
+lazy_static! {
+    static ref CPP_REGEX: Regex = Regex::new("(?P<n>::.*)[(<].*").unwrap();
```
jonhoo: This is really just replicating [this regex](https://github.com/brendangregg/FlameGraph/blob/1b1c6deede9c33c5134c920bdb7a44cc5528e9a7/stackcollapse.pl#L88), right? We probably want to re-write this in straight Rust code rather than relying on a regex, as regexes are usually a fair amount slower. I'm on my way out the door, but maybe @jasonrhansen has an idea for how this can be re-written to not be a regex?

src/collapse/dtrace.rs
Licenser: I might misread the perf code or fundamentally misunderstand rust but the way I read the perf code I see one (small) vector per line being allocated in 278 along with a string parsing, then the same string is parsed again in 281.

But if the double allocation is a concern I can optimize it!

src/collapse/dtrace.rs
jonhoo: I _think_ with the proposed change where we just look for counts, we can simply ignore empty lines, no?

src/collapse/dtrace.rs
Licenser: It implicitly does [upstream](https://github.com/brendangregg/FlameGraph/blob/9d7c1b1233839de1f842135be84e54f2470adb42/stackcollapse.pl#L54)  just with a constant to do so.

jonhoo:
Ouch,  `tests/data/collapse-dtrace/test/dtrace-example.txt` is quite a file... I wonder if we instead want to just un-`gz` it from our `flamegraph` submodule (I assume it's the same one that's there) directly in the test rather than import it wholesale into inferno? /cc @jasonrhansen

Licenser:
Good call on the submodule, for test data! Removed it :)

src/collapse/dtrace.rs
Licenser: ahhh that makes more sense!

src/collapse/dtrace.rs
Licenser: The upstream script seems to be bugged there, it removes the offset `if $includeoffset` and `unless $includeoffset`.

src/collapse/dtrace.rs
Licenser: Got what you meant with the loop! removed :)

src/collapse/dtrace.rs
Licenser: I missread that, it jsut removes the last one if `includeoffset` is set to true.

Licenser: re-added

src/collapse/dtrace.rs
Licenser: added the logic for offset inclusion from upstream

src/collapse/dtrace.rs
Licenser: 👍 done

src/collapse/dtrace.rs
jasonrhansen: I would use something like the following instead of using the regex in `on_stack_line`:

```Rust
        // Remove arguments from C++ function names:
        if let Some(scope) = frame.rfind("::") {
            if let Some(open) = frame[scope + 2..].find(|c| c == '(' || c == '<') {
                frame = &frame[..scope + 2 + open];
            }
        }
```

Licenser: That's not exactly the same as the regex, there are some cases where results are different on the input. But it seems like the loop is closer to 'correct'.

I'll put in some test cases on this so we can see if it makes sense to not be fully compatible

jasonrhansen: You're correct that it doesn't match the regex exactly, but does it matter on real input?

The following lines in your tests give different results when using my code, but do lines like this ever exist in real profiles? Specifically do the places you've put `(blorg)` and `<blorg>` make sense?
```
TestClass::TestClass2(const char*)[__1cJTestClass2t6Mpkc_v_](blorg)
TestClass::TestClass2(const char*)<blorg>[__1cJTestClass2t6Mpkc_v_]
TestClass::TestClass2(const char*)[__1cJTestClass2t6Mpkc_v_]<blorg>
```

This one also gives different results:
```
TestClass::TestClass2<bla>(const char*)[__1cJTestClass2t6Mpkc_v_]
```
With regex: `TestClass::TestClass2<bla>`
With my code: `TestClass::TestClass2`

I'm wondering if not stripping the `<bla>` was intended, or a bug with the regex. If the `<bla>` isn't supposed to be stripped, then why would the regex even look for the `<` character at all?

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,214 @@
...
+impl Collapse for Folder {
+    fn collapse<R, W>(&mut self, mut reader: R, writer: W) -> io::Result<()>
+    where
+        R: BufRead,
+        W: Write,
+    {
+        let mut line = String::new();
+        for _ in 0..HEADER_LINES {
+            reader.read_line(&mut line)?;
+        }
```
jonhoo: I wonder if we could even get rid of the constant here with:
```rust
// skip header lines -- first empty line marks start of data
loop {
    line.clear();
    reader.read_line(&mut line)?;
    if line.is_empty() {
        break;
    }
}
```

jonhoo: You may also technically need to check for `read_line() == 0` here.

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,214 @@
...
+    //       1
+    fn on_stack_line(&mut self, line: &str) {
+        let line = line.trim_start();
+        let frame = if self.opt.includeoffset {
+            line
+        } else {
+            Self::remove_offset(line)
+        };
+
+        let mut frame = CPP_REGEX.replace_all(frame, "$n").to_string();
```
jonhoo: Here's another reason to get rid of the regex -- it would mean we could stick with `&str` (and no allocation) here (I think).

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,214 @@
...
+        let mut inline = false;
+
+        if frame.is_empty() {
+            frame = "-".to_string();
+        };
+
+        for func in frame.split("->") {
+            let mut func = func.replace(';', ":").trim_start_matches('L').to_string();
```
jonhoo: I _think_ this ends up allocating two strings?

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,214 @@
...
+    fn on_stack_line(&mut self, line: &str) {
+        let line = line.trim_start();
+        let frame = if self.opt.includeoffset {
+            line
+        } else {
+            Self::remove_offset(line)
+        };
+
+        let mut frame = CPP_REGEX.replace_all(frame, "$n").to_string();
+        let mut inline = false;
```
jonhoo: This is nitpicky, but can you move this down to just before the loop where it's used?

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,214 @@
...
+        };
+
+        for func in frame.split("->") {
+            let mut func = func.replace(';', ":").trim_start_matches('L').to_string();
+            if inline {
+                func.push_str("_[i]")
+            };
+            inline = true;
+            self.stack_str_size += func.len() + 1;
+            self.stack.push_front(func);
```
jonhoo: I don't _think_ this is quite right. Specifically, notice the use of `unshift` [here](https://github.com/brendangregg/FlameGraph/blob/1b1c6deede9c33c5134c920bdb7a44cc5528e9a7/stackcollapse.pl#L104). Consider something like:

```
baz
foo->bar
main
```

we want to generate

```
main;foo;bar;baz
```

The code above, unless I'm mistaken, won't do that. It'll iteratively build:

```
[baz], [foo, baz], [bar, foo, baz], [main, bar, foo, baz]
```

which results in

```
main;bar;foo;baz
```

Notice that the inlining is reversed. This is why the perf version does this strange [`push-pop-push_front`] dance, which I think we have to do here as well. We [use a `SmallVec`](https://github.com/jonhoo/inferno/blob/10a1fd7c4c891dbc6a3c6ed74541d884eb89b1e9/src/collapse/perf.rs#L278L279) for the inlined entries because _in the common case_ there isn't any inlining information (i.e., the function is just one function), and in that case `SmallVec` won't allocate anything!

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,214 @@
...
+            if first {
+                first = false
+            } else {
+                stack_str.push_str(";");
+            }
+            stack_str.push_str(&e);
+        }
+        //trim leaf offset if these were retained:
+        if self.opt.includeoffset {
+            stack_str = Self::remove_offset(&stack_str).to_string();
```
jonhoo: It'd be great if we could find a way to not re-allocate a `String` here since we already have a `String`.

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,214 @@
...
+    }
+}
+
+impl From<Options> for Folder {
+    fn from(opt: Options) -> Self {
+        Self {
+            stack: VecDeque::default(),
+            occurrences: HashMap::default(),
+            opt: opt,
+            stack_str_size: 0,
```
jonhoo: The perf collapse code just [computes this value](https://github.com/jonhoo/inferno/blob/10a1fd7c4c891dbc6a3c6ed74541d884eb89b1e9/src/collapse/perf.rs#L333); is there a particular reason you wanted to keep a running tally instead?

jonhoo:
This is starting to look pretty good! Left some more comments following the latest changes :)

src/collapse/dtrace.rs
Licenser: it is but I'll move it non the less :P

src/collapse/dtrace.rs
Licenser: you are right, I added a test for this

src/collapse/dtrace.rs
Licenser: 👍

src/collapse/dtrace.rs
Licenser: I have the impression that performance matters to the project, so keeping a tally is cheaper then re-iterating over something we already went over during parsing.

src/collapse/dtrace.rs
Licenser: I generally agree the find-version seems to deliver the correcter results but I'm unsure how bug-compatible @jonhoo wants to be ;)

src/collapse/dtrace.rs
Licenser: swapping this around allows to get rid of the to_string with the same effect :)

Licenser:
On the topic of bug compatibility, there is another case:

```
Diff < left / right > :
<"unix`0xfffffffffb8001d6;unix`trap+0x2c7;unix`pagefault+0x96;genunix`as_fault 1"
>"unix`0xfffffffffb8001d6;unix`trap+0x2c7;unix`pagefault 1"
```

left ours, right upstream.

The stack in question is:
```
              genunix`segvn_fault
              unix`pagefault+0x96
              unix`trap+0x2c7
              unix`0xfffffffffb8001d6
                1
```

the when `includeoffset` is true it 'removes the last leave offset' from the stack.  Upstream does that by removing everything after the last `+`, so it would remove `+0x96;genunix`as_fault` instead of just leaving things as they are.

Given this is the 2nd bug I'll drop bug compatibilty for now - we can re-add it if requried.

src/collapse/dtrace.rs
Licenser: I moved this inside the loop, that way it can remain a &str and at the same time operates over a smaller string so a win win :)

src/collapse/dtrace.rs
```diff
@@ -75,6 +88,7 @@ impl From<Options> for Folder {
         Self {
             stack: VecDeque::default(),
             occurrences: HashMap::default(),
+            java_vec: Vec::with_capacity(JAVA_VEC_SIZE),
```
jonhoo: Ah, yes, this is a good way to do it! Maybe we should do the same in the perf version rather than relying on `SmallVec`. What do you think @jasonrhansen? You initially chose `SmallVec` in #13.

src/collapse/dtrace.rs
jonhoo: Sure, that's a fair point. I doubt that particular counting is very expensive since it's just `O(#stack segments)`, but seems fine to me to do it this way too :)

src/bin/collapse-dtrace.rs
```diff
@@ -26,13 +26,22 @@ struct Opt {
     #[structopt(short = "v", long = "verbose", parse(from_occurrences))]
     verbose: usize,
 
+    /// Include offsets
+    #[structopt(short = "o", long = "includeoffset")]
```
jonhoo: Let's not use `-o` for this, since `-o` commonly means "output file". Instead, let's make the `long` shorter. How about `--offsets`?

src/collapse/dtrace.rs
jonhoo: I'm okay with us fixing things that are _obviously_ bugs, but I'm not so convinced by the above. But isn't it also easy to fix? Just make the `rfind` a `find` and the `find` an `rfind`? I think that should give us compatibility.

jonhoo:
@Licenser Yeah, I agree with you, that seems like an upstream bug. May be worth reporting even, given that it seems to affect their dtrace example a decent amount.

src/bin/collapse-dtrace.rs
Licenser: 👍

Licenser: hmm, writing it, no offset really doesn't tell muych about what it does, how about `-i`? and keep it the same name as upstream?

jonhoo: `-i` usually denotes "input file" :)
I think `--offsets` is pretty reasonable given the context -- what makes you think it isn't? Keep in mind that users will also see the help text (which we should probably expand a little) when they run `-h` or `--help`. Maybe make the doc-comment for this option:
```rust
/// Include function offsets for all but the last location in the call chain.
```

Licenser: It seems reasonable for a flag to say what it does, a flag named `--offset` usually lets you specify an offset to something, not prevent the program to skip something. `--includeoffset` does what it says, it includes the offset. The argument for the help text can as well be made for `-i` or `-o`; That all said I don't think that flag was ever used upstream, how about just not giving it a short?

src/collapse/dtrace.rs
jasonrhansen: Yeah, I'm also not sure it's a bug. That's why I asked the question. I was hoping one of you had a better I idea about what the intended behavior was. Switching `rfind` and `find` works unless there is more than one `::`.

src/bin/collapse-dtrace.rs
Licenser: also `-i` on grep is "ignore case" on sed it's "add extension" for ls it's "find serial number"

jonhoo: Ah, no, `--offsets`. I think the plural is important. That said, I'm okay with just keeping the long version and having it be `--includeoffset`.

The reason I'm pushing back on `-i` and `-o` is that I think it's likely we'll want to add arguments specifically for input and output files in the future, and those would be the obvious candidates for those arguments :)

src/collapse/dtrace.rs
jonhoo: @jasonrhansen I think if there is more than one `::`, then
```rust
// Remove arguments from C++ function names:
if let Some(scope) = frame.find("::") {
    if let Some(open) = frame[scope + 2..].rfind(|c| c == '(' || c == '<') {
	frame = &frame[..scope + 2 + open];
    }
}
```
(that is, with the `find` and `rfind` swapped) Is still correct, no?

jonhoo: Correct in the sense of matching the original regex that is :p

src/collapse/dtrace.rs
jasonrhansen: Sounds like I good idea to me!

src/bin/collapse-dtrace.rs
Licenser: 👍 `--includeoffset` only seems like the best compromise then!

src/collapse/dtrace.rs
jonhoo: https://github.com/jonhoo/inferno/issues/80

Licenser:
Good idea, noted in https://github.com/brendangregg/FlameGraph/issues/202

src/collapse/dtrace.rs
jasonrhansen: Actually, I guess you're right!

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,187 @@
...
+            opt: opt,
+            stack_str_size: 0,
+        }
+    }
+}
+
+impl Folder {
+    fn uncpp(probe: &str) -> &str {
+        if let Some(scope) = probe.rfind("::") {
+            if let Some(open) = probe[scope + 2..].find(|c| c == '(' || c == '<') {
```
jonhoo: Based on the conversation [here](https://github.com/jonhoo/inferno/pull/78#discussion_r264042450), I believe this should be:
```rust
if let Some(scope) = frame.find("::") {
    if let Some(open) = frame[scope + 2..].rfind(|c| c == '(' || c == '<') {
```
Notice the two swaps of `rfind` and `find`.

jonhoo:
I think the cpp demangling comment above is then the last thing to resolve before merging! :tada:
Oh, and possibly also the `Cargo.lock` conflict -- should be easy enough to merge with `master`.

src/collapse/dtrace.rs
Licenser: I think to find the last `::` is what we want. Given the comment, it is supposed to remove the arguments so the classpath should remain intact, not only the first class.

src/collapse/dtrace.rs
Licenser: I've added some tests, `find` vs `rfind` makes no difference in them, so both seems fine. I'll change it to `find`.

Licenser:
I rebased to master to get rid of the merge requiest 👍

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,203 @@
+use super::Collapse;
+use std::collections::{HashMap, VecDeque};
+use std::io;
+use std::io::prelude::*;
+
+// Most code should not have more then 10 calls in the java stack.
+const JAVA_VEC_SIZE: usize = 10;
```
jonhoo: I don't think we even need this -- just use `::new()` :)

src/collapse/dtrace.rs
```diff
@@ -0,0 +1,203 @@
...
+    stack: VecDeque<String>,
+
+    /// Number of times each call stack has been seen.
+    occurrences: HashMap<String, usize>,
+
+    /// Keep track of stack string size while we consume a stack
+    stack_str_size: usize,
+
+    /// vecotr for processing java stuff
+    java_vec: Vec<String>,
```
jonhoo:
```suggestion
    cache_inlines: Vec<String>,
```

src/collapse/dtrace.rs
jonhoo: Ah, no, the second one specifically needs to be `rfind` for this to correctly match the regex :)

tests/collapse-dtrace.rs
```diff
@@ -0,0 +1,98 @@
...
+    let test_file = "./tests/data/collapse-dtrace/test/java.txt";
+    let result_file = "./tests/data/collapse-dtrace/test/results/java.txt";
+    test_collapse_dtrace(test_file, result_file, Options::default()).unwrap()
+}
+
+#[test]
+fn collapse_dtrace_compare_to_flamegraph_bug() {
+    // There is a bug in flamegraph that causes the following stack to render
+    // badly. We fix this but keep the test around to point out this breakage
+    // of bug compatibility.
```
jonhoo: Maybe link to the bug report here too for posterity?

jonhoo:
Just for future reference: it's actually easier to continue the review if you merge than if you rebase :) We can always just squash at the end!

jonhoo:
Re 650af068489f7d17bbe0ea93286c2448146982e5, you probably have to rename the fields elsewhere too :)

tests/collapse-dtrace.rs
Licenser: 👍

src/collapse/dtrace.rs
Licenser: 👍

jonhoo:
Just making sure you saw https://github.com/jonhoo/inferno/pull/78#discussion_r264058670

src/collapse/dtrace.rs
Licenser: 👍

Licenser: That will re-add the <...> to the data, is that what we want? I'm genuenly unsure since I couldn't find any examples on C++ dtrace stack traces.

jonhoo: Yeah, let's stick with the way the original Perl regex operates here, which would be (I think) exactly like `find` followed by `rfind`. And then we can refine later in another PR if we come up with compelling reasons why it should be different.

Licenser: okay

jonhoo:
@Licenser Looks like the `cpp_test [no longer passes](https://travis-ci.org/jonhoo/inferno/jobs/504446660). Should be a quick fix!

Licenser:
Rebased and squashed for merge.

jonhoo:
Awesome work :tada:

For future reference, it's actually better if we do the squash as part of merging the PR, because otherwise I have to read through all the changes again just to make sure nothing weird happened as part of your rebase :)

PR #82, opened 2019-03-10
Remove SmallVec from collapse/perf

### LINKED ISSUE #80: Remove `SmallVec` in favor of using a long-term `Vec`
See https://github.com/jonhoo/inferno/pull/78#discussion_r264041253. Specifically, currently we use a `SmallVec` for temporarily holding inlined java method calls:

https://github.com/jonhoo/inferno/blob/10a1fd7c4c891dbc6a3c6ed74541d884eb89b1e9/src/collapse/perf.rs#L278-L279

However, in #78, @Licenser decided to instead re-use a `Vec` in the `Folder`, which seems like a better strategy since it generally won't allocate even if there are inlined method calls (unlike `SmallVec`, which will allocate every time that happens).

jasonrhansen:
Fixes #80

### DIFF
--- a/src/collapse/perf.rs
+++ b/src/collapse/perf.rs
@@ -64,8 +64,9 @@ pub struct Folder {
     /// Function entries on the stack in this entry thus far.
     stack: VecDeque<String>,
 
-    /// Keep track of inlined Java functions so they can be annotated with "_[i]".
-    java_inline: Vec<String>,
+    /// General String cache that can be used while processing lines.
+    /// Currently only used to keep track of functions for Java inlining.
+    cache_line: Vec<String>,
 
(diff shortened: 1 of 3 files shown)


src/collapse/perf.rs
```diff
@@ -65,6 +64,9 @@ pub struct Folder {
     /// Function entries on the stack in this entry thus far.
     stack: VecDeque<String>,
 
+    /// Keep track of inlined Java functions so they can be annotated with "_[i]".
+    java_inline: Vec<String>,
```
jonhoo: I'd like to call this `cache_line` or something, just because it doesn't specifically _need_ to be Java inlining (although in practice I suppose it basically always is).

PR #84, opened 2019-03-11
Add integration tests that test the CLI programs

jasonrhansen:
Also share common test code among collapse implementations.


jonhoo:
Hmm, looks like coverage now [times out](https://travis-ci.org/jonhoo/inferno/jobs/504780310#L629)? :/

jonhoo:
I'll try restarting it...

jonhoo:
Looks like the coverage now [times out](https://travis-ci.org/jonhoo/inferno/jobs/504780310#L627), even after a re-run :/ I wonder if it's because we now do a full project build from _inside_ the tests.. Maybe the better approach here would be to expose most of `main()` as a function that we can call from the tests instead, so that we don't have to re-build and go via the command-line?

jasonrhansen:
Yeah, that's unfortunate. I wish Cargo had a better way to test binaries. I'll get this fixed.

PR #86, opened 2019-03-12
Add "magic" collapse dispatcher (collapse-guess)

### LINKED ISSUE #81: Provide magic collapse-dispatcher
Once #78 lands, we'll have two different `Collapse` implementations. It'd be cool to have a `inferno-collapse-guess`, which attempts to use whichever `Collapse` implementation is appropriate for a given input! This may end up being a little tricky, but here's a sketch of how it might work:

 - Add a method `is_applicable(&str) -> Option<bool>` to `Collapse`. `None` means "not sure -- need more input", `Some(true)` means "yes, this implementation should work with that string, and `Some(false)` means "no, this implementation definitely won't work.
 - Add a new collapse implementation, "guess", which internally buffers the string it reads from the underlying streams, and knows about all the other `Collapse` impls (maybe using something like [`inventory`](https://crates.io/crates/inventory)?). For each line it reads, it calls `is_applicable` on all impls that haven't returned `Some(false)` thus far with the input it has collected. If any of them return `Some(true)`, it decides to use that impl, and then:
 - Implements `BufRead` for a struct that wraps a `String` plus a `BufRead`. The `String` is the string the guesser has read so far, and that's what it'll read from first. Once all of that string has been read, it proceeds to read from the `BufRead`.

jasonrhansen:
Fixes #81


src/collapse/guess.rs
```diff
@@ -0,0 +1,70 @@
...
+        let mut line_start = 0;
+        let mut buffer = String::new();
+        loop {
+            let bytes = reader.read_line(&mut buffer)?;
+            let line = &buffer[line_start..];
+
+            macro_rules! try_collapse_impl {
+                ($collapse:ident, $index:expr) => {
+                    if !not_applicable[$index] {
+                        match $collapse.is_applicable(line) {
```
jonhoo: Ah, I was actually thinking that `is_applicable` would get _all_ the input gathered thus far, rather than just the current line. That way we wouldn't need all the impls to keep track of what they've seen so far.

jasonrhansen: OK, that makes sense. If we're going to do it that way, should we read more than a single line at a time instead of sending all of the gathered input every time we get a line? Maybe read 5 or 10 lines each loop iteration?

jonhoo: Yeah, that seems like a reasonable plan!

jasonrhansen: It also makes the `is_applicable` implementations slightly more complex because now they have to iterate through the lines themselves. They'll need to keep track of the same information as before, but it can be stored in local variables instead, which I guess is the main benefit.

jonhoo: `is_applicable` is also less performance sensitive, so it could be that it can be reduced almost down to a single (multi-line) regex for many of them.

jasonrhansen: True, but I wanted to share parsing code with `is_applicable` so we wouldn't have to maintain two different ways of doing almost the same thing. Also it feels a little strange to bring in the regex crate when we don't even use it for core functionality.

But on the other hand it does save us the boilerplate of looping through the lines in every impl.

I'm going to first make it work without regexes, then we can decide if it's worth switching.

jonhoo: I guess the observation I was hoping we might make is that it is easier to check whether a given file is likely a valid file for a given collapser than it is to actually parse the full format. For example, for DTrace, you may be able to just parse the first few header lines. For perf, you may just look for whether you have an even line + a stack line, etc.

jasonrhansen: I'm going to push up a commit that does what your first comment says. Once we feel good about everything else, I can look into changing the `is_applicable` implementations.

jasonrhansen: Can we always count on DTrace to have the following as the first header line?
```
CPU     ID                    FUNCTION:NAME
```
I'm just not sure what we can count on in the header to know if it's valid. That's why I have it checking the actual stack and count lines since that's the part we'll be parsing in `collapse`. But if all DTrace input will have this line we can just compare to this, ignoring whitespace.

jonhoo: /cc @spacejam @Licenser ^

Licenser: No in many cases that's sadly not included provided :(

src/collapse/mod.rs
```diff
@@ -58,4 +61,11 @@ pub trait Collapse {
             }
         }
     }
+
+    /// Returns whether this implementation is appropriate for the given input.
+    ///
+    /// None means "not sure -- need more input".
+    /// Some(true) means "yes, this implementation should work with this string".
+    /// Some(false) means "no, this implementation definitely won't work".
```
jonhoo: Let's make this a list with ` - ` and add backticks around `None` and `Some(..)`

jonhoo:
This is starting to look really good! :D

src/collapse/guess.rs
jasonrhansen: That's too bad. What I have it doing now works, but I'm open to suggestions on making it better.

jonhoo: I wonder if the backticks are always there. Maybe look for the first empty line, then any number of lines that are either empty or start with `              ` and contain a backtick, and then a lines that contains the leading whitespace and a number?

jasonrhansen: That sounds reasonable to me.

jasonrhansen: I found one case where the backticks aren't there. I can also check for lines with hex numbers, but I wonder if there are other cases.
```
              libsystem_kernel.dylib`read+0xa
              libzip.dylib`ZIP_Read+0xd3
              libzip.dylib`Java_java_util_zip_ZipFile_read+0x60
              0x104e7945e
              0x105048088
              0x5a3ef800758e
                2
```

jonhoo: I think the backtick denotes "module", and address (`0x`) would be the fallback if the module/func can't be resolved. Maybe just look for the indent and ignore the line contents (beyond the count)?

jasonrhansen: Are you ok with the way it is now, where it detects a stack line as either a module/func if there's a backtick, or as an address if it starts with 0x and the rest can be parsed as a hexadecimal? I've added a test for detecting addresses. 

I could change it to ignore contents for stack lines, but that's what it was doing before. However it didn't check for an indent since the line was being trimmed.

jonhoo: Yes, this looks great!

PR #88, opened 2019-03-12
Add collapse benchmarks

Licenser:


jonhoo:
Should we run these on CI somehow? I'm a little sad that perf has to go through decoding, but it may be necessary if we don't want to pull in the entire decompressed file.

Licenser:
My experience is I wouldn't, for once they take a long time (3 minutes on my machine) and then especially free CI services are so unreliable performance wise that run to run variance is much to high to give meaningful results.

jonhoo:
Good point. I'd like to at least see a section in the `README` on how to run the benchmarks though!

In fact, thinking about it some more, maybe the perf benchmark should _first_ decompress and write the stacks to a temporary file, and then run the benchmark on that decompressed file.

Licenser:
The benchmark itself is purely run on memory, I wanted to take the IO out of it since we can't affect that part of the performance :).

README.md
```diff
@@ -93,6 +93,12 @@ It requires [hyperfine](https://github.com/sharkdp/hyperfine), and you
 must make sure you also check out Inferno's
 [submodules](https://github.blog/2016-02-01-working-with-submodules/).
 
+# Benchmarks
+Benchmarks are implemented using the [criterion](https://bheisler.github.io/criterion.rs/book/criterion_rs.html) library.
```
jonhoo:
```suggestion
Benchmarks are implemented using [criterion](https://github.com/bheisler/criterion.rs).
```

README.md
```diff
@@ -93,6 +93,12 @@ It requires [hyperfine](https://github.com/sharkdp/hyperfine), and you
...
+It should be noted that criterion saves runs data in `target/criterion` directory and by that shows run to run changes.
```
jonhoo:
```suggestion
Note that criterion saves its results in `target/criterion/`, and uses that to recognize changes in performance.
```

README.md
```diff
@@ -93,6 +93,12 @@ It requires [hyperfine](https://github.com/sharkdp/hyperfine), and you
...
+
+To run benchmarks simply run `cargo bench`.
```
jonhoo:
```suggestion
You can run the benchmarks with `cargo bench`.
```

jonhoo:
Ah, yes, I misread the code and though you were doing the I/O as part of the benchmark! Ignore me :sweat_smile: A few wording tweaks and I think we're good to go!

PR #90, opened 2019-03-13
Make working with flamegraph tests a little nicer

jonhoo:


jasonrhansen:
I love this! Are you going to do the same with the collapse tests?

jonhoo:
@jasonrhansen done!

jonhoo:
Also note that this _does_ change the external interface for `flamegraph`, but it does so in such a way that it is _more_ flexible. You now no longer need to pass a `Vec` specifically, and you also don't need to consume the `Options` (which'd be annoying since `Options` isn't `Clone`).

PR #93, opened 2019-03-13
Performance improvements on the dtrace collapser

Licenser:
Each change is one commit with perf delta appended.
Overall towards master:

```
collapse/dtrace         time:   [11.244 ms 11.398 ms 11.585 ms]
                        thrpt:  [113.80 MiB/s 115.68 MiB/s 117.25 MiB/s]
                 change:
                        time:   [-39.575% -37.504% -35.438%] (p = 0.00 < 0.05)
                        thrpt:  [+54.891% +60.011% +65.493%]
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  6 (6.00%) high mild
  1 (1.00%) high severe
```

### DIFF
--- a/src/collapse/dtrace.rs
+++ b/src/collapse/dtrace.rs
@@ -114,7 +114,7 @@ impl From<Options> for Folder {
     fn from(opt: Options) -> Self {
         Self {
             stack: VecDeque::default(),
-            occurrences: HashMap::default(),
+            occurrences: HashMap::with_capacity(100),
             cache_inlines: Vec::new(),
             opt,
             stack_str_size: 0,
--- a/src/collapse/dtrace.rs
+++ b/src/collapse/dtrace.rs
@@ -138,7 +138,7 @@ impl Folder {
     }
 
     fn remove_offset(line: &str) -> &str {
-        if let Some(offset) = line.find('+') {
+        if let Some(offset) = line.rfind('+') {
             &line[..offset]
         } else {
             line
@@ -153,7 +153,6 @@ impl Folder {
     //     unix`sys_syscall+0x10e
     //       1
     fn on_stack_line(&mut self, line: &str) {
-        let line = line.trim_start();
         let frame = if self.opt.includeoffset {
             line
         } else {
@@ -192,7 +191,7 @@ impl Folder {
             if first {
                 first = false
             } else {
-                stack_str.push_str(";");
+                stack_str.push(';');
             }
             //trim leaf offset if these were retained:
             if self.opt.includeoffset && i == last {
--- a/src/collapse/dtrace.rs
+++ b/src/collapse/dtrace.rs
@@ -142,6 +142,9 @@ impl Folder {
         let mut could_be_cpp = false;
         let mut has_semicolon = false;
         let mut last_offset = line.len() - 1;
+        // This seems risly but dtrace stacks are c-strings as can be seen in the function
+        // responsible for printing them:
+        // https://github.com/opendtrace/opendtrace/blob/1a03ea5576a9219a43f28b4f159ff8a4b1f9a9fd/lib/libdtrace/common/dt_consume.c#L1331
         let bytes = line.as_bytes();
         for offset in 0..bytes.len() {
             match bytes[offset] {
--- a/src/collapse/dtrace.rs
+++ b/src/collapse/dtrace.rs
@@ -114,7 +114,7 @@ impl From<Options> for Folder {
     fn from(opt: Options) -> Self {
         Self {
             stack: VecDeque::default(),
-            occurrences: HashMap::with_capacity(100),
+            occurrences: HashMap::with_capacity(512),
             cache_inlines: Vec::new(),
             opt,
             stack_str_size: 0,
@@ -194,12 +194,10 @@ impl Folder {
             while let Some(func) = self.cache_inlines.pop() {
                 self.stack.push_front(func);
             }
+        } else if has_colon {
+            self.stack.push_front(frame.replace(';', ":"))
         } else {
-            if has_colon {
-                self.stack.push_front(frame.replace(';', ":"))
-            } else {
-                self.stack.push_front(frame.to_owned())
-            }
+            self.stack.push_front(frame.to_owned())
         }
     }
 
--- a/src/collapse/dtrace.rs
+++ b/src/collapse/dtrace.rs
@@ -208,10 +208,10 @@ impl Folder {
     }
 
     fn finish<W: Write>(&self, mut writer: W) -> io::Result<()> {
-        let mut keys: Vec<_> = self.occurrences.keys().collect();
+        let mut keys: Vec<(&String, &usize)> = self.occurrences.iter().collect();
         keys.sort();
-        for key in keys {
-            writeln!(writer, "{} {}", key, self.occurrences[key])?;
+        for (key, count) in keys {
+            writeln!(writer, "{} {}", key, count)?;
         }
         Ok(())
     }
--- a/src/collapse/dtrace.rs
+++ b/src/collapse/dtrace.rs
@@ -137,12 +137,21 @@ impl Folder {
         }
     }
 
-    fn remove_offset(line: &str) -> &str {
-        if let Some(offset) = line.rfind('+') {
-            &line[..offset]
-        } else {
-            line
+    fn remove_offset(line: &str) -> (bool, bool, bool, &str) {
+        let mut could_be_java = false;
+        let mut could_be_cpp = false;
+        let mut has_colon = false;
+        let bytes = line.as_bytes();
+        for offset in 0..bytes.len() {
+            match bytes[offset] {
+                b'>' if offset > 0 && bytes[offset - 1] == b'-' => could_be_java = true,
+                b':' if offset > 0 && bytes[offset - 1] == b':' => could_be_cpp = true,
+                b';' => has_colon = true,
+                b'+' => return (could_be_java, could_be_cpp, has_colon, &line[..offset]),
+                _ => (),
+            }
         }
+        (could_be_java, could_be_cpp, has_colon, line)
     }
 
     // we have a stack line that shows one stack entry from the preceeding event, like:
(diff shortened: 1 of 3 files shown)


src/collapse/dtrace.rs
```diff
@@ -113,7 +114,7 @@ impl From<Options> for Folder {
     fn from(opt: Options) -> Self {
         Self {
             stack: VecDeque::default(),
-            occurrences: HashMap::default(),
+            occurrences: HashMap::with_capacity(100),
```
jonhoo: If this indeed results in a performance increase, then we should do this for `perf` too. Also, why 100 specifically?

src/collapse/dtrace.rs
```diff
@@ -136,15 +137,21 @@ impl Folder {
...
-        if let Some(first) = line.next() {
-            first
-        } else {
-            second
+    fn remove_offset(line: &str) -> (bool, bool, bool, &str) {
+        let mut could_be_java = false;
+        let mut could_be_cpp = false;
+        let mut has_colon = false;
+        let bytes = line.as_bytes();
+        for offset in 0..bytes.len() {
```
jonhoo: This seems really problematic to me, because it means that UTF-8 characters will break the parsing (the indexing below would crash). I think keeping Unicode support for a low-percentage performance penalty is worthwhile.

jonhoo: It could be that you can use `line.chars()` instead though, and still keep your loop below.

src/collapse/dtrace.rs
```diff
@@ -136,15 +137,21 @@ impl Folder {
...
+            match bytes[offset] {
+                b'>' if offset > 0 && bytes[offset - 1] == b'-' => could_be_java = true,
+                b':' if offset > 0 && bytes[offset - 1] == b':' => could_be_cpp = true,
+                b';' => has_colon = true,
+                b'+' => return (could_be_java, could_be_cpp, has_colon, &line[..offset]),
```
jonhoo: This is no longer the same as the old function -- you're now removing everything following the _first_ `+`, rather than the _last_. Don't know whether the two are different in practice though.

src/collapse/dtrace.rs
```diff
@@ -155,31 +162,44 @@ impl Folder {
...
-            if inline {
-                func.push_str("_[i]")
-            };
-            inline = true;
-            self.stack_str_size += func.len() + 1;
-            self.cache_inlines.push(func);
-        }
-        while let Some(func) = self.cache_inlines.pop() {
-            self.stack.push_front(func);
+        if could_be_java {
```
jonhoo: I'd probably rename this to `has_inlines`, since it isn't necessarily Java-specific. And also because if it contains `->`, it _does_ have inlines.

src/collapse/dtrace.rs
```diff
@@ -136,15 +137,21 @@ impl Folder {
...
+    fn remove_offset(line: &str) -> (bool, bool, bool, &str) {
+        let mut could_be_java = false;
+        let mut could_be_cpp = false;
+        let mut has_colon = false;
+        let bytes = line.as_bytes();
+        for offset in 0..bytes.len() {
+            match bytes[offset] {
+                b'>' if offset > 0 && bytes[offset - 1] == b'-' => could_be_java = true,
+                b':' if offset > 0 && bytes[offset - 1] == b':' => could_be_cpp = true,
+                b';' => has_colon = true,
```
jonhoo: Should this be `has_semicolon`?

src/collapse/dtrace.rs
```diff
@@ -155,31 +162,44 @@ impl Folder {
...
+            }
+            while let Some(func) = self.cache_inlines.pop() {
+                self.stack.push_front(func);
+            }
+        } else {
+            if has_colon {
+                self.stack.push_front(frame.replace(';', ":"))
+            } else {
+                self.stack.push_front(frame.to_owned())
+            }
```
jonhoo: Does this optimization actually make any performance difference in practice?

src/collapse/dtrace.rs
```diff
@@ -211,10 +231,10 @@ impl Folder {
     }
 
     fn finish<W: Write>(&self, mut writer: W) -> io::Result<()> {
-        let mut keys: Vec<_> = self.occurrences.keys().collect();
+        let mut keys: Vec<(&String, &usize)> = self.occurrences.iter().collect();
```
jonhoo: Do you really need the annotation here?

src/collapse/dtrace.rs
```diff
@@ -211,10 +231,10 @@ impl Folder {
...
         keys.sort();
```
jonhoo: So we're now relying on the fact that tuples are sorted in element order? I guess that's reasonable. May be worth adding a comment about it though.

jonhoo: Also, we should probably add this optimization to the perf collapser as well.

src/collapse/dtrace.rs
Licenser: Just gut feeling and experience, having something in the range of 100 elements in the stack seems reasonable. The example file has 400ish. I experimented a bit too, 1000 was significantly slower, so I picked something that seemed the right order of magnitude

src/collapse/dtrace.rs
Licenser: I'm fairly sure that dtrace ustack symbols can't be utf8.

src/collapse/dtrace.rs
jonhoo: This isn't the number of things in the stack though, it's the number of distinct observed total stacks, which I'd expect to be more than 100. :man_shrugging: I'm fine with ~100 if you've found that it empirically seems to be a good balance. I'd personally lean towards slightly higher + power of two, but I don't feel too strongly about it.

src/collapse/dtrace.rs
Licenser: I was wondering the same but in the examples I have couldn't find a case where a + was anywhere but at the end.

src/collapse/dtrace.rs
jonhoo: Is that so? I mean, maybe.. Is there a good reason not to use `.chars()` though?

src/collapse/dtrace.rs
jonhoo: Yeah, you're probably right that it's extremely uncommon if it ever happens. I doubt that looking through each string fully once carries all that much overhead though. Maybe we can quantify the difference? I'd _like_ to prefer the current semantics if the performance difference is minimal.

src/collapse/dtrace.rs
Licenser: Just that I go through the discussions top to bottom on commenting then a second time on testing :P

src/collapse/dtrace.rs
Licenser: 👍 much better

src/collapse/dtrace.rs
Licenser: 😂 yup!

src/collapse/dtrace.rs
Licenser: it does, it saves a full string rebuild and traversal on nearly all elements since ; isn't a common character

src/collapse/dtrace.rs
Licenser: ja I split them out one at a time so it's easy to track and carry over at the end

src/collapse/dtrace.rs
Licenser: tried a few more, 512 still doesn't make a noticeable difference so we can settle there :) still not at the point where I saw degradation so it should be good.

src/collapse/dtrace.rs
Licenser: Now there is a good reason:
```
collapse/dtrace         time:   [12.756 ms 12.946 ms 13.172 ms]
                        thrpt:  [100.09 MiB/s 101.84 MiB/s 103.36 MiB/s]
                 change:
                        time:   [+15.040% +17.531% +19.861%] (p = 0.00 < 0.05)
                        thrpt:  [-16.570% -14.916% -13.074%]
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high mild
``` 
:(

src/collapse/dtrace.rs
Licenser: nope 👍

src/collapse/dtrace.rs
Licenser: Tested: Going to the end makes very little (way below noise) difference (as the + is late in the line anyway) so lets do it propper :) 👍

src/collapse/dtrace.rs
jonhoo: Ouch, that's pretty severe. And surprising given that it's supposed to be an ASCII string, so there shouldn't even be any multi-byte characters to decode. I guess one thing we could do as a sanity check is also add an arm for each of the [leading code points](https://en.wikipedia.org/wiki/UTF-8#Description) of the multi-byte UTF-8 sequences. That way we'll at least detect if things are about to go terribly wrong.

Licenser: Okay so here comes the research results: dtrace stack parts are c-strings

Looking at [the function responsible for printing the stacktracek](https://github.com/opendtrace/opendtrace/blob/1a03ea5576a9219a43f28b4f159ff8a4b1f9a9fd/lib/libdtrace/common/dt_consume.c#L1385) all variables are `char *` not wide chars so it's c-strings and should be save to be addressed as bytes.

I added the link and an explanation to the code

jonhoo: Perfect! Thank you!

src/collapse/dtrace.rs
Licenser: All migrated, I'm good with this for now :)

jonhoo:
I kind of want to see some more of this (`has_inline` in particular) applied to the `perf` collapser too. We could probably tease out a decent bit more performance there as well :D

PR #94, opened 2019-03-14
Add benchmark for flamegraph

jasonrhansen:


benches/flamegraph.rs
```diff
@@ -0,0 +1,37 @@
...
+use std::io::{BufReader, Cursor, Read};
+
+fn flamegraph_benchmark(c: &mut Criterion) {
+    let mut f = File::open("tests/data/collapse-perf/results/example-perf-stacks-collapsed.txt")
+        .expect("file not found");
+
+    let mut bytes = Vec::new();
+    f.read_to_end(&mut bytes).expect("Could not read file");
+
+    let len = bytes.len();
```
jonhoo: This is the size of the input, not the output, yet you're passing it to `with_capacity` for `result`. That doesn't seem right?

benches/flamegraph.rs
```diff
@@ -0,0 +1,37 @@
+extern crate criterion;
+extern crate inferno;
+
+use criterion::*;
+use inferno::flamegraph;
+use std::fs::File;
+use std::io::{BufReader, Cursor, Read};
+
+fn flamegraph_benchmark(c: &mut Criterion) {
+    let mut f = File::open("tests/data/collapse-perf/results/example-perf-stacks-collapsed.txt")
```
jonhoo: I wonder if it might make sense to also define a second, identical benchmark that uses the collapsed DTrace output? Just so we have two data-points instead of one. Alternatively, we could even flamegraph across the concatenation of the two.

benches/flamegraph.rs
jasonrhansen: Yeah, it doesn't seem right. I copied this from the collapse benchmarks. I can fix it in those, too. I'll just use `Vec::new()` for `result`.

benches/collapse.rs
```diff
@@ -2,30 +2,38 @@ extern crate criterion;
...
+use inferno::collapse::dtrace;
+use inferno::collapse::perf;
 use inferno::collapse::Collapse;
 use libflate::gzip::Decoder;
 use std::fs::File;
 use std::io::{BufReader, Cursor, Read};
 
-fn dtrace_benchmark(c: &mut Criterion) {
-    let mut f = File::open("flamegraph/example-dtrace-stacks.txt").expect("file not found");
+fn collapse_benchmark<C, S>(c: &mut Criterion, mut collapser: C, id: S, infile: &str)
```
jonhoo: I think we can safely just do `id: &str` here :p

benches/flamegraph.rs
jonhoo: So, I'm not quite sure that's what we want to do either, because it means we'll also be measuring a bunch of re-allocations of that `Vec` as it grows. I think the trick is going to use `fs::metadata().len()` on a location that we know is the resulting, folded file.

jasonrhansen: Even better, why not just use `io::sink()` for the result since we don't even need the result anyway?

jonhoo: That's a great idea! Let's do the same in the others too.

PR #96, opened 2019-03-15
Port difffolded.pl as inferno-diff-folded

### LINKED ISSUE #62: Port difffolded.pl
Now that #60 has landed, it would be nice if inferno could generate the differential profile files that this feature expects. Currently you need to use [difffolded.pl](https://github.com/brendangregg/FlameGraph/blob/master/difffolded.pl) from upstream FlameGraph. We should port this to Rust.

See [Differential Flame Graphs](http://www.brendangregg.com/blog/2014-11-09/differential-flame-graphs.html) for more info.

jasonrhansen:
Fixes #62

I also changed flamegraph to always sort input lines since the output of diff-folded isn't sorted. This matches the behavior of upstream flamegraph.


src/lib.rs
```diff
@@ -121,3 +135,10 @@ pub mod collapse;
 ///
 ///   [crate-level documentation]: ../index.html
 pub mod flamegraph;
+
+/// Tool for creating an output required to generate differential flame graphs.
+///
+/// See the [crate-level documentation] for details.
+///
+///   [crate-level documentation]: ../index.html
+pub mod diff_folded;
```
jonhoo:
```suggestion
pub mod differential;
```

src/lib.rs
```diff
@@ -77,6 +77,19 @@
...
+//! They let you easily visualize the differences between two profiles performed before and
+//! after a code change. See Brendan Gregg's [differential flame graphs] blog post for a great
+//! writeup. To create one you must first pass the two folded stack files to `inferno-diff-folded`,
+//! then send the output to `inferno-flamegraph`. Example:
+//!
+//! ```console
+//! $ inferno-diff-folded stacks1.folded stacks2.folded | inferno-flamegraph > differential.svg
+//! ```
+//!
+//! And then open `differential.svg` in your viewer of choice.
```
jonhoo:
```suggestion
//! And then open `differential.svg` in your viewer of choice.

```

tests/data/flamegraph/grey-frames/grey-frames.txt
```diff
@@ -1,5 +1,5 @@
-cksum;_start;__libc_start_main;main;-- 31
 cksum;- 6
+cksum;_start;__libc_start_main;main;-- 31
```
jonhoo: Why this ordering change?

jasonrhansen:
I'm bothered enough about the sorting this PR adds to flamegraph that I think we should provide a way to opt out if you know the input is already sorted. We could make compatibility with upstream the default, but have a flag like `--no-sort` which is basically a promise that the input is already sorted. What do think?

src/bin/diff-folded.rs
```diff
@@ -0,0 +1,64 @@
+use env_logger::Env;
+use std::io;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+use inferno::diff_folded::{self, Options};
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "inferno-diff-folded", author = "")]
+struct Opt {
```
jonhoo: I'd like to see some of the help docs from [upstream](https://github.com/brendangregg/FlameGraph/blob/1b1c6deede9c33c5134c920bdb7a44cc5528e9a7/difffolded.pl#L10L26) printed both by the CLI and in the Rust library docs.

tests/data/flamegraph/grey-frames/grey-frames.txt
jasonrhansen: This is because of the change in flamegraph to sort lines.

src/bin/diff-folded.rs
jonhoo: There are also a few lines [here](https://github.com/brendangregg/FlameGraph/blob/1b1c6deede9c33c5134c920bdb7a44cc5528e9a7/difffolded.pl#L68L70) that may be good to include

src/diff_folded/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+    R1: BufRead,
+    R2: BufRead,
+    W: Write,
+{
+    let mut stack_counts = HashMap::new();
+    let total1 = parse_stack_counts(&opt, &mut stack_counts, reader1, true)?;
+    let total2 = parse_stack_counts(&opt, &mut stack_counts, reader2, false)?;
+    if opt.normalize && total1 != total2 {
+        for counts in stack_counts.values_mut() {
+            counts.first = counts.first * total2 / total1;
```
jonhoo: I wonder whether we may want to do this computation using floating-point?

src/diff_folded/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+#[derive(Default)]
+struct Counts {
+    first: usize,
+    second: usize,
+}
+
+/// Configure the generated output.
+///
+/// All options default to off.
+#[derive(Debug, Default)]
```
jonhoo:
```suggestion
#[derive(Debug, Clone, Default)]
```

src/diff_folded/mod.rs
```diff
@@ -0,0 +1,158 @@
+use std::collections::HashMap;
+use std::fs::File;
+use std::io;
+use std::io::prelude::*;
+use std::path::Path;
+
+const READER_CAPACITY: usize = 128 * 1024;
+
+#[derive(Default)]
```
jonhoo:
```suggestion
#[derive(Debug, Clone, Copy, Default)]
```

src/diff_folded/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+        writeln!(writer, "{} {} {}", stack, first, second)?;
+    }
+    Ok(())
+}
+
+// Parse stack and sample count from line.
+fn parse_line(line: &str, strip_hex: bool) -> Option<(String, usize)> {
+    let counti = line.rfind(' ')?;
+    let count = &line[(counti + 1)..].trim_end();
+    let count = count.parse::<usize>().ok()?;
```
jonhoo: Won't this choke on fractional samples?

src/diff_folded/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+    let mut stack = line[..counti].trim_end().to_string();
+    if strip_hex {
+        stack = strip_hex_address(stack);
```
jonhoo: I think this ends up allocating two strings every time? Once for `.to_string()` above, and once inside `strip_hex_address`? If we make `strip_hex_address` take `&str`, I _think_ we should be able to get rid of one of those.

src/lib.rs
jasonrhansen: Do you also want to change the name of the binary?

src/diff_folded/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+        let ndigits = stack[start + idx + 2..]
+            .chars()
+            .take_while(|c| c.is_digit(16))
+            .count();
+        if ndigits > 0 {
+            stack = format!(
+                "{}0x...{}",
+                &stack[..start + idx],
+                &stack[start + idx + 2 + ndigits..]
+            );
```
jonhoo: It might be that it'd be faster to just `push` each character as you iterate past it rather than do multiple `format!` calls. You can even do `String::with_capacity(stack.len())` at the beginning to ensure it won't allocate!

tests/diff-folded.rs
```diff
@@ -0,0 +1,159 @@
...
+    let rand: u64 = rand::random();
+    let tm = std::env::temp_dir().join(format!("test-{}.svg", rand));
+    if fs::write(&tm, result.get_ref()).is_ok() {
+        eprintln!("test output in {}", tm.display());
+    }
+    // and then compare
+    result.set_position(0);
+    compare_results(result, expected, expected_result_file);
+    Ok(return_value)
+}
```
jonhoo: We really ought to find a way to share this code across all of our different tests. I _think_ it should be doable by passing in a closure that goes from input(s) to output.

tests/diff-folded.rs
```diff
@@ -0,0 +1,159 @@
...
+        .zip(expected_lines.into_iter())
+        .enumerate()
+    {
+        assert_eq!(
+            result_line, expected_line,
+            "\n{}:{}",
+            expected_file, line_num
+        );
+    }
+}
```
jonhoo: This too I sort of wish we could extract out and share. Not sure how easy that'll be though..

tests/diff-folded.rs
```diff
@@ -0,0 +1,159 @@
...
+    let sink = io::sink();
+    let _ = diff_folded::from_readers(&mut options, r1, r2, sink);
+    testing_logger::validate(asserter);
+}
+
+#[test]
+fn diff_folded_default() {
+    let infile1 = "./tests/data/diff-folded/before.txt";
+    let infile2 = "./tests/data/diff-folded/after.txt";
+    let expected_result_file = "./tests/data/diff-folded/results/default.txt";
```
jonhoo: And just to be paranoid, you've checked all of the `results/` files against the version produced by upstream, right? I almost wonder whether we should have a script that re-generates all of the things in `results` from upstream.. But that's for another time.

src/lib.rs
jonhoo: Hmm, that's a good question... No, I think let's leave that the way it is.

tests/data/flamegraph/grey-frames/grey-frames.txt
jonhoo: But this is the _input_ to `flamegraph`, right? Or is that you specifically want to test and see that it sorts? It does seem to change the output of one SVG.

jonhoo:
@jasonrhansen I think something like `--no-sort` is a great idea. We may then want to warn if we detect unsorted input though. That can be an entirely local check. Just make sure that each line lexiographically follows the previous one. Maybe even do it really stupidly by only checking the first character.

src/diff_folded/mod.rs
jasonrhansen: Yup, sure will.

tests/diff-folded.rs
jasonrhansen: Yes, I've thought of doing that. I just haven't gotten to it yet.

tests/diff-folded.rs
jasonrhansen: This one is quite a bit different than the other ones though because we sort the outputs before comparing. And because we have to read in all of the lines before we sort, we can compare the lengths of the outputs at the beginning.

tests/diff-folded.rs
jasonrhansen: The files in `/results` were actually produced by upstream, so yes.

jasonrhansen:
I like the idea of detecting and warning about unsorted input when using `--no-sort`.

tests/data/flamegraph/grey-frames/grey-frames.txt
jasonrhansen: I guess this file didn't need to be changed. What happened is the SVG output changed, and I realized it was because it's sorting lines now. I don't know why the original input file wasn't sorted before; it must have been hand-edited. I updated both the SVG file and the input file, but I really only had to change the SVG. The result should be the same either way. I can change it back if you want, though.

jonhoo: Nah, I think it's fine -- just wanted to understand the diff. We may want a test along with this change that unsorted and sorted inputs both produce the same SVG though!

src/diff_folded/mod.rs
jasonrhansen: I changed the sample count type to `f64` so it can handle fractional samples, but now the formatting isn't matching the Perl version.

Rust: `14.234398324872284`
Perl: `14.2343983248723`

Is there a built-in way in Rust to format with a maximum number of decimal places, but with no trailing zeroes?

jonhoo: Ah, sorry, I was thinking we should ditch the fractional part [like we do in `flamegraph`](https://github.com/jonhoo/inferno/blob/643cf821ee271e968259fafd1499130186499c59/src/flamegraph/merge.rs#L191L196).

jasonrhansen: OK, I guess if we're already different in flamegraph, it doesn't hurt here either.

tests/diff-folded.rs
jasonrhansen: I'm leaving this for a separate PR. There's enough differences among `flamegraph`, `collapse`, and `diff-folded` that make it a bit complicated.

- Collapse uses `collapser.collapse` if the input file is a gzip, but `collapser.collapse_file` otherwise.
- They all have incompatible implementations of `compare_results`.
- `diff-folded` requires exactly two input files, but the others don't.

It would be nice to share code across these tests, but I don't think we can do it with a single closure. Maybe we could solve it using a trait?

tests/flamegraph.rs
```diff
@@ -592,3 +592,40 @@ fn load_palette_map_file(palette_file: &str) -> PaletteMap {
     let path = Path::new(palette_file);
     PaletteMap::load_from_file_or_empty(&path).unwrap()
 }
+
+#[test]
+fn flamegraph_unsorted_input_files() {
+    let input_file =
+        "./tests/data/flamegraph/multiple-inputs/perf-vertx-stacks-01-collapsed-all-unsorted-1.txt";
+    let expected_result_file = "./tests/data/flamegraph/unsorted-input/unsorted-input.svg";
```
jonhoo: I don't think you need a _separate_ SVG for this. You should be able to just compare it to the one produced from the sorted version of the input, no?

jasonrhansen: The input file is half of a two-part input that's used in the multiple-inputs test. I reused it because it was already unsorted, but it only has half of the stacks of the original test file from upstream. We don't have another SVG to compare to.

The alternative is to create a new input file with all of the stack lines and randomize the order of those, then compare to the one produced by the sorted version like you're suggesting. I guess this option has the added advantage of making sure the outputs of the sorted and unsorted versions always match. I'll change the test to do this.

tests/data/flamegraph/grey-frames/grey-frames.txt
jasonrhansen: I pushed a commit that uses the same test SVG for 3 different tests.

1. Multiple inputs, unsorted
2. Single input, unsorted
3. Single input, sorted

This way we know they all produce the same output.

src/flamegraph/merge.rs
```diff
@@ -115,12 +115,23 @@ where
...
         let mut line = line.trim();
         if line.is_empty() {
             continue;
         }
 
+        if !unsorted_detected {
+            if let Some(prev_line) = prev_line {
+                if prev_line > line {
+                    unsorted_detected = true;
+                    warn!("Unsorted input lines detected");
```
jonhoo: I think we should just return if we detect unsorted input. The output will already be messed up.

src/differential/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+///
+/// The readers are expected to contain folded stack lines of before and after profiles with
+/// the following whitespace-separated fields:
+///
+///  - A semicolon-separated list of frame names (e.g., `main;foo;bar;baz`).
+///  - A sample count for the given stack.
+///
+/// The output written to the `writer` will be similar to the inputs, except there will be two
+/// sample count columns -- one for each profile.
+pub fn from_readers<R1, R2, W>(opt: Options, reader1: R1, reader2: R2, writer: W) -> io::Result<()>
```
jonhoo: I wonder if we should give `reader1` and `reader2` better names, or at least mention those names explicitly in the text above, so the reader knows which one is "before" and which one is "after".

jasonrhansen: Should we just name the `before` and `after`?

src/flamegraph/merge.rs
jasonrhansen: Good point.

jasonrhansen: So should we just return the values we have up to this point then after logging an error?

jonhoo: I think we should actually just return a full-out `Err`, since the results just won't be useful at that point.

jasonrhansen: Ok, so you want to change the return type to a `Result`? What about the error type? We don't currently have a custom error type for flamegraph. `from_sorted_lines` just returns `quick_xml::Result<()>`. Do we want to add a `flamegraph::Result` type that it returns instead?

PR #98, opened 2019-03-16
WIP: Split out the dtrace parser and collaps

Licenser:
I started to scatch out how a split out parse step as mentioned in #30 could look like.

The dtrace collapse is effectively generic so it's just 'collapse'.

Note 1:

For the start, I'll iron this out for DTrace and then port the result to perf.

Note 2:
At this point, this comes at a performance penalty for the
workflow of pure collapsing. However, the split has undergone
no performance work at this point and we are not measuring
the difference of full flamegraph creation that could now save
a whole set of encoding and decoding stack traces.


jonhoo:
This will probably need some work to combine nicely with #128, but would be great to land!

PR #99, opened 2019-03-17
Add support for --reverse option

### LINKED ISSUE #21: Reversed stack ordering
We should support the [`--reverse` option](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L148), which, well, [reverses the order of all stacks](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L586L597). Note that this _does_ require that we also [sort our input](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L607), because otherwise the trick we play in `flow` in `flamegraph/merge.rs` doesn't work!

jasonrhansen:
Fixes #21


jasonrhansen:
Currently the performance when using the `--reverse` option is terrible. I'd like to figure out a way to reduce allocations.

jasonrhansen:
I reduced allocations, but didn't see a huge speedup.

It looks like the main reason the performance can be much worse with the `--reverse` option is not from parsing and reversing the stacks. Of course that will slow it down some, and we can probably make it faster, but no matter how much we optimize that part it won't help much.

The big slowdown is because not many frames get merged when the stack order is reversed. It makes sense and is really obvious when you look at the resulting flame graph.

image

As an example, when creating a flame graph using `tests/data/collapse-perf/results/example-perf-stacks-collapsed.txt` as input, `merge::frames` returns only 1,139 frames without `--reverse`, but 18,247 frames with `--reverse`. That's a big difference.

Just to make sure that's what was making it slower, I created a version of `example-perf-stacks-collapsed.txt` where the stack orders were already reversed. I benchmarked flamegraph using this as input and it was almost as bad as the benchmark where I used the original file and passed the `--reverse` flag. I also benchmarked using the reversed file AND the `--reverse` flag and it was pretty fast. Not as fast as the original benchmark without `--reverse`, but not much worse.

jonhoo:
It'd be interesting to see _why_ we're so much slower when producing many frames. Surely it isn't _just_ from having to write more bytes to the output? If so, maybe we need a `BufWriter`?

jasonrhansen:
Yes, it would be worth investigating why producing more frames is so much slower. It would be nice to improve the performance overall. I was just initially surprised when I benchmarked with the `--reverse` option and the performance difference was much larger than I expected. I was thinking I must be doing something really bad.

jasonrhansen:
I tried using a `BufWriter` and it actually slowed it down. 

Also, just to be clear, `inferno-flamegraph --reverse` is still so fast on a single run that it might not even matter. When I run it on `example-perf-stacks-collapsed.txt` it appears to finish instantly. It's only when I run the benchmark (with 5050 iterations) that I can see it's much slower.
```
flamegraph/dtrace       time:   [7.6151 ms 7.6532 ms 7.6919 ms]
                        thrpt:  [9.8767 MiB/s 9.9267 MiB/s 9.9763 MiB/s]
                 change:
                        time:   [+466.22% +471.31% +476.06%] (p = 0.00 < 0.05)
                        thrpt:  [-82.641% -82.496% -82.339%]
                        Performance has regressed.

flamegraph/perf         time:   [41.598 ms 41.804 ms 42.030 ms]
                        thrpt:  [14.669 MiB/s 14.748 MiB/s 14.821 MiB/s]
                 change:
                        time:   [+721.96% +727.51% +733.03%] (p = 0.00 < 0.05)
                        thrpt:  [-87.996% -87.916% -87.834%]
                        Performance has regressed.
```

jonhoo:
This is benchmarking `--reverse` against _not_ `--reverse`, or just against the codebase as of this PR?

jasonrhansen:
That was benchmarking `--reverse` against _not_ `--reverse` on this branch.

Here is this branch against master (not using `--reverse` with either one).
```
flamegraph/dtrace       time:   [1.3086 ms 1.3230 ms 1.3421 ms]
                        thrpt:  [56.606 MiB/s 57.425 MiB/s 58.056 MiB/s]
                 change:
                        time:   [-3.5107% -2.5537% -1.6082%] (p = 0.00 < 0.05)
                        thrpt:  [+1.6344% +2.6206% +3.6385%]
                        Performance has improved.

flamegraph/perf         time:   [4.8170 ms 4.8403 ms 4.8662 ms]
                        thrpt:  [126.69 MiB/s 127.37 MiB/s 127.99 MiB/s]
                 change:
                        time:   [-2.0915% -1.4600% -0.8109%] (p = 0.00 < 0.05)
                        thrpt:  [+0.8175% +1.4817% +2.1362%]
                        Change within noise threshold.
```

src/flamegraph/mod.rs
```diff
@@ -149,6 +150,12 @@ pub struct Options<'a> {
     /// make sure the lines are sorted.
     pub no_sort: bool,
 
+    /// Generate stack-reversed flame graph.
+    ///
+    /// Note that stack lines must always be sorted after reversing the stacks so the `no-sort`
```
```suggestion
    /// Note that stack lines must always be sorted after reversing the stacks so the `no_sort`
```

src/bin/flamegraph.rs
```diff
@@ -116,7 +116,14 @@ struct Opt {
     /// If you set this flag you need to be sure your
     /// input stack lines are already sorted.
     #[structopt(long = "no-sort")]
-    pub no_sort: bool,
+    no_sort: bool,
+
+    /// Generate stack-reversed flame graph.
+    /// Note that stack lines must always be sorted
+    /// after reversing the stacks so the no-sort
```
```suggestion
    /// after reversing the stacks so the `no_sort`
```

src/flamegraph/mod.rs
```diff
@@ -325,19 +333,51 @@ impl Rectangle {
...
 #[allow(clippy::cyclomatic_complexity)]
-pub fn from_sorted_lines<'a, I, W>(opt: &mut Options, lines: I, writer: W) -> quick_xml::Result<()>
+pub fn from_lines<'a, I, W>(opt: &mut Options, lines: I, writer: W) -> quick_xml::Result<()>
 where
     I: IntoIterator<Item = &'a str>,
     W: Write,
 {
-    let (bgcolor1, bgcolor2) = color::bgcolor_for(opt.bgcolors, opt.colors);
+    let mut reversed = StrStack::new();
+    let (mut frames, time, ignored, delta_max) = if !opt.no_sort && !opt.reverse_stack_order {
```
jonhoo: Can we just do `let lines = if ...` and then place the call to `merge::frames` on its own line after? I think that'd be more readable.

src/flamegraph/mod.rs
```diff
@@ -325,19 +333,51 @@ impl Rectangle {
...
+        // Sort lines by default.
+        let mut lines: Vec<&str> = lines.into_iter().collect();
+        lines.sort_unstable();
+        merge::frames(lines)?
+    } else if opt.reverse_stack_order {
+        if opt.no_sort {
+            warn!("Input lines are always sorted when using --reverse. The --no-sort flag is being ignored.");
```
jonhoo: This should already be enforced by the CLI opt parser, so this error will only appear for someone calling `from_lines` directly and setting `no_sort` and `reverse` in `Options`. We should probably warn about those fields instead of talking about command-line flags.

src/flamegraph/mod.rs
```diff
@@ -733,3 +758,22 @@ fn filled_rectangle<W: Write>(
     }
     svg.write_event(&cache_rect)
 }
+
+// Tries to find a sample count at the end of a line and returns its index.
+fn rfind_samples(line: &str) -> Option<usize> {
```
jonhoo: Is there a way we can combine this with the corresponding function for extracting the sample used when we produce the flamegraph?

src/bin/flamegraph.rs
jasonrhansen: This gets printed in the help text of the CLI, and the name of the option to the user is `no-sort` with a dash, not an underscore.

src/flamegraph/mod.rs
```diff
@@ -325,19 +333,51 @@ impl Rectangle {
...
-pub fn from_sorted_lines<'a, I, W>(opt: &mut Options, lines: I, writer: W) -> quick_xml::Result<()>
+pub fn from_lines<'a, I, W>(opt: &mut Options, lines: I, writer: W) -> quick_xml::Result<()>
 where
     I: IntoIterator<Item = &'a str>,
     W: Write,
 {
-    let (bgcolor1, bgcolor2) = color::bgcolor_for(opt.bgcolors, opt.colors);
+    let mut reversed = StrStack::new();
+    let (mut frames, time, ignored, delta_max) = if !opt.no_sort && !opt.reverse_stack_order {
+        // Sort lines by default.
```
jonhoo: It's a little weird to me for the default case to come first and with a bunch of conditions. Maybe we should do:
```rust
if opt.reverse_stack_order
else if opt.no_sort
else
```

src/flamegraph/mod.rs
```diff
@@ -325,19 +333,51 @@ impl Rectangle {
...
+        let mut reversed: Vec<&str> = reversed.iter().collect();
+        reversed.sort_unstable();
+        merge::frames(reversed)?
+    } else {
+        // Lines don't need sorting.
+        merge::frames(lines)?
+    };
 
-    let mut buffer = StrStack::new();
-    let (mut frames, time, ignored, delta_max) = merge::frames(lines)?;
```
jonhoo: I'm not sure why these lines had to change?

src/flamegraph/mod.rs
```diff
@@ -608,33 +649,21 @@ where
...
-        .map_err(quick_xml::Error::Io)?;
-
-    if opt.no_sort {
-        from_sorted_lines(opt, input.lines(), writer)
-    } else {
-        let mut lines: Vec<&str> = input.lines().collect();
-        lines.sort_unstable();
-        from_sorted_lines(opt, lines, writer)
-    }
+    from_readers(opt, iter::once(reader), writer)
```
jonhoo: I _think_ this change means that even with `no_sort`, the entire file gets read into memory before we do any processing. That seems a little unfortunate, though maybe it's a cost we're willing to pay? It means that if I have a _huge_ trace file (like, many GB, which isn't unreasonable with debug symbols) that I know is sorted, I may be unable to plot it at all since it won't fit in memory all at once.

src/bin/flamegraph.rs
jonhoo: Ah, yes, you're right, I missed that this is in the CLI.

src/flamegraph/mod.rs
jasonrhansen: But in the `else` we don't create a `Vec`, but we do in the other two blocks. So the if and else would have incompatible types.

jonhoo: Ohhh, I missed that, you are completely right of course. Never mind me :angel:

src/flamegraph/mod.rs
jasonrhansen: The `buffer` is not being used until we start writing the SVG, so I moved it down to right before that.

The second line was replaced by all of the stuff above.

src/flamegraph/mod.rs
jasonrhansen: Good point, I completely overlooked that.

jasonrhansen: It looks like it was already reading the entire file into memory before the change. I guess there are a couple of reasons for this.
- `from_reader` takes a `Read`, not a `BufRead`
- Even if it took a `BufRead`, the `lines` method returns an iterator that produces `io::Result<String>`, so it can't be passed to `from_lines`.

jonhoo: I guess we require `Iterator<Item = &str>`, so that still assumes you have all the strings still lying around somewhere. This may be worth thinking harder about, but not relevant to this PR :thinking:

src/flamegraph/mod.rs
```diff
@@ -345,21 +345,20 @@ where
...
-    let (mut frames, time, ignored, delta_max) = if !opt.no_sort && !opt.reverse_stack_order {
-        // Sort lines by default.
-        let mut lines: Vec<&str> = lines.into_iter().collect();
-        lines.sort_unstable();
-        merge::frames(lines)?
-    } else if opt.reverse_stack_order {
+    let (mut frames, time, ignored, delta_max) = if opt.reverse_stack_order {
         if opt.no_sort {
-            warn!("Input lines are always sorted when using --reverse. The --no-sort flag is being ignored.");
+            warn!("Input lines are always sorted when `reverse_stack_order` is `true`. The `no_sort` option is being ignored.");
```
jonhoo: Just FYI, you can break strings using `\` at the end of the line :)

jasonrhansen: I knew you could, I just wasn't thinking about it because I had line wrapping on.

jonhoo:
@jasonrhansen fwiw, I opened #102.

PR #103, opened 2019-03-19
Unify defaults across CLI and impl Default

### LINKED ISSUE #91: Unify defaults across CLI and impl Default
#52 introduces a lot of default values, and they are duplicated between the `impl Default for Options` and the CLI `StructOpt` definitions. We should find a way to declare these as global constants, and then use those in both places. There's been a bunch of discussion of that in https://github.com/jonhoo/inferno/pull/52#discussion_r262319114, which should provide a good starting point.

jasonrhansen:
Fixes #91

One problem this PR doesn't address is the fact that the default values are also in the doc comments of the library's `Options` struct. It would be really nice if these could also use the constants. Unfortunately solving this isn't as easy. We could probably do it with an attribute macro, but is it worth it?

### DIFF
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -26,15 +26,6 @@ const FRAMEPAD: usize = 1; // vertical padding for frames
 
 /// Default values for [`Options`].
 pub mod defaults {
-    /// Default value with a string representation
-    pub struct DefaultValue<T> {
-        /// Default value
-        pub value: T,
-
-        /// Default value string representation
-        pub strval: &'static str,
-    }
-
     macro_rules! doc {
         ($str:expr, $($def:tt)*) => {
             #[doc = $str]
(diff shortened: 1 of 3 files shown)


jonhoo:
I think the way to get around that is to make the `defaults` module public and not `#[doc(hidden)]`, and then use https://github.com/rust-lang/rust/issues/43466 to link to the defaults in the docs.

jasonrhansen:
I changed the design a bit. I didn't like having the `DefaultValue` struct so I got rid of it. I put the &str values of the constants in a submodule. It makes the code and the docs cleaner.

`defaults::MIN_WIDTH.value` becomes `defaults::MIN_WIDTH`
and
`defaults::MIN_WIDTH.strval` becomes `defaults::str::MIN_WIDTH`

The `defaults` module has docs, but `defaults::str` is `#[doc(hidden)]`.

src/flamegraph/mod.rs
```diff
@@ -96,64 +105,62 @@ pub struct Options<'a> {
     /// Whether to plot a plot that grows top-to-bottom or bottom-up (the default).
     pub direction: Direction,
 
-    /// The search color for flame chart.
+    /// The search color for flame graph.
     ///
-    /// Defaults to `rgb(230,0,230)`.
+    /// [Default value](defaults::SEARCH_COLOR)
```
jonhoo: Missing a `.` at the end.

jonhoo:
Perfect! Just those missing full stops and then I think we're good to merge!

PR #104, opened 2019-03-19
Fix bug with &str defaults in CLI options

jasonrhansen:
For all of the default constants, two different versions are created.
One in `flamegraph::defaults` and the other in
`flamegraph::defaults::str`. Unfortunately, because of the way the macro
for defining the constants works, if the type of the constant is `&str`, the
version in `flamegraph::defaults::str` includes the quotes of the string
literal in the string constant.

To fix this, I used the version of the constants in `flamegraph::defaults` for
all of the &str constants.

### DIFF
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -29,7 +29,7 @@ struct Opt {
     #[structopt(
         short = "c",
         long = "colors",
-        raw(default_value = "defaults::str::COLORS"),
+        raw(default_value = "defaults::COLORS"),
         raw(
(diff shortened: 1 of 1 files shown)


jasonrhansen:
I tried to get the macro to treat the `&str` constants differently, but so far I haven't found a good way to do it.

jonhoo:
Oh, yeah, that makes sense, but is definitely pretty unfortunate. It probably also means that if you try to use `_` as a separator or add type ascriptions to numbers, you'll have problems with the `str` version..

jasonrhansen:
Yes, we need to be really careful with constants.

Also, I really wish we had a good way to test the CLI directly to help prevent bugs like this in the future.

PR #107, opened 2019-03-21
Add tests for flamegraph CLI options

jasonrhansen:
The `default_options` test makes sure the defaults of the CLI match the defaults of the library.

The `options` test checks that the command line options get set correctly on the `Options` of the library.

The tests also found a bug where the `--title` option didn't actually do anything.


jonhoo:
That's awesome! And yay for catching that bug :tada:

jonhoo:
I wonder if we might be able to use something like https://github.com/assert-rs/assert_cmd to take this even one step further.

jasonrhansen:
Yes, it looks like that would do the trick!

PR #108, opened 2019-03-21
Cache SVG event for text elements

jasonrhansen:
```
flamegraph/dtrace       time:   [1.1678 ms 1.1783 ms 1.1897 ms]
                        thrpt:  [63.858 MiB/s 64.475 MiB/s 65.056 MiB/s]
                 change:
                        time:   [-14.732% -13.694% -12.801%] (p = 0.00 < 0.05)
                        thrpt:  [+14.680% +15.866% +17.277%]
                        Performance has improved.

flamegraph/perf         time:   [4.4693 ms 4.4966 ms 4.5231 ms]
                        thrpt:  [136.30 MiB/s 137.11 MiB/s 137.94 MiB/s]
                 change:
                        time:   [-12.899% -12.014% -11.197%] (p = 0.00 < 0.05)
                        thrpt:  [+12.608% +13.654% +14.810%]
                        Performance has improved.
```

PR #109, opened 2019-03-21
do not reallocate start event

andy128k:
The same as https://github.com/jonhoo/inferno/pull/108 but without changing interfaces

### DIFF
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -439,9 +439,6 @@ where
         Writer::new(writer)
     };
 
-    // reuse for all text elements to avoid allocations
-    let mut cache_text = Event::Start(BytesStart::owned_name("text"));
-
     if time == 0 {
         error!("No stack counts found");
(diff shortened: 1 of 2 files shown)


jonhoo:
Oooh, that's interesting... This might be something we could do with the other caches too so we don't have to pass them around everywhere. Thoughts @jasonrhansen? I kind of like it!

jasonrhansen:
I like this for `write_str` since it gets called from several different places. The other caches wouldn't really benefit. The only other function we have that takes one is `filled_rectangle`, but it's only called from one place.

jonhoo:
@andy128k I'd be happy to merge something like this. Could you re-apply this change on top of master since I merged #108 before you filed this?

andy128k:
@jonhoo Done

jonhoo:
@andy128k oh, looks like rustfmt [complained](https://travis-ci.com/jonhoo/inferno/jobs/186827251). should be a quick fix.

jonhoo:
:heart:

PR #110, opened 2019-03-22
backport modern css js

### LINKED ISSUE #50: Port  Optimize FlameGraphs by utilizing CSS and modern JS from FlameGraph

AnderEnder:
closes #50


AnderEnder:
Only nameattr functionality and tests are left to merge properly. I am not yet sure about requirements for this case.

jonhoo:
cc @jasonrhansen @jbcden

jasonrhansen:
This is looking a lot like the partial work I did on this. As for what needs to happen for nameattr functionality, I think this is what needs to change:
- Remove everything in `attrs.rs` related to `onmouseover`, `onmouseout`, and `onclick`.
- Add `pub(super) id: Option<String>` field to `GElementAttrs` so the `id` can be overridden using `nameattr`.
- Add `id: Option<&'a str>` field to `FrameAttributes` and update `override_or_add_attributes` to handle it exactly like `class`.
- When writing the optional attributes for the `g` element (on line 584 of flamegraph/mod.rs) write the 'id' like the 'class':
```Rust
        if let Some(id) = frame_attributes.id {
            c.extend_attributes(std::iter::once(("id", id)));
        }

```
-  Remove `onmouseover`, `onmouseout`, and `onclick` from the test nameattr files.

src/flamegraph/mod.rs
```diff
@@ -621,8 +577,35 @@ where
...
+                    g.extend_attributes(std::iter::once(("style", style)));
+                }
+                if let Some(extra) = frame_attributes.g_extra {
+                    g.extend_attributes(extra.iter().map(|(k, v)| (k.as_str(), v.as_str())));
+                }
+            } else {
+                unreachable!("cache wrapper was of wrong type: {:?}", cache_g);
+            }
+
+            svg.write_event(&cache_g)?;
```
jasonrhansen: All of the attributes that get added to the `<g>` element here need to also be added to the `<a>` for the `href` case. Now that `<a>` can take the place of the `<g>` as the container, it needs to have everything the `<g>` would have plus all of the stuff that only belongs on the `<a>`.

It might be helpful to pull this common "container" writing code out into a separate function so it can be reused in both cases.

```Rust
        // Use either <a> or <g> element for the container.
        if let Some(href) = frame_attributes.href {
            write_href_attributes(&mut cache_a, &frame_attributes, href);
            write_container_attributes(&mut cache_a, &frame_attributes, false);
            svg.write_event(&cache_a)?;
        } else {
            write_container_attributes(&mut cache_g, &frame_attributes, true);
            svg.write_event(&cache_g)?;
        }
```

```Rust
fn write_href_attributes<'a>(event: &mut Event, frame_attributes: &FrameAttributes, href: &str) {
    if let Event::Start(ref mut a) = event {
        a.clear_attributes();

        a.extend_attributes(args!(
            "xlink:href" => href,
            "target" => frame_attributes.target
        ));
        if let Some(extra) = frame_attributes.a_extra {
            a.extend_attributes(extra.iter().map(|(k, v)| (k.as_str(), v.as_str())));
        }
    } else {
        unreachable!("cache wrapper was of wrong type: {:?}", event);
    }
}

fn write_container_attributes<'a>(
    event: &mut Event,
    frame_attributes: &FrameAttributes,
    clear_attributes: bool,
) {
    if let Event::Start(ref mut c) = event {
        if clear_attributes {
            c.clear_attributes();
        }

        if let Some(id) = frame_attributes.id {
            c.extend_attributes(std::iter::once(("id", id)));
        }
        if let Some(class) = frame_attributes.class {
            c.extend_attributes(std::iter::once(("class", class)));
        }
        if let Some(extra) = frame_attributes.g_extra {
            c.extend_attributes(extra.iter().map(|(k, v)| (k.as_str(), v.as_str())));
        }
    } else {
        unreachable!("cache wrapper was of wrong type: {:?}", event);
    }
}
```

jasonrhansen:
Here's another thing to add to the list of required changes for nameattr:

Update `FuncFrameAttrMap::from_reader` to load the `id` in the `match`.
```
"id" => funcattrs.g.id = Some(value.to_string()),
```

AnderEnder:
Should `style` be removed as well? I see it is absent in your example.

jasonrhansen:
Yes, it does look like `style` should be removed. The Perl version only uses `id` and `class` there.
```Perl
my @g_attr = map {
    exists $attr->{$_} ? sprintf(qq/$_="%s"/, $attr->{$_}) : ()
} qw(id class);
```

jasonrhansen:
You'll want to remove `style` everywhere in `attrs.rs` and the test nameattr files as well.

AnderEnder:
Could you please manually check any complicated graphs to verify JS functionality on this branch? I tested only one simple case.

jasonrhansen:
This is looking really good to me. I'll check out your branch and do some testing.

jasonrhansen:
I did some manual checking using `example-perf-stacks-collapsed.txt` as input and also using the test nameattr file. It looks to be working correctly to me.

Also, I benchmarked the changes and as I'd hoped the performance improved. So now the generated SVG is more optimized, and the time to generate it has gone down!
```
flamegraph/dtrace       time:   [1.0352 ms 1.0460 ms 1.0624 ms]
                        thrpt:  [71.507 MiB/s 72.631 MiB/s 73.390 MiB/s]
                 change:
                        time:   [-10.155% -9.3227% -8.4092%] (p = 0.00 < 0.05)
                        thrpt:  [+9.1813% +10.281% +11.303%]
                        Performance has improved.

flamegraph/perf         time:   [4.0804 ms 4.0912 ms 4.1041 ms]
                        thrpt:  [150.22 MiB/s 150.69 MiB/s 151.09 MiB/s]
                 change:
                        time:   [-8.5026% -8.0967% -7.7188%] (p = 0.00 < 0.05)
                        thrpt:  [+8.3644% +8.8100% +9.2927%]
                        Performance has improved.
```
@jonhoo This looks ready to merge to me. Do you want to look it over first?

jasonrhansen:
@AnderEnder Actually there's one more thing I can think of. We're currently not testing the `id` attribute in a nameattr file. Do you want to update `tests/data/flamegraph/nameattr/nameattr.txt` to use an `id` on `main` then update the expected svg file?

src/flamegraph/svg.rs
```diff
@@ -114,9 +111,18 @@ where
...
-        ".func_g:hover { stroke:black; stroke-width:0.5; cursor:pointer; }",
-    )))?;
+
+    let titlesize = &opt.font_size + 5;
+    svg.write_event(Event::Text(BytesText::from_escaped_str(&format!(
+        "
+text {{ font-family:{}; font-size:{}px; fill:rgb(0,0,0); }}
+#title {{ text-anchor:middle; font-size:{}px; }}
+{}",
+        &opt.font_type,
```
jonhoo: This probably needs to be quoted and escaped.

src/flamegraph/mod.rs
```diff
@@ -679,30 +630,57 @@ where
...
         if href_is_some {
-            svg.write_event(Event::End(BytesEnd::borrowed(b"a")))?;
+            svg.write_event(&cache_a_end)?;
+        } else {
+            svg.write_event(&cache_g_end)?;
         }
-        svg.write_event(Event::End(BytesEnd::borrowed(b"g")))?;
     }
 
+    svg.write_event(&cache_g_end)?;
```
jonhoo: Doesn't this end `g` twice (we also end it in the `else` just above)?

jonhoo: Ah, never mind, this ends the `g#frames`.

jonhoo:
LGTM subject to that one comment I left about `font_type` :+1:

src/flamegraph/svg.rs
AnderEnder: nice catch

AnderEnder:
@jasonrhansen,
I have just realised that `id` should be unique and can not be assigned to multiple entities in xml.

jasonrhansen:
@AnderEnder I see what you mean, but you can still use it to assign an `id` to a unique entity if the function name is used only in one place. I think we should support it if only because upstream flamegraph does.

@jonhoo, what are your thoughts?

jonhoo:
Hmm, yeah, I'm inclined to follow upstream here, though I _do_ think we should make sure that we don't add _two_ ids to a single elements (that may already be the case?), because that will break all sorts of stuff.

Beyond that, this now looks pretty good to me :)

jasonrhansen:
I think currently with `g_extra` and `a_extra` it's possible to include _any_ attribute multiple times on a single element. You wouldn't want to do that, but nothing currently prevents it. In my opinion fixing this is outside of the scope of this PR.

I suggest we just add `id` to the test nameattr file, then create a new issue for the duplicate attribute problem.

jonhoo:
I'm going to go ahead and merge this once CI passes. @AnderEnder a PR for more elaborate attribute management would be welcome, but I don't consider it necessary to land this :) Thanks again!

PR #111, opened 2019-03-24
Add CLI tests using assert_cmd

jasonrhansen:
Test calling the various CLI programs, including the use of STDIN.

Some of these were originally added in #84, but had to be removed because they were causing problems with CI because of the way the commands were using `cargo run`. Using `assert_cmd` instead should fix this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ pretty_assertions = "0.6"
 testing_logger = "0.1.1"
 libflate = "0.1"
 criterion = "0.2"
+assert_cmd = "0.11"
 
 [profile.release]
 debug = true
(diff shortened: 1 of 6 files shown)


jonhoo:
I actually don't know whether or not this will still trip up the code coverage tester, since it might still have to re-compile a bunch of things _while_ measuring coverage. I have no idea. It _might_ be that we have to ignore those tests for coverage.

jonhoo:
Huh, apparently not! That's cool. It's probably because it [re-uses the cargo test binaries](https://docs.rs/assert_cmd/0.11.1/assert_cmd/cargo/index.html). Sadly it doesn't seem to counted in coverage, but I think it's a good addition nonetheless!

PR #113, opened 2019-03-26
Don't allow duplicate attributes with nameattr

jasonrhansen:
With a nameattr line like the following, you could include the same attribute multiple times. In this case the `id` is included 3 times.
```
noploop	id=test	id=test2	g_extra=id="test3"
```
This PR makes sure the attribute only gets added once. It'll only use the last value, which is "test3" in this case, and log a warning that it is being overwritten. Before, this would just create an invalid element with multiple `id`s and the SVG wouldn't load.

Also, #110 made a change where if an `href` is defined for a function name, the `<a>` tag takes the place of the `<g>` tag as the container. Because of this, it no longer makes sense to keep track of `a_extra` and `g_extra`. All attributes are now combined into one map.


src/flamegraph/attrs.rs
```diff
@@ -83,44 +88,35 @@ impl FuncFrameAttrsMap {
...
-
-    /// Extra attributes to include
-    pub(super) extra: Vec<(String, String)>,
-}
+impl FrameAttrs {
+    fn add_attr(&mut self, func: &str, name: String, value: String) {
+        match self.attrs.entry(name) {
+            Entry::Occupied(mut e) => {
+                warn!(
+                    "duplicate attribute `{}` in nameattr file for `{}`; replacing value \"{}\" with \"{}\"",
```
jonhoo: So we'll get this warning if any of the standard attributes are overwritten then I take it?

src/flamegraph/attrs.rs
```diff
@@ -60,13 +62,16 @@ impl FuncFrameAttrsMap {
...
-                    "g_extra" => parse_extra_attrs(&mut funcattrs.g.extra, value),
-                    "a_extra" => parse_extra_attrs(&mut funcattrs.a.extra, value),
+                    "title" => {
+                        funcattrs.title = Some(value.to_string());
+                    }
+                    "href" => {
+                        funcattrs.add_attr(func, "xlink:href".to_string(), value.to_string());
+                    }
+                    "id" | "class" | "target" => {
+                        funcattrs.add_attr(func, name.to_string(), value.to_string());
```
jonhoo: Double the number of `String`s... I guess it doesn't really matter since this isn't in performance-sensitive code.

src/flamegraph/attrs.rs
```diff
@@ -1,3 +1,5 @@
+use indexmap::map::Entry;
+use indexmap::IndexMap;
```
jonhoo: Is there a particular reason we need `indexmap` for this?

src/flamegraph/attrs.rs
```diff
@@ -205,51 +202,36 @@ mod test {
...
-                    extra: foo_a_extra,
-                },
+                attrs: foo_attrs,
             },
         );
 
-        let bar_a_extra: Vec<(String, String)> = vec![
-            ("aextra1".to_owned(), "foo".to_owned()),
-            ("aextra2".to_owned(), "bar".to_owned()),
+        let bar_attrs: IndexMap<String, String> = indexmap![
```
jonhoo: If you just want the macro, you can just use the [`maplit` crate](https://docs.rs/maplit/).

src/flamegraph/mod.rs
```diff
@@ -557,21 +503,28 @@ where
...
+                has_href = true;
+            } else {
+                write_container_attributes(&mut cache_g, &frame_attributes);
+                svg.write_event(&cache_g)?;
+            }
+            if let Some(ref t) = frame_attributes.title {
+                title = t.as_str();
+            }
+        } else if let Event::Start(ref mut c) = cache_g {
+            c.clear_attributes();
```
jonhoo: Why does this need to clear the cache when the other arms above don't?

src/flamegraph/attrs.rs
jonhoo: Ah, it's for the tests... That's annoying. I wonder if we should just use a non-random map instead? It'd be faster too. Like an `FnvHashMap` without `RandomState`.

src/flamegraph/attrs.rs
```diff
@@ -83,44 +88,35 @@ impl FuncFrameAttrsMap {
 
 /// Attributes to set on the SVG elements of a frame
 #[derive(PartialEq, Eq, Debug, Default)]
-pub(super) struct FrameAttrs {
+pub struct FrameAttrs {
```
jonhoo: Why does this need to be `pub`? Can't it be `pub(crate)`?

src/flamegraph/mod.rs
```diff
@@ -277,60 +277,6 @@ impl Default for Direction {
...
-fn override_or_add_attributes<'a>(
-    title: &'a str,
-    attributes: Option<&'a attrs::FrameAttrs>,
-) -> FrameAttributes<'a> {
-    let mut title = title;
-    let mut id = None;
-    let mut class = None;
-    let mut g_extra = None;
-    let mut href = None;
-    let mut target = "_top";
```
jonhoo: I think I missed it -- where is this default now kept?

src/flamegraph/attrs.rs
jasonrhansen: The container elements for a function won't have any attributes except those that are added using the nameattr file. This will warn if any of those attributes are duplicated.

src/flamegraph/attrs.rs
jasonrhansen: Yes, for the non-extra attributes it will be double the number of `String`s and I don't think it matters at all for loading nameattr files. It makes everything else simpler to put all of the attributes in a map.

src/flamegraph/attrs.rs
jasonrhansen: Yes, for tests. I'll look into `FnvHashMap`. I've never used it.

src/flamegraph/mod.rs
jasonrhansen: The other arms do in `write_container_attributes`, but I could pull it out and put it above the `if`.

src/flamegraph/mod.rs
jasonrhansen: Yeah, good catch. I noticed this right after I submitted the PR. I'll fix it.

src/flamegraph/mod.rs
jasonrhansen: Actually, I guess it can't go above the `if` because `c` doesn't exist there.

jonhoo: Ohh, I missed that it was done in the function. Seems fine then :+1:

src/flamegraph/attrs.rs
jonhoo: Ah, I think I misread the code a little. I thought this map was going to hold _all_ the attributes, but not realize that there is only really one default attribute now, and that is `target`. I don't know where you're planning on adding that, but let's try to make sure we don't emit this warning when the user tries to set `target`.

src/flamegraph/attrs.rs
jonhoo: Take a look at the leading [documentation for `HashMap`](https://doc.rust-lang.org/std/collections/struct.HashMap.html), and note how it uses a "randomly seeded" hashing algorithm that is resistant to HashDoS. We don't need that here, and are paying for it both in terms of speed and randomized ordering. The [`FnvHashMap`](https://doc.servo.org/fnv/type.FnvHashMap.html) from the [`fnv`](https://doc.servo.org/fnv/) crate does _not_ use a random seed, and uses a weaker but faster hashing function, and should serve this purpose well. It'll also not randomize the ordering of our keys. It's perhaps worth noting that if the hashes of the strings should ever change in the future for some weird reason, then the attribute order will change, but I think that's relatively unimportant.

src/flamegraph/attrs.rs
```diff
@@ -202,15 +209,14 @@ mod test {
...
-            "aextra2".to_owned() => "foo aextra2".to_owned(),
-        };
+        let mut foo_attrs = FnvHashMap::default();
+        foo_attrs.insert("class".to_owned(), "foo class".to_owned());
+        foo_attrs.insert("xlink:href".to_owned(), "foo href".to_owned());
+        foo_attrs.insert("target".to_owned(), "foo target".to_owned());
+        foo_attrs.insert("gextra1".to_owned(), "gextra1".to_owned());
+        foo_attrs.insert("gextra2".to_owned(), "foo gextra2".to_owned());
+        foo_attrs.insert("aextra1".to_owned(), "foo aextra1".to_owned());
+        foo_attrs.insert("aextra2".to_owned(), "foo aextra2".to_owned());
```
jonhoo: I'd probably add a `dev-dependencies` on `maplit` to clean this up :)

src/flamegraph/attrs.rs
```diff
@@ -75,6 +75,13 @@ impl FuncFrameAttrsMap {
                     _ => warn!("invalid attribute {} found for {}", name, func),
                 }
             }
+
+            if funcattrs.attrs.contains_key("xlink:href") && !funcattrs.attrs.contains_key("target")
```
jonhoo: It looks like no tests caught this. We should probably change that.

src/flamegraph/attrs.rs
jasonrhansen: Unfortunately it doesn't work with `FnvHashMap`.

jonhoo: `.into_iter().collect()` :joy:

tests/flamegraph.rs
```diff
@@ -232,6 +232,22 @@ fn flamegraph_nameattr_empty_attribute() {
     test_flamegraph(input_file, expected_result_file, options).unwrap();
 }
 
+#[test]
+fn flamegraph_nameattr_href_no_target() {
```
jonhoo: Oh, I was just thinking we could add it to the existing nameattr test. Do you think it's worthwhile adding this as a separate test?

PR #115, opened 2019-03-26
Add flamegraph benchmarks with reversed stacks

jasonrhansen:


jonhoo:
I wonder if we should just make the reversed stack one the main benchmark and ditch the others?

jasonrhansen:
I wouldn't be against it. The `perf_reverse` one takes around 2 minutes to run on my machine, which is much longer than the others, but I guess that's ok.

jonhoo:
I'm guessing that's `hyperfine` with `--reverse`? That's a good sign for me actually -- it means this is a longer benchmark, which also makes it somewhat more representative. The fast things are sort of fast regardless, and also have a lot more noise.

jasonrhansen:
I'm not sure what this has to do with `hyperfine` since the benchmarks are using `criterion`, but I get what you're saying. The reverse versions are going to be more useful for testing performance, even if it's not the most common way to use flamegraph. I'll remove the ones that don't use reverse.

jonhoo:
Oh, right, sorry, `criterion`. What I meant was that the benchmark gets run many many times :)

benches/flamegraph.rs
```diff
@@ -44,6 +43,8 @@ macro_rules! flamegraph_benchmarks {
 }
 
 flamegraph_benchmarks! {
-    dtrace: "tests/data/collapse-dtrace/results/dtrace-example.txt",
-    perf: "tests/data/collapse-perf/results/example-perf-stacks-collapsed.txt"
+    dtrace_reverse: ("tests/data/collapse-dtrace/results/dtrace-example.txt",
```
jonhoo: I'd just call them `dtrace` and `perf`. Arguably we don't even need both of these, since it doesn't matter for the flamegraph benchmark what the source format was; only the folded stacks matter. I think we keep whichever one is larger. Thoughts?

jasonrhansen: The perf one is quite a bit larger. Let's just keep that one.

jasonrhansen: Do we want to call it something else instead of `perf` since there will only be one.

jonhoo: Yeah, I think we call it just `flamegraph`?

jasonrhansen: Since the group is also called `flamegraph` I thought it would appear in the benchmarks as `flamegraph/flamegraph`, but I just tried it and I was wrong.

PR #118, opened 2019-03-29
Do not allocate memory additionally for tags `title` and closed `text`

AnderEnder:
* cache `title` tags
* use lazy static for closed tag `text`


jonhoo:
Hmm, does `::borrowed` actually allocate? I didn't think it did?

AnderEnder:
While `::borrowed` does not allocate, `Event::Start` and `Event::End` allocate.

jonhoo:
@AnderEnder those are just constructing enum variants -- they can't do any heap allocation?

PR #125, opened 2019-04-23
add flame chart mode

### LINKED ISSUE #23: Flame chart mode
In `flamegraph.pl`, you can pass the [`--flamechart` flag](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L150) to produce a "flame chart" in which stacks frames are _not_ merged. To [quote Brendan](http://www.brendangregg.com/flamegraphs.html):

> Flame charts were first added by Google Chrome's WebKit Web Inspector ([bug](https://bugs.webkit.org/show_bug.cgi?id=111162)). While inspired by flame graphs, flame charts put the passage of time on the x-axis instead of the alphabet. This means that time-based patterns can studied. Flame graphs reorder the x-axis samples alphabetically, which maximizes frame merging, and better shows the big picture of the profile. Multi-threaded applications can't be shown sensibly by a single flame chart, whereas they can with a flame graphs (a problem flame charts didn't need to deal with, since it was initially used for single-threaded JavaScript analysis). Both visualizations are useful, and tools should make both available if possible (eg, TraceCompass does). Some analysis tools have implemented flame charts and mistakingly called them flame graphs.

I'm not _entirely_ sure how time factors into this, since the collapsed stacks don't have any timing information from the original execution in them, but people seem to find them useful, so let's implement them! Also, if you _do_ find them useful, please include docs and examples! If you just want to do the implementation work, that's fine too, just include the quoted explanation above somewhere in the docs.

In terms of code, `--flamechart` mainly [avoids sorting](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L605) the incoming stacks, and [changes the plot title](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L198).

AnderEnder:
Closes #23


jonhoo:
I haven't forgotten about this, but would like to see the response to https://github.com/jonhoo/inferno/issues/23#issuecomment-486225905 first, as I _think_ it means that these charts don't actually turn out useful.

jonhoo:
Wow, sorry, I dropped the ball on this even though the referenced comment has now seen some further discussion! I still wonder if it'd be useful to have a "non-collapsing" folder that _just_ turns stacks into the right line format for the flamegraph tool to consume, but I think this is useful even without that (and such a tool should probably be a separate PR anyway). Let me update my review of this, and hopefully we should be able to land this without too much trouble!

src/bin/flamegraph.rs
```diff
@@ -132,6 +132,14 @@ struct Opt {
     /// JavaScript in all of the test files.
     #[structopt(raw(hidden = "true"), long = "no-javascript")]
     no_javascript: bool,
+
+    /// Produce a flame chart (sort by time, do not merge stacks)
+    #[structopt(
+        long = "flamechart",
+        conflicts_with = "no-sort",
```
jonhoo, line 139: I _think_ this should be `requires = "no-sort"`, unless I'm missing something?

src/flamegraph/mod.rs
```diff
@@ -317,7 +323,7 @@ where
     W: Write,
 {
     let mut reversed = StrStack::new();
-    let (mut frames, time, ignored, delta_max) = if opt.reverse_stack_order {
+    let (mut frames, time, ignored, delta_max) = if opt.reverse_stack_order || opt.flame_chart {
```
jonhoo, line 326: I'm not sure I follow why flame chart mode should use reverse stack order?

src/bin/flamegraph.rs
AnderEnder: I do not remember why :). Right now I think differently.

src/flamegraph/mod.rs
AnderEnder: Looks like I confused reverse stack order with reverse timeline.

src/bin/flamegraph.rs
jonhoo: Oh, I think I see what's going on. Basically, flamechart mode _forcibly_ disables sorting (since it makes no sense with sorting enabled). So, either we should _require_ that the user also give `--no-sort`, or we should _reject_ it if the user passes `--no-sort` (since there's no need to). I'd lean towards the former, but either is fine.

src/bin/flamegraph.rs
```diff
@@ -393,6 +402,7 @@ mod tests {
             bgcolors: Some(color::BackgroundColor::Blue),
             hash: true,
             palette_map: Default::default(),
+            #[cfg(feature = "nameattr")]
```
jonhoo, line 405: This seems weird? Why is this part of this PR?

jonhoo:
@That3Percent want to give this PR a shot and see whether it produces the kind of output you're after?

src/bin/flamegraph.rs
AnderEnder: This attribute `func_frameattrs` is conditional under `nameattr ` feature. Rust analyser shows this place as an issue and I confirm it. Ideally this case should be tested under CI as building and testing under different feature flags, but your pipeline is not customisable.

We can move this change into separate PR. I created 2 additional commits: clippy fix and this change, but supposed they are too small to place into separate PR.

Off-topic: do you want to move from azure pipelines to github actions?

src/bin/flamegraph.rs
AnderEnder: Indeed. Original code just ignore this option.

src/bin/flamegraph.rs
jonhoo: Ah, that's awkward.. That is not an okay `#[cfg]`, since `Options` and all of its fields are `pub`. Cargo features _must_ be additive, which this feature then isn't. We'll need to add `#[non_exhaustive]` to `Options` then, or remove the `#[cfg]` on that field. That's probably a priority to fix, and should happen in a separate PR.

We can customize the CI pipeline pretty easily. See for example
https://github.com/jonhoo/flurry/blob/f623ac994f20d35861799d7f106178a875f6e661/azure-pipelines.yml#L5 . I don't particularly want to move to GitHub actions, though I don't feel too strongly either way.

As for how to test this in CI, [`cargo-hack`](https://github.com/taiki-e/cargo-hack) would be the way to go.

That3Percent:
I gave this a try and got an error: `Io(Custom { kind: InvalidData, error: "unsorted input lines detected" })`

Using options:
```
let mut options = flamegraph::Options {
        count_name: "ns".to_owned(),
        hash: true,
        flame_chart: true,
        ..Default::default()
    };
```

And the following data, which I hope is correct:
[flames.txt](https://github.com/jonhoo/inferno/files/4795348/flames.txt)

jonhoo:
Ohh, interesting, yeah, that error will definitely need to be suppressed for a flame chart. @AnderEnder it's here:
https://github.com/jonhoo/inferno/blob/0bf9331e37f6d0b56fc861a111d2dedcf33994bb/src/flamegraph/merge.rs#L131-L136

We'll want that check to be disabled if flamechart mode is on.

jonhoo:
@That3Percent The latest commit should at least not fail with that error — care to try again?

AnderEnder:
@jonhoo  and @That3Percent,

I added a condition to `frames`. And result is produced for your example: [flame.svg](https://gist.githubusercontent.com/AnderEnder/1a9fc2c1cd2fee5aa909412108ac78d9/raw/d22c59f607813251f36d43ce84417d4cc26137fa/flame.svg)

That3Percent:
I gave this a try and I'm _really_ happy with the result. This is now much better than the flame graph that `flame` produces because the bars line up with the actual time within the method that an event occurred rather than shifting to fill any gaps. Great! This retains a lot of useful information.

Once this makes it into a release on crates.io I'll include it in `firestorm` and do a proper release of that with docs and the full treatment.

jonhoo:
@That3Percent Excellent, glad to hear it! Should be ready for release in not too long.

@AnderEnder Could you include the provided input file and the produced output file as a test case? If you get stuck, @jasonrhansen may be able to give you some pointers too. Once we have that, I think this is good to land!

jonhoo:
Oh, and an update to the `CHANGELOG` please!

That3Percent:
@AnderEnder Attached is a new input file. there were some minor issues with the last one (when I was removing entries that had 0 time, it was possible to have adjacent stacks with the same name that were actually the same stack). 

[flames.txt](https://github.com/jonhoo/inferno/files/4808383/flames.txt)

src/bin/flamegraph.rs
jonhoo: I'm handling this in https://github.com/jonhoo/inferno/pull/181.

AnderEnder:
@jonhoo, done. Should commit "make func_frameattrs atribute contidional" be removed from this PR?

jonhoo:
Nope, all good, thanks!

jonhoo:
Released in `0.10.0`.

PR #126, opened 2019-05-28
WIP: Add support for collapsing PHP Xdebug traces

daniellockyer:
This PR adds support for collapsing PHP Xdebug traces. It is 10-20x faster than the PHP version bundled in the original flamegraph repo.

Still a lot of code tidying to do, along with tests, but I wanted to get this PR started.


jonhoo:
This is awesome, thank you! I'll give it a read.

Cargo.lock
```diff
@@ -0,0 +1,991 @@
+# This file is automatically @generated by Cargo.
```
jonhoo: `Cargo.lock` should be listed in `.gitignore`, and so shouldn't be included.

src/bin/collapse-xdebug.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use inferno::collapse::xdebug::{Folder, Options};
+use inferno::collapse::Collapse;
+
+#[derive(Debug, StructOpt)]
+#[structopt(
+    name = "inferno-collapse-xdebug",
+    author = "",
+    after_help = "\
+[1] perf script must emit both PID and TIDs for these to work; eg, Linux < 4.1:
```
jonhoo: This help string is probably not accurate

src/bin/collapse-xdebug.rs
```diff
@@ -0,0 +1,59 @@
...
+struct Opt {
+    /// Silence all log output
+    #[structopt(short = "q", long = "quiet")]
+    quiet: bool,
+
+    /// Verbose logging mode (-v, -vv, -vvv)
+    #[structopt(short = "v", long = "verbose", parse(from_occurrences))]
+    verbose: usize,
+
+    /// perf script output file, or STDIN if not specified
```
jonhoo: This should probably say `xdebug`?

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+struct Frames<'a> {
+    calls: &'a CallStack,
+    stack: &'a [usize],
+}
+
+/// Iterate over tab separated fields.
+///
+/// This is essentially the same as `str.split('\t').filter(|slice| slice.len() > 0)` but seemingly
+/// faster. This probably because we know that the character in question is ascii, while the
+/// standard library `StrSearcher` for `char` is not optimized around this case.
```
jonhoo: It'd be great if you could link to a spec somewhere that states that inputs are all known to be ASCII.

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+
+enum Interned<T> {
+    Old(T),
+    New(T),
+}
+
+#[derive(Default)]
+struct CallStack {
+    strings: HashMap<Rc<str>, usize>,
+    interned_string: Vec<Rc<str>>,
```
jonhoo: I'm curious how much of a speedup you see from string interning? If it's significant, we might want to consider doing the same for the other stackcollapsers.

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+
+            if let Ok(n) = input.read_line(&mut line) {
+                if n == 0 {
+                    break;
+                } else {
+                    return Some(false);
+                }
+            }
+
+            if line.starts_with(TRACE_START) {
```
jonhoo: If you know that an xdebug file always starts with `TRACE_START` on the very first line, then this check should probably test specifically for that, and return `Some(false)` otherwise.

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+        let mut input = input.as_bytes();
+        let mut line = String::new();
+        loop {
+            line.clear();
+
+            if let Ok(n) = input.read_line(&mut line) {
+                if n == 0 {
+                    break;
+                } else {
+                    return Some(false);
```
jonhoo: This doesn't seem right? Won't this never return `Some(true)` except if it fails to parse a line?

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+    }
+}
+
+impl<'a> Iterator for Fields<'a> {
+    type Item = &'a str;
+
+    fn next(&mut self) -> Option<&'a str> {
+        let begin = self
+            .line
+            .as_bytes()
```
jonhoo: Any particular reason you don't use `self.line.as_bytes().split(b'\t')`? That should also be fast since it also basically assumes ASCII.

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+                    "1" => true,
+                    "0" => false,
+                    a => panic!(format!("uh oh: {}", a)),
+                };
+
+                let time = time.parse::<f32>().unwrap();
+
+                (is_exit, time)
+            } else {
+                continue;
```
jonhoo: Should probably also warn for non-empty lines that hit this.

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+            };
+
+            if is_exit && current_stack.is_empty() {
+                warn!("Found function exit without corresponding entrance. Discarding line. Check your input.");
```
jonhoo: It might be useful to include a line number in this warning.

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+                continue;
+            }
+
+            {
+                let current = current_stack.current();
+                let duration = SCALE_FACTOR * (time - prev_start_time);
+
+                // hash `current` for lookup and insertion.
```
jonhoo: I think it'd be useful to add a comment here about exactly why you're doing this in such a roundabout way. I _suspect_ it is because you don't want to clone the string if you don't have to do the insertion, which the regular entry API would require you to do?

@jasonrhansen this might be something we want to do for the other collapse implementations too.

src/collapse/xdebug.rs
jonhoo: /cc @jasonrhansen

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+                current_stack.pop();
+            } else {
+                let _ = parts.next();
+                let func_name = parts.next();
+                let _ = parts.next();
+                let path_name = parts.next();
+
+                if let (Some(func_name), Some(path_name)) = (func_name, path_name) {
+                    current_stack.call(func_name, path_name);
+                }
```
jonhoo: Should the `else` here have a `warn!`?

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+            }
+
+            prev_start_time = time;
+        }
+
+        for (key, value) in stacks {
```
jonhoo: Even though it's sad, I think you should sort these like we do in the other collapsers:
https://github.com/jonhoo/inferno/blob/65a2f20246611229f35b8f701c386dbb9531b8e7/src/collapse/perf.rs#L397-L405

We sort because that's what upstream does, and because flamegraph requires its input to be sorted (and re-sorting is faster if it's already sorted).

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+use log::warn;
+use std::fmt;
+use std::hash::{BuildHasher, Hash, Hasher};
+use std::io::prelude::*;
+use std::io::{self, Write};
+use std::rc::Rc;
+// Ideas: str_stack and colosseum to keep allocations closer together or something.
+
+const SCALE_FACTOR: f32 = 1_000_000.0;
+static CALLS: &[&str] = &["require", "require_once", "include", "include_once"];
```
jonhoo: Why are these necessary?

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+use hashbrown::hash_map::RawEntryMut;
+use log::warn;
+use std::fmt;
+use std::hash::{BuildHasher, Hash, Hasher};
+use std::io::prelude::*;
+use std::io::{self, Write};
+use std::rc::Rc;
+// Ideas: str_stack and colosseum to keep allocations closer together or something.
+
+const SCALE_FACTOR: f32 = 1_000_000.0;
```
jonhoo: Should this be configurable?

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+            interned_string: CALLS.iter().cloned().map(Rc::from).collect(),
+            calls: HashMap::new(),
+            interned: Vec::new(),
+            stack: Vec::with_capacity(16),
+        }
+    }
+
+    /// Push a `call` line on top of the stack.
+    pub fn call(&mut self, name: &str, path: &str) {
+        let new_or_not = match self.intern_str(name) {
```
jonhoo: I can't quite wrap my head around this code. A comment explaining what it does and why would be helpful.

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+    pub fn pop(&mut self) {
+        self.stack.pop();
+    }
+
+    /// An empty stack means we finished main.
+    fn is_empty(&self) -> bool {
+        self.stack.is_empty()
+    }
+
+    /// Get a comparable, hashable representation of the call stack. of the call stack.
```
jonhoo: Trailing words.

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+    fn current(&self) -> &[usize] {
+        self.stack.as_slice()
+    }
+
+    /// Intern a string, return the unique index.
+    ///
+    /// `Ok` when the string was already present.
```
jonhoo: There is no `Ok` return here?

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+    fn intern_str(&mut self, string: &str) -> Interned<Str> {
+        let mut hasher = self.strings.hasher().build_hasher();
+        string.hash(&mut hasher);
+        let hash = hasher.finish();
+
+        let entry = self
```
jonhoo: Given that you also do this trick elsewhere, I think it'd be useful to pull it out into a standalone function. Maybe what we really want here is a version of entry that takes a `Cow` now that I think about it...

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+
+        for (key, value) in stacks {
+            writeln!(writer, "{} {}", current_stack.frames(&key), value)?;
+        }
+
+        Ok(())
+    }
+}
+
+impl CallStack {
```
jonhoo: I wonder if we could pull this out and re-use it for the other collapsers... :thinking: 
/cc @jasonrhansen

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+        let element: Rc<str> = Rc::from(string);
+        self.interned_string.push(element.clone());
+        vacant.insert_hashed_nocheck(hash, element, index);
+        Interned::New(Str(index))
+    }
+
+    /// Intern a `Call` into a `usize` index.
+    ///
+    /// Some tests revealed that this was faster than comparing the `Call` instances directly, so
+    /// it was kept. This seems a bit odd but was backed by data at the time.
```
jonhoo: I'm not sure I understand? This function doesn't do any comparison...

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+    fn intern(&mut self, call: Interned<Call>) -> usize {
+        let new = match call {
+            // The strings were not seen before, definitely new.
+            Interned::New(t) => t,
+            // The strings used were seen before, but maybe not in this call. So retest.
```
jonhoo: Not sure I understand this comment

src/collapse/xdebug.rs
```diff
@@ -0,0 +1,367 @@
...
+    fn frames<'a>(&'a self, stack: &'a [usize]) -> Frames<'a> {
+        Frames { calls: self, stack }
+    }
+
+    /// Create a name for the current stack.
+    ///
+    /// This is potentially costly.
+    fn write_name(&self, indices: &[usize], buffer: &mut fmt::Formatter) -> fmt::Result {
+        let mut indices = indices.iter().cloned();
+        if let Some(first) = indices.by_ref().next() {
```
jonhoo: I think I'd just use `.peekable()` on the iterator for this, but :man_shrugging:

jonhoo:
The failing tests should be fixed by https://github.com/jonhoo/inferno/pull/127 by the way :)

src/collapse/xdebug.rs
neosilky: This change was lumped in with some other optimizations sent to me by a contributor, so it's hard to say how much of a benefit it was. Still, might be worth a try.

jonhoo: I have half a mind to merge this without string interning, and then do the interning as a separate PR afterwards. It'd be a little work to separate out, but I think it would be valuable to evaluate it on its own. What do you think?

src/collapse/xdebug.rs
jasonrhansen: One problem is we just switched back to using `HashMap` in `std`, but the raw entry API is nightly-only there. We'd have to switch back to `hashbrown`.

src/collapse/xdebug.rs
jasonrhansen: I think it would make sense to reuse it for the other collapsers, but only if the string interning gives enough of a performance improvement.

src/collapse/xdebug.rs
jonhoo: Ah, good point, I didn't think of that... I think I'd be okay with switching back if we found this to yield a significant speed-up though (and it may well do!).

jonhoo:
@daniellockyer I still think it'd be awesome to try and land this. With the new multicore work from #128, we might even be able to get higher speed-ups by automatically taking advantage of more cores! Just FYI, I'll be offline for a few weeks starting a week from now, but I think my comments from above still apply.

jonhoo:
Ooooh, exciting to see new commits on this! :tada:

daniellockyer:
Unfortunately I don't have the use-case to work on this anymore, so I'm going to close the PR. If there's sufficient desire for the feature, I can look at rebasing everything just for fun 🙂

jonhoo:
Thanks for checking back in — closing seems reasonable. Should someone need this in the future they can always pick up from where you left off.

PR #127, opened 2019-05-28
Updates following hashbrown's inclusion in std

jonhoo:
`hashbrown` no longer needs to be included as a dependency. It also has a different ordering than the old `HashMap`, which caused our nameattr tests to start failing. Instead of relying on the new order (which would make tests fail on stable), I changed the implementation to use an `IndexMap` (which preserves insertion order) so that we aren't dependent on the map order any more. This is probably also more reasonable, since it means that the attribute order from the nameattr file will be preserved.

This PR also switches most maps to use FNV hashing, since we don't really need the adverserial robustness that randomized hashing gives us in this context.

### DIFF
--- a/src/differential/mod.rs
+++ b/src/differential/mod.rs
@@ -1,4 +1,4 @@
-use hashbrown::HashMap;
+use std::collections::HashMap;
 use log::warn;
 use std::fs::File;
 use std::io;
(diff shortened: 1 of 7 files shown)


src/differential/mod.rs
```diff
@@ -1,4 +1,4 @@
-use hashbrown::HashMap;
+use fnv::FnvHashMap as HashMap;
```
jasonrhansen: Is there a reason you're rebinding `FnvHashMap` as `HashMap` in this file but not the others? I'd prefer to make the binding consistent across files unless there's a good reason not to.

PR #129, opened 2019-06-12
Add unit tests for JavaScript semantic coloring

jordins:
Added some unit tests for Javascript semantic coloring.

https://github.com/jonhoo/inferno/issues/63

Note:
I'm very new into Rust and to your project and I would appreciate any feedback in order to improve!
Thanks!

PR #130, opened 2019-06-13
Add unit tests for Perl semantic coloring

jkurian:
Added some unit tests for Perl semantic coloring.

#63

* Used same template as  found in PR #129


jonhoo:
Awesome! Can you re-base this change now that #129 has landed? Should be pretty straightforward :)

jkurian:
Thanks for the quick reply @jonhoo, I merged with upstream master!

PR #132, opened 2019-07-01
Add demangle option to collapsers

### LINKED ISSUE #87: Add back demangling support
Prior to #74, we had support for demangling function names (although only with `--inline`). While perf is _supposed_ to be doing it already (see [`--demangle`](http://man7.org/linux/man-pages/man1/perf-script.1.html)) there are [reports](https://www.reddit.com/r/rust/comments/b08oex/cargoflamegraph_generate_flame_graphs_directly/eidhhz8/) that it doesn't demangle everything super well. Having our own demangling step using one of the [many demangling crates](https://crates.io/search?q=demangle&sort=recent-downloads) seems like it might be a good idea!

jasonrhansen:
Fixes #87 

This is only useful if you pass `--no-demangle` to `perf script`, or `-xmangled` to `dtrace`.

Since DTrace doesn't do a great job demangling Rust names, you can get better results by leaving names mangled in profiles, then passing `--demangle` to `inferno-collapse-dtrace`. It would probably make sense for [ferrous-systems/flamegraph](https://github.com/ferrous-systems/flamegraph) to use DTrace this way.


src/collapse/dtrace.rs
```diff
@@ -182,8 +187,17 @@ impl Folder {
             frame = Self::uncpp(frame);
         }
 
-        if frame.is_empty() {
-            frame = "-";
+        let frame = if frame.is_empty() {
+            Cow::Owned("-".to_owned())
```
jonhoo:
```suggestion
            Cow::Borrowed("-")
```

src/collapse/dtrace.rs
```diff
@@ -182,8 +187,17 @@ impl Folder {
...
+        } else if self.opt.demangle {
+            let mut parts = frame.split('`');
```
jonhoo: Are there only ever two parts to this? Or do we want something like `rsplitn(..., 2)`?

jonhoo:
Looks like `rustfmt` is complaining: https://travis-ci.com/jonhoo/inferno/jobs/212451904

src/collapse/dtrace.rs
jasonrhansen: I believe there is only ever one `` ` ``, but this made me think of something else. If the `includeoffset` option is used this probably won't work in its current state since it will have an offset, such as `+0x21`, at the end. I'll change it to handle this case.

src/collapse/dtrace.rs
```diff
@@ -164,6 +169,21 @@ impl Folder {
         )
     }
 
+    fn demangle<'a>(&self, frame: &'a str) -> Cow<'a, str> {
+        let mut parts = frame.splitn(2, '`');
```
jonhoo: Should this be `rsplitn`?

src/collapse/dtrace.rs
```diff
@@ -164,6 +169,21 @@ impl Folder {
...
+        if let (Some(pname), Some(func)) = (parts.next(), parts.next()) {
+            if self.opt.includeoffset {
+                let mut parts = func.rsplitn(2, '+');
+                if let (Some(offset), Some(func)) = (parts.next(), parts.next()) {
+                    return Cow::Owned(format!("{}`{}+{}", pname, demangle(func), offset));
```
jonhoo: This is a little confusing as this function is itself called `demangle`, so this looks like recursion, even though it is not. I think I'd prefer using `symbolic_demangle::demangle` explicitly.

src/collapse/dtrace.rs
```diff
@@ -164,6 +169,21 @@ impl Folder {
...
+                }
+            }
+            return Cow::Owned(format!("{}`{}", pname, demangle(func)));
```
jonhoo: Same here.

src/collapse/dtrace.rs
jasonrhansen: I could change it to use `rsplitn`, but it's not clear to me why that would be better. If there is only ever one `` ` ``, it could be written either way. Is there a reason you'd prefer to change it?

src/collapse/dtrace.rs
jasonrhansen: Yes, I can see how that looks like recursion.

src/collapse/dtrace.rs
jonhoo: I guess it's a question of where we would expect to see another ``` if there ever was one. But you're probably right -- if we'd see it anywhere, it would be in the function name, not in the module name.

PR #133, opened 2019-07-05
Add collapser for sample on macOS

jasonrhansen:
This diverges a bit in behavior from [stackcollapse-sample.awk](https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-sample.awk) from upstream. For some reason it was implemented in a way that it outputs a bunch of erroneous stacks with 0 samples. This Rust version doesn't do that. However, the resulting flamegraphs should be the same between the two since those lines with 0 samples will be skipped over when generating the flamegraph.

One unfortunate thing about `sample` is it has the same problem demangling Rust names as `DTrace`. At least with DTrace there's a way to disable demangling so we can handle it in our collapser. I didn't add demangling to this collapser because, as far as I can tell, there's no way to disable demangling for `sample`.


src/collapse/perf.rs
```diff
@@ -280,7 +280,10 @@ impl Folder {
         let pc = line.next()?.trim_end();
         let mut line = line.next()?.rsplitn(2, ' ');
         let mut module = line.next()?;
-        // module is always wrapped in (), so remove those
+        // module should always be wrapped in (), so remove those
+        if !module.starts_with('(') || !module.ends_with(')') {
```
jonhoo: This seems like an unrelated change?

src/collapse/sample.rs
```diff
@@ -0,0 +1,226 @@
...
+        } else {
+            error!("Found stack line with only indent characters:\n{}", line);
+        }
+
+        Ok(())
+    }
+
+    fn write_stack<W: Write>(&self, writer: &mut W) -> io::Result<()> {
+        if let Some(func) = self.stack.last() {
+            for symbol in IGNORE_SYMBOLS {
```
jonhoo: How sure are we that everyone will always want to ignore these?

jonhoo:
Depending on how the demangling is by default, it could be that it makes sense for us to just do it again?

src/collapse/perf.rs
jasonrhansen: When using `inferno-collapse-guess` on a `sample` profile it would panic here when called from `is_applicable`. This is to prevent that.

jonhoo: Maybe update the comment to reflect that?

src/collapse/sample.rs
```diff
@@ -0,0 +1,226 @@
...
+                for _ in 0..=prev_depth - depth {
+                    self.stack.pop();
+                }
+            } else if depth > prev_depth + 1 {
+                error!("Skipped indentation level at line:\n{}", line);
+            }
+
+            if let Some((samples, func, module)) = self.line_parts(&line[4 + indent_chars..]) {
+                if let Ok(samples) = samples.parse::<usize>() {
+                    self.current_samples = samples;
```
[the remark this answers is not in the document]
jasonrhansen: This was intentional based on my understanding of how `sample` works. The way I see it, for each sample, whatever function it was in when sampled will get a leaf entry in the call graph. The children counts should always add up to the count of the parent. The function of a parent node could also show up as a leaf node if it was also sampled when it was at the top of the call stack. In the example below, `std::sys::unix::fs::File::open_c` is both a parent node and a leaf node.
```
    +                   !     618 grep_searcher::searcher::Searcher::search_path::h4867309d69f27419  (in rg) + 127  [0x103749bdf]
    +                   !     : 618 std::fs::OpenOptions::_open::h1cd1c180d181e2f5  (in rg) + 33  [0x103a1e481]
    +                   !     :   618 std::sys::unix::fs::File::open::hb90e1c1c787080f0  (in rg) + 198  [0x103a14ad6]
    +                   !     :     617 std::sys::unix::fs::File::open_c::h3660b29577d73852  (in rg) + 338  [0x103a14c72]
    +                   !     :     | 617 std::sys::unix::cvt_r::h0e253dc028f09451  (in rg) + 53  [0x103a1ac55]
    +                   !     :     |   617 __open  (in libsystem_kernel.dylib) + 10  [0x7fff739051ee]
    +                   !     :     1 std::sys::unix::fs::File::open_c::h3660b29577d73852  (in rg) + 372  [0x103a14c94]
```
Am I not understanding this correctly?

src/collapse/sample.rs
jasonrhansen: Not sure at all. I was just following upstream. Maybe we should add an option for this?

jasonrhansen:
For `rustc_demangle` to properly demangle Rust names, the names need to be mangled. Taking the improperly demangled Rust names from `sample` and demangling again doesn't work. At least `DTrace` has an option to leave names mangled, but not `sample`.

jasonrhansen:
I don't know why the `collapse_sample_cli` test times out on Travis CI for Windows. I just tried running the tests on Windows without a problem.

src/collapse/sample.rs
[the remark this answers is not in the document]
jonhoo: Oh, that might be completely correct, I just wanted to point it out as I don't know the format used by `sample`. It might be good to verify this (perhaps from the man page) and then document it in the code above.

jonhoo:
Seems to fail on FreeBSD too? Is it just really slow?

jasonrhansen:
It's not slow at all when I run it on my computer. The whole test suite runs in 4 or 5 seconds.

jonhoo:
Hmm, that's _very_ interesting. It seems to run just fine on my box too. The test _does_ use `assert-cmd`, and I came across https://github.com/assert-rs/assert_cmd/issues/42, which indicates that there's deadlock potential if stdin/stdout is particularly large, which may be the case for this test?

jasonrhansen:
The input is 125k, but the input file we use for dtrace is 1.4M and that one doesn't deadlock.

jonhoo:
I'll try restarting it and see if that helps. If not, maybe try adding
```
-- --test-threads=1 --nocapture
```
and some basic `println!` statements to that test so we can at least see where the test hangs?

jonhoo:
@jasonrhansen hmm, nope, looks like it got stuck on the exact same test again :/

jasonrhansen:
I haven't yet figured out why the tests are timing out on CI, but I came up with a solution to the poorly demangled Rust symbols. Using `rustc_demangle` on these doesn't work, but I took relevant parts of it and wrote a function that fixes these "partially demangled" Rust names. This could be used in the dtrace collapser as well, which would be much nicer than having to run dtrace with `-xmangled` and `inferno-collapse-dtrace` with `--demangle`.

jasonrhansen:
I couldn't figure out how to get that test to work so I just disabled it for now.

src/collapse/util/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+/// For example:
+///     `_$LT$grep_searcher..searcher..glue..ReadByLine$LT$$u27$s$C$$u20$M$C$$u20$R$C$$u20$S$GT$$GT$::run::h30ecedc997ad7e32`
+/// becomes
+///     `<grep_searcher::searcher::glue::ReadByLine<'s, M, R, S>>::run`
+///
+/// Non-Rust symobols, or Rust symbols that are already demangled, will be returned unchanged.
+///
+/// Based on code in https://github.com/alexcrichton/rustc-demangle/blob/master/src/legacy.rs
+#[allow(clippy::cognitive_complexity)]
+pub(crate) fn fix_partially_demangled_rust_symbol(symbol: &str) -> Cow<str> {
```
jonhoo: It'd be good to have some unit tests for this.

jonhoo:
Excellent! Thank you :tada:

PR #134, opened 2019-07-06
Fix Rust symbols in collapse-dtrace and collapse-perf

### LINKED ISSUE #87: Add back demangling support

jasonrhansen:
This fixes partially demangled Rust function names like we did with collapse-sample.

#132 was already meant to fix #87, but it required `dtrace` to be run with `-xmangled` to leave the names mangled. You then had to pass `--demangle` to `inferno-collapse-dtrace`. This PR will give good results for Rust symbols with the default options.


jasonrhansen:
I did the same for `collapse-perf`. When I run `perf` on my machine, the Rust names are mostly demangled correctly, with the only exception I can see being closures. However, the Rust test profiles from upstream flamegraph have symbols that look like the partially demangled symbols from `dtrace` or `sample`. I'm guessing older versions of `perf` didn't demangle Rust as well?

src/collapse/dtrace.rs
```diff
@@ -168,26 +169,39 @@ impl Folder {
...
                 if let (Some(offset), Some(func)) = (parts.next(), parts.next()) {
-                    return Cow::Owned(format!(
-                        "{}`{}+{}",
-                        pname,
-                        symbolic_demangle::demangle(func),
-                        offset
-                    ));
+                    if let Cow::Owned(func) = transform(func.trim_end()) {
+                        return Cow::Owned(format!("{}`{}+{}", pname, func, offset));
+                    }
```
jonhoo: I think this is missing an `else` so we don't double-evaluate `transform` with the `if let` below.

PR #136, opened 2019-07-10
Add support for "fluid" drawing

### LINKED ISSUE #135: Add support for "fluid" drawing
Currently, the width of the flamegraph, and the individual sample boxes, is set in pixels. It'd be really cool if we could instead use percentages, as it would mean that users with a wider monitor would automatically be able to take advantage of their screen size.

jasonrhansen:
Closes #135


jasonrhansen:
It seems to work well in Firefox and Chrome, but I tested in Safari and Edge, and they both had weird behaviors that I need to look into.

jasonrhansen:
OK, I've got it working well in Safari and Edge now.

src/flamegraph/flamegraph.js
```diff
@@ -1,21 +1,63 @@
-"use strict";
-var details, searchbtn, unzoombtn, matchedtxt, svg, searching;
+"use strict"
```
jonhoo: I _think_ this semicolon still needs to be there to not trip up older browsers?

src/flamegraph/flamegraph.js
```diff
@@ -1,21 +1,63 @@
...
+                svg.attributes.viewBox.value = "0 0 " + svg.width.baseVal.value + " " + svg.height.baseVal.value;
+            }
 
+            // Keep consistent padding on left and right of frames container.
+            frames.attributes.width.value = svg.width.baseVal.value - xpad * 2;
+
+            // Text truncation needs to be adjusted for the current width.
+            var el = frames.children;
+            for(var i = 0; i < el.length; i++) {
+                update_text(el[i]);
```
jonhoo: I think we may be able to do _way_ better here by letting CSS take care of the text truncation using [`text-overflow: ellipsis`](https://css-tricks.com/almanac/properties/t/text-overflow/).

src/flamegraph/flamegraph.js
jasonrhansen: Oops, I didn't change that intentionally.

src/flamegraph/flamegraph.js
jasonrhansen: Yeah, I can give that a try. I was just keeping it the way they were handling text.

jonhoo: I think as we're moving to fluid layouts, we'll explicitly _not_ want to manually crop text, because it requires walking all the elements on every resize! Should also make the initial render faster I think.

jasonrhansen: It doesn't look like `text-overflow` works for SVG text elements.

jonhoo: Ah, that's a shame. I almost wonder whether it's worth just using `overflow: hidden` and not add the ellipses everywhere... Maybe with a very gentle fade on the right-hand side to indicate the crop?

jasonrhansen: I can't get that to work either for SVG, unfortunately.

src/flamegraph/flamegraph.js
```diff
@@ -143,6 +185,11 @@ function update_text(e) {
     }
     t.textContent = "";
 }
+function update_search_elements() {
```
jonhoo: I may be missing something, but this doesn't seem to be called?

src/flamegraph/flamegraph.js
```diff
@@ -123,17 +164,18 @@ function g_to_func(e) {
...
+    t.attributes.x.value = format_percent((parseFloat(r.attributes.x.value) + (100 * 3 / framesWidth)));
     // Smaller than this size won't fit anything
     if (w < 2 * fontsize * fontwidth) {
         t.textContent = "";
         return;
     }
     t.textContent = txt;
     // Fit in full text width
-    if (/^ *\$/.test(txt) || t.getSubStringLength(0, txt.length) < w)
+    if (/^ *\$/.test(txt) || t.getComputedTextLength() < w)
```
jonhoo: What's the reason for this change?

src/flamegraph/flamegraph.js
```diff
@@ -158,29 +205,29 @@ function zoom_child(e, x, ratio) {
     if (e.attributes != undefined) {
         if (e.attributes.x != undefined) {
             orig_save(e, "x");
-            e.attributes.x.value = (parseFloat(e.attributes.x.value) - x - xpad) * ratio + xpad;
+            e.attributes.x.value = format_percent((parseFloat(e.attributes.x.value) - x) * ratio);
```
jonhoo: Looks like `xpad` has been removed in a bunch of places. What's the reasoning behind that?

src/flamegraph/mod.rs
```diff
@@ -27,11 +27,17 @@ use std::iter;
...
+use svg::{Dimension, StyleOptions};
 
-const XPAD: usize = 10; // pad lefm and right
+const XPAD: usize = 10; // pad left and right
 const FRAMEPAD: usize = 1; // vertical padding for frames
 
+// If no image width is given, this will be the initial width, but the embedded JavaScript will set
+// the width to 100% when it loads to make the width "fluid". The reason we give an initial width
+// even when the width will be "fluid" is so it looks good in previewers or viewers that don't run
+// the embedded JavaScript.
```
jonhoo: Nice, I like this.

src/flamegraph/mod.rs
```diff
@@ -434,16 +441,21 @@ where
...
 
     // create frames container
-    if let Event::Start(ref mut g) = cache_g {
-        g.extend_attributes(std::iter::once(("id", "frames")));
-    }
-    svg.write_event(&cache_g)?;
+    let container_x = format!("{}", XPAD);
+    let container_width = format!("{}", image_width as usize - XPAD - XPAD);
+    svg.write_event(Event::Start(
+        BytesStart::borrowed_name(b"svg").with_attributes(vec![
```
jonhoo: Don't we already write out `<svg>` in `write_prelude` above? Should this be `<g>` instead?

src/flamegraph/mod.rs
```diff
@@ -711,9 +730,9 @@ fn filled_rectangle<W: Write>(
     color: Color,
     cache_rect: &mut Event<'_>,
 ) -> quick_xml::Result<usize> {
-    let x = write_usize(buffer, rect.x1);
+    let x = write!(buffer, "{:.4}%", rect.x1_pct);
```
jonhoo: I wonder how much this is going to slow things down...

src/flamegraph/flamegraph.js
jasonrhansen: You're right, I moved the body of this into `update_for_width_change` above and forgot to delete it.

tests/data/flamegraph/colors/java-perl.svg
```diff
@@ -1,1442 +0,0 @@
-<?xml version="1.0" standalone="no"?>
```
jonhoo: Why has this entire file been removed?

src/flamegraph/flamegraph.js
jasonrhansen: I was getting errors because sometimes `txt.length` was `0`. We want the whole text here anyway.

src/flamegraph/flamegraph.js
jasonrhansen: I wanted to keep the left and right padding a fixed number of pixels when resizing. Instead of updating the positions of all of the rectangles each time, I only update the position and width of the parent element and the rectangles no longer need to care about the padding.

In order to do this, I had to change the "frames" element from a `g` to a nested `svg`.

src/flamegraph/mod.rs
jasonrhansen: I changed the "frames" container to `<svg>` to make it easier to maintain fixed padding on the left and right when resizing.

src/flamegraph/mod.rs
jasonrhansen: I'll benchmark it to see, but what can we do about it? Integer percentages wouldn't give us enough precision.

tests/data/flamegraph/colors/java-perl.svg
jasonrhansen: There weren't any tests using this or `js-perl.svg`.

src/flamegraph/mod.rs
jonhoo: Is an `<svg>` even allowed to contain additional `<svg>` elements? Why can't it continue to be a `<g>` that you then set the position/width on?

src/flamegraph/flamegraph.js
jonhoo: Ah, I see. `xpad` was really just a global x padding, not a "per element" thing.

src/flamegraph/mod.rs
jonhoo: No, I think it's fine, I'd just be curious to see the numbers.

jasonrhansen: Somehow this branch outperforms master by quite a bit, even though formatting the percentages does slow it down.

**master -> fluid-drawing**
```
flamegraph              time:   [17.145 ms 17.208 ms 17.283 ms]
                        thrpt:  [35.672 MiB/s 35.827 MiB/s 35.959 MiB/s]
                 change:
                        time:   [-29.290% -28.977% -28.586%] (p = 0.00 < 0.05)
                        thrpt:  [+40.028% +40.799% +41.422%]
                        Performance has improved.
```


**format percent -> write_usize**
```
flamegraph              time:   [13.489 ms 13.538 ms 13.592 ms]
                        thrpt:  [45.358 MiB/s 45.540 MiB/s 45.704 MiB/s]
                 change:
                        time:   [-20.828% -20.198% -19.557%] (p = 0.00 < 0.05)
                        thrpt:  [+24.312% +25.310% +26.307%]
                        Performance has improved.
```

src/flamegraph/mod.rs
jasonrhansen: You would think that'd work, but it doesn't. The `x` and `width` attributes can't be used on a `<g>`. Working with SVG is frustrating because many things don't work like you'd think. This could probably be done much more easily in html.

src/flamegraph/flamegraph.js
```diff
@@ -158,29 +205,29 @@ function zoom_child(e, x, ratio) {
...
             if(e.tagName == "text")
-                e.attributes.x.value = find_child(e.parentNode, "rect[x]").attributes.x.value + 3;
+                e.attributes.x.value = format_percent(parseFloat(find_child(e.parentNode, "rect[x]").attributes.x.value) + 3);
```
jonhoo: This should probably not be `+3`?

src/flamegraph/flamegraph.js
```diff
@@ -123,17 +164,18 @@ function g_to_func(e) {
 function update_text(e) {
     var r = find_child(e, "rect");
     var t = find_child(e, "text");
-    var w = parseFloat(r.attributes.width.value) -3;
-    var txt = find_child(e, "title").textContent.replace(/\\([^(]*\\)\$/,"");
-    t.attributes.x.value = parseFloat(r.attributes.x.value) + 3;
+    var framesWidth = frames.width.baseVal.value;
```
jonhoo: What is this `baseVal` business, and why isn't this just `frames.attributes.width.value`?

src/flamegraph/flamegraph.js
```diff
@@ -192,13 +239,13 @@ function zoom(node) {
...
     var width = parseFloat(attr.width.value);
     var xmin = parseFloat(attr.x.value);
-    var xmax = parseFloat(xmin + width);
+    var xmax = xmin + width;
     var ymin = parseFloat(attr.y.value);
-    var ratio = (svg.width.baseVal.value - 2 * xpad) / width;
+    var ratio = 100 / width;
     // XXX: Workaround for JavaScript float issues (fix me)
-    var fudge = 0.0001;
+    var fudge = 0.001;
```
jonhoo: What's the reasoning behind lowering the fudge factor?

src/flamegraph/flamegraph.js
jasonrhansen: Right, the 3 pixels should be converted to a percent here.

src/flamegraph/flamegraph.js
jasonrhansen: `<svg>` elements have `width` and `height` properties that that are objects with the `SVGAnimatedLength` interface. The `baseVal` will give the value before animation. For the top level `<svg>` we use this because we remove the `width` attribute, but in this case `frames` is the inner `<svg>` and it does have a width attribute. I agree that `frames.attributes.width.value` would be better here.

src/flamegraph/flamegraph.js
jasonrhansen: When switching to percentages, the original fudge factor didn't work well anymore. It was comparing numbers like `100.0` and `99.9996`, and not seeing them as the same. This meant a lot of rectangles wouldn't draw when they should, and vice versa.

jonhoo: I guess I almost wonder whether it should be even larger, but if you've found this to work, that's good enough for me.

jasonrhansen: This value has worked well in all of the testing I've done.

bobrik:
While it's a great change, this is really expensive to render for complex flamegraphs.

Compare initial load and render times:

* Fixed width flamegraph:

image

image

* Fluid width flamegraph:

image

image

Zoom is also a bit slower:

* Fixed width flamegraph:

image

image

* Fluid width flamegraph:

image

image

With 40-60% of the time spent in `update_text` I'm wondering if it makes sense given `--truncate-text-right`.

jonhoo:
@bobrik Yeah, there's definitely a cost involved with having the browser do this work. I think there's a decent argument for us just making monospace fonts the default — what do you think?

bobrik:
That sounds reasonable to me. We should probably allow passing multiple font families at the same time too, so that users can have a preference list rather than one choice that might not be available on every platform.

PR #140, opened 2019-07-24
Fix comparison script on macOS

brinsche:
`zcat` seems to behave differently on macOS and fails with `zcat: can't stat:` as described here: https://serverfault.com/questions/570024/zcat-gzcat-works-in-linux-not-on-osx-general-linux-osx-compatibility/570026#570026 

The fix is also from that serverfault answer, I think this should still work on linux.

Signed-off-by: Bastian Rinsche <bastian.rinsche@gmail.com>


jonhoo:
Wow, how silly. Thanks!

PR #143, opened 2019-08-28
[WIP] Add fuzzer tests

Kobzol:
Hi, I wanted to experiment with fuzzing in Rust, so I thought that I would start with this crate, since you have suggested that some fuzz tests would be nice (https://github.com/jonhoo/inferno/issues/63).

I simply added `cargo fuzz` and I created a single fuzz target for `perf`. Since the `Collapse` trait is very simple, I think that creating a fuzz target for each `Collapse` implementation should be enough. I wanted to ask if you think this is fine before I continue though.

What's more important is to define what do you actually consider a bug in your code and what do you consider a "bug" in the input. For example immediately after I started fuzzing `perf`, it crashed, because it fails on an assertion (src/collapse/perf.rs:173) if the input (file) is empty.

Do you consider this a bug in the code? For cmd usage this is probably not that much different from a controlled notification about a bad input with a `Result`, but for a library usage a crash like that on an invalid input is probably not something that you want.


fuzz/fuzz_targets/common.rs
```diff
@@ -0,0 +1,13 @@
+use std::io::{Error, Write};
+
+pub struct Null;
+
+impl Write for Null {
```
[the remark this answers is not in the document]
Kobzol: I was looking for something like this before I implemented it, I saw `sink` in the documentation, but glanced over it because I thought that something like this would surely be named in a different way :-) Thank you.

Kobzol:
@jonhoo Do you want to run fuzzing on CI? Cirrus CI is only used for FreeBSD, the rest is tested on Azure, which uses some predefined templates?

I could add `cargo +nightly install cargo-fuzz && cargo +nightly fuzz run <targets>` to Cirrus, but I'm not sure how to modify the Azure templates :)

jonhoo:
I think we probably only need to run fuzzing on one of the CIs. Probably Azure. What you'll want to do is add another "stage" before the empty line before the line that reads `resources:`
```yaml
 - stage: fuzz
   displayName: fuzzing tests
   jobs:
     - job: fuzz
       displayName: cargo +nightly fuzz
       pool:
         vmImage: ubuntu-16.04
       steps:
        - template: azure/install-rust.yml@templates
          parameters:
            rust: nightly
        - checkout: self
          submodules: recursive
        - script: cargo install cargo-fuzz
          displayName: Install cargo fuzz
        - script: cargo fuzz run <targets>
          displayName: Run fuzzer
```

azure-pipelines.yml
```diff
@@ -3,6 +3,27 @@ stages:
...
+         vmImage: ubuntu-16.04
+       steps:
+         - template: azure/install-rust.yml@templates
+           parameters:
+             rust: nightly
+         - checkout: self
+           submodules: recursive
+         - script: cargo install cargo-fuzz
+           displayName: Install cargo fuzz
+         - script: |
```
jonhoo: I would probably just make these separate `- script` steps with their own `displayName` :)

Kobzol:
Is Azure currently integrated into this GitHub (like for example in `tokio`) or do you only see the CI results privately? We have to choose some timeout for the fuzzing and maybe some conditions for the fuzzing stage (for example only for merge commits).

azure-pipelines.yml
Kobzol: Good idea :+1:

jonhoo:
Looks like it was misconfigured in some way — should be fixed now! You may want to merge with master.

Kobzol:
Ok, I rebased onto master. However there is a problem with `clippy` that will possibly require an update of `crossbeam-channel`.

jonhoo:
Ah, yes, that seems to be https://github.com/crossbeam-rs/crossbeam/issues/404.
I've pushed changes to `master` that should make things work again. Try merging again?

jonhoo:
Looks like coverage sometimes crashes due to https://github.com/xd009642/tarpaulin/issues/190. Since your fuzzing stage now comes _after_ coverage testing, it might not even run if that triggers. Try adding the following to the new stage you added:
```yaml
   dependsOn: test
```

jonhoo:
Now we're getting somewhere! At least one fuzz failure:
https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=179&view=logs&s=921c3b4a-2264-5a5b-3600-c937c6937e79&j=4181b88e-0661-5860-37d0-7e60733b20ba

Kobzol:
Cool :) Do you have access to the artifact directory? The test was written to `/home/vsts/work/1/s/fuzz/artifacts/perf/crash-da39a3ee5e6b4b0d3255bfef95601890afd80709`, but it probably has to be picked up manually. Maybe it would be possible to recreate the same input locally by running fuzzer with the same seed, but I didn't find such option in the documentation of `cargo fuzz` nor `libfuzzer` (it expects the generated corpus, but that is also on the Azure filesystem).

jonhoo:
Hmm, I think for that you need to add another step at the end that publishes a [Pipeline Artifact](https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml):
```yaml
- publish: $(System.DefaultWorkingDirectory)/fuzz/artifacts/
  condition: failed()
  artifact: fuzz
```
I _think_ that step should still execute even if the previous once failed (due to the `condition`), but not _entirely_ sure. I also don't know where to download that artifact _from_ once it's been published, though the docs seem to hint at it being possible.

Kobzol:
Uhh, the terminology here is atrocious, multiple types of artifacts.. in GitLab it's pretty easy, in Azure I'm a bit lost :-) I think that you might be able to click on the pipeline run and then it shows `Artifacts: -`, maybe it will become clickable if there are any build artifacts exported.

I half expect that the `failed()` condition will only fire when the immediate predecessor step fails (let's hope not :-) ). Let's see.

jonhoo:
Huh, that seems to not even have triggered a build...? :/

Issue #139, opened 2019-07-24
Reduce the number of dependencies


jonhoo:
```console
$ cargo tree --no-indent | sed 's/ (\*)//' | sort -u | wc -l
126
```

My guess is that some of these are unnecessary. Let's see if we can't do something to help improve build times.

felippemr:
Hey @jonhoo I tried building the project locally without each one of the non dev deps but the project would only build with all of them. What is the path you were thinking for this?

felippemr:
When I try to `cargo build`  the project locally I get:
``` 
 --> src/collapse/perf.rs:4:5
  |
4 | use symbolic_demangle::demangle;
  |     ^^^^^^^^^^^^^^^^^ use of undeclared type or module `symbolic_demangle`

error[E0433]: failed to resolve: use of undeclared type or module `symbolic_demangle`
   --> src/collapse/dtrace.rs:332:45
    |
332 |         self.transform_function_name(frame, symbolic_demangle::demangle)
    |                                             ^^^^^^^^^^^^^^^^^ use of undeclared type or module `symbolic_demangle`

error: aborting due to 2 previous errors
```

jonhoo:
Ah, sorry, to be clear, it's not as though we currently list unused dependencies, but rather I would like to see us re-write parts of the code so that we *can* remove dependencies. I'm not sure what the issue you are pointing to with `symbolic_demangle` is -- it compiles fine for me locally?

jasonrhansen:
I think we should be able to remove `symbolic_demangle` as a dependency.

#133 and #134 added to our collapsers the ability to fix partially demangled Rust symbols with no external dependencies. I originally implemented #132 to work around the same problem, but it required you to pass `--no-demangle` to perf script, or `-xmangled` to dtrace since `symbolic_demangle` doesn't work when the symbols are partially demangled. Now that we are able to take the mostly demangled symbols the profiling tools give us and fix them up from there, the `demangle` option on the collapsers seems to be unnecessary.

What do you think?

jasonrhansen:
Now we're down to 109 dependencies after #144.

```Console
$ cargo tree --no-indent | sed 's/ (\*)//' | sort -u | wc -l
109
```

jasonrhansen:
#145, #146, and #147 have reduced dependencies further and made more of them optional.
```Console
$ cargo tree --no-indent | sed 's/ (\*)//' | sort -u | wc -l
107
[...]

PR #144, opened 2019-09-05
Remove demangle option from collapsers

jasonrhansen:
I think we should be able to remove `symbolic_demangle` as a dependency to help with #139. Doing so would also remove the following dependencies.
```
backtrace v0.3.37
backtrace-sys v0.1.31
cc v1.0.41
cpp_demangle v0.2.13
debugid v0.5.3
failure v0.1.5
failure_derive v0.1.5
glob v0.3.0
memmap v0.7.0
msvc-demangler v0.7.0
rustc-demangle v0.1.16
symbolic-common v6.1.4
symbolic-demangle v6.1.4
synstructure v0.10.2
uuid v0.7.4
```

#133 and #134 added to our collapsers the ability to fix partially demangled Rust symbols with no external dependencies. I originally implemented #132 to work around the same problem, but it required you to pass `--no-demangle` to perf script, or `-xmangled` to dtrace since `symbolic_demangle` doesn't work when the symbols are partially demangled. Now that we are able to take the mostly demangled symbols the profiling tools give us and fix them up from there, the `demangle` option on the collapsers seems to be unnecessary.

Do you think it makes sense to remove?


jonhoo:
So, my only concern with this is if using `inferno` to produce flamegraphs for non-Rust code. The code we wrote is specific to Rust code (I believe)?

jasonrhansen:
I believe `dtrace`, `perf`, and `sample` do a good job demangling C++ symbols. I don't know about Swift symbols, which is the other language `symbolic_demangle` can handle.

From what I can tell, Rust symbols are being handled by these tools as if they were C++ when demangling, which gets them close, but we have to fix them up from there.

jonhoo:
Oh, I see what you mean -- we're still relying on the underlying tools to do _most_ of the demangling, and then just fixing it up ourselves. Yeah, I think I'm good with that then!

jasonrhansen:
I'm not able to see the details on Azure Pipelines to see what  failed. If I click on the "View more details" link it says "Build not found." Any suggestions?

jonhoo:
Hmm, I think you're running into https://github.com/xd009642/tarpaulin/issues/190, which I've also seen in the past. You can see the build pipeline here: https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=201&view=results. I'll try a re-run, which may invalidate that link. You _should_ always be able to see builds through https://dev.azure.com/jonhoo/jonhoo/_build?definitionId=10&_a=summary

jasonrhansen:
Strange, now I'm able to get to everything just fine. Earlier I couldn't see any of the builds and I tried in multiple browsers. It's too bad tarpaulin keeps segfaulting.

jonhoo:
Released in `0.9.0` :rocket:

PR #145, opened 2019-09-07
update structopts to 0.3

andy128k:


jonhoo:
Hmm, this seems to bump the minimum supported Rust version. That's probably fine, but it'd be good if you could find out what it is now updated to, and then set the parameter `minrust` to that version in `azure-pipelines.yml`.

andy128k:
@jonhoo It looks like parameter `minrust` is set outside of this repo, so, I cannot change it. You need to change it yourself. According to changelog of `structopts`, minimal supported version is 1.36.

jonhoo:
@andy128k I'm not sure I follow? You should be able to just add it below https://github.com/jonhoo/inferno/blob/5603e4a6ba8db60a30902687e5deedab19d75d6c/azure-pipelines.yml#L3-L6

with:
```yaml
minrust: 1.36.0
```

jonhoo:
Looks great, thank you!

jonhoo:
Released in `0.9.0` :rocket:

PR #147, opened 2019-09-10
Make nameattr an optional feature

jasonrhansen:
This makes `indexmap` an optional dependency.


jonhoo:
Released in `0.9.0` :rocket:

Issue #142, opened 2019-08-26
sport vtune


Eh2406:
Upstream has code for working from a vtune trace. (Yay windows support.)
https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-vtune.pl

Can we add functionality to work with vtune?

PR #148, opened 2019-09-12
Add collapser for VTune

jasonrhansen:
Implements #142


src/collapse/vtune.rs
```diff
@@ -1,10 +1,19 @@
 use std::io::{self, BufRead};
 
-use log::{error, warn};
+use log::warn;
 
 use crate::collapse::common::Occurrences;
 use crate::collapse::Collapse;
 
+macro_rules! invalid_data {
```
jonhoo: We can probably share this macro across many collapsers.

tests/common/collapse.rs
```diff
@@ -104,3 +104,27 @@ where
     collapser.collapse(r, std::io::sink()).unwrap();
     test_logger::validate(asserter);
 }
+
+pub fn test_collapse_error<C>(
```
jonhoo: Is this useful for other collapsers too?

src/collapse/common.rs
```diff
@@ -11,6 +11,15 @@ use crossbeam::channel;
 use fnv::FnvHashMap;
 use lazy_static::lazy_static;
 
+macro_rules! invalid_data_error {
```
jonhoo: Are there other collapsers that could use this right now?

tests/common/collapse.rs
```diff
@@ -104,3 +104,24 @@ where
     collapser.collapse(r, std::io::sink()).unwrap();
     test_logger::validate(asserter);
 }
+
+pub fn test_collapse_error<C>(mut collapser: C, test_filename: &str) -> io::Error
```
jonhoo: Are there other collapsers that could use this right now?

src/collapse/common.rs
jasonrhansen: Yes, I'll do a separate PR for changes to the sample collapser that will use this.

tests/common/collapse.rs
jasonrhansen: Same here, I'll use it for the sample collapser.

PR #150, opened 2019-09-13
Switch back to `testing_logger`

jasonrhansen:
If we run just the log tests single-threaded, we should be able to run
our tests without `--test-threads=1`.


jonhoo:
Hmm, does that really work? I thought it'd be an issue across tests too?

jasonrhansen:
It shouldn't be an issue across tests because the log events are captured in a thread_local variable. It's only a problem when running both the tests and the collapsers multithreaded. So as long as we limit the tests that use `testing_longer` to a single thread we should be good.

Issue #17, opened 2019-02-02
Add support for timemax option


jonhoo:
`flamegraph.pl` has a [`--total` option](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L173) which sets the internal variable [`timemax`](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L113). I think it basically ["compresses"](https://github.com/brendangregg/FlameGraph/blob/f857ebc94bfe2a9bfdc4f1536ebacfb7466f69ba/flamegraph.pl#L1079) the flamegraph, but not sure exactly what effect that has. If someone cleverer than me can figure out what it is for, document it, and then implement it, that'd be awesome!

PR #154, opened 2019-12-01
add --total option. See https://github.com/brendangregg/FlameGraph/pu…

smitty3268:
add --total option for generating the flame graph. see https://github.com/brendangregg/FlameGraph/pull/14 and https://github.com/jonhoo/inferno/issues/17.

The use case for this option is comparing multiple perf runs directly against one another. Specifying the # of samples to use for the total width means the individual rectangles in each graph can be directly compared to each other rather than being resized differently if one run finished more quickly than another.

### DIFF
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -12,6 +12,7 @@ mod merge;
 mod rand;
 mod svg;
 
+use std::cmp;
 use std::fs::File;
 use std::io::prelude::*;
 use std::io::{self, BufReader};
@@ -142,6 +143,11 @@ pub struct Options<'a> {
     /// Defaults to None.
     pub subtitle: Option<String>,
 
+    /// # of samples to size the flame graph width to
+    ///
+    /// Defaults to None, which means it will be determined by the input being charted.
+    pub total_samples: Option<usize>,
+
     /// Width of the flame graph
     ///
     /// Defaults to None, which means the width will be "fluid".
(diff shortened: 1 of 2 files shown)


jonhoo:
This looks great, though it appears that it doesn't fully compile?

src/flamegraph/mod.rs
```diff
@@ -142,6 +143,11 @@ pub struct Options<'a> {
     /// Defaults to None.
     pub subtitle: Option<String>,
 
+    /// # of samples to size the flame graph width to
+    ///
+    /// Defaults to None, which means it will be determined by the input being charted.
```
jonhoo: I think it'd be useful to document here why you might want to use this option as well.

jonhoo:
Are there tests in the original PR too? If there are, would you mind porting them over? Otherwise, just copy an existing basic `flamegraph` test, run it with `total`, and commit the result.

smitty3268:
I'll take a look at adding a test, but it may be a few days before I have a chance.

This is my first pull request on github, does everything here look ok? In particular, do I need to clean up the commits in my fork, or is ok to just have them all there separately like they are now?

jonhoo:
No rush — take whatever time you need :)

That's exciting! Yes, it looks all good. You _can_ tidy up your commit history if you wish, but otherwise I'll just squash them all into one commit before merging. It's probably not worth doing a lot of commit tweaking for a change this small, and individual commits are easier to review anyway!

PR #157, opened 2020-01-25
#156: truncate text on the left by default, but make it configurable

### LINKED ISSUE #156: Cut off left part of text, rather than right part of text when SVG is too small
Let's say your function is too wide to fit on the screen. Right now, the right side will be truncated:

```
for (var x = txt.length - 2; x > 0; x--) {
        if (t.getSubStringLength(0, x + 2) <= w) {
            t.textContent = txt.substring(0, x) + "..";
            return;
        }
    }
    t.textContent = "";
```

So e.g. in my case I have `/home/itamarst/Devel/memory-profiler/venv/lib64/python3.8/site-packages/numpy/core/numeric.py:ones` truncated to `/home/itamarst/Devel/memory-profiler/venv/lib64/..`.

But, you'll notice the most significant information is on the right side of that string: the function name, the main file, etc. It would be better to truncate it in the opposite direction: `../python3.8/site-packages/numpy/core/numeric.py:ones`, or at least have that as a rendering option.

itamarst:
Fixes #156.

I tested the flag manually, but suggestions for additional automated tests are welcome.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 # Ensure git doesn't change the line-endings of our data files on Windows.
-tests/data/**/* binary
+tests/data/**/* binary diff
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 
 use env_logger::Env;
 use inferno::flamegraph::color::{BackgroundColor, PaletteMap, SearchColor};
-use inferno::flamegraph::{self, defaults, Direction, Options, Palette};
+use inferno::flamegraph::{self, defaults, Direction, Options, Palette, TextTruncateDirection};
 
 #[cfg(feature = "nameattr")]
 use inferno::flamegraph::FuncFrameAttrsMap;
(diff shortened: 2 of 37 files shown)


jonhoo:
This looks good to me! Could you add a small test case for this too?

itamarst:
One of the "read this file in, get this file out" kinda tests?

itamarst:
That is, should it be the style of test in tests/flamegraph.rs, or something else? If the latter, what would it assert?

jonhoo:
Yup, a read-in check-out test seems great for this!

itamarst:
So I actually have to go and regenerate all the flamegraph SVGs in the integration tests, since they no longer match checked-in output... is there a utility to do that?

jonhoo:
Ah, because you changed the default I suppose. I don't think so, but @jasonrhansen might have something handy? An alternative is to make the test suite continue to run with cropping on the right, and then just add one test that does cropping on the left.

jasonrhansen:
If you delete the flamegraph SVGs, the tests will recreate them for you. You could do this, then `git diff` and make sure the new SVGs match the originals except for the truncation.

itamarst:
Ok, I have:

1. Updated existing tests.
2. Added a new test.

jonhoo:
Looks great to me, thanks!
Oh, could you please also submit a PR that updates `CHANGELOG.md` to point to these changes under "Unreleased"?

PR #161, opened 2020-02-02
Adjust height based on subtitle.

### LINKED ISSUE #159: Top-heavy flamegraphs hide the subtitle
If there's too long of a line at the top of the graph, the subtitle is hidden, probably because it ends up being covered up

For example:

1. If the top line is very long: `echo "hello 1" | inferno-flamegraph --subtitle "Hello! this is a subtitle" > regular.svg`

2. If the top line is very long because it's an inverted graph: `cat flamegraph/test/results/perf-java-stacks-01-collapsed-pid.txt | inferno-flamegraph --subtitle "Hello! this is a subtitle"  --inverted > inverted.svg`

itamarst:
Fixes #159.

To test, rerun the two commands in #159 and notice if subtitle is visible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Fix bug where subtitles would often be hidden ([#161](https://github.com/jonhoo/inferno/pull/161) by [@itamarst](https://github.com/itamarst))
+
 ### Removed
 
 ## [0.9.3] - 2020-02-02
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -235,9 +235,14 @@ pub struct Options<'a> {
 }
 
 impl<'a> Options<'a> {
-    /// Calculate pad top, including title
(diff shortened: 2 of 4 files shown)


jonhoo:
Interesting.. Do you have a reference to the corresponding code in the flamegraph codebase? I just want to make sure that we do the same thing they do.

itamarst:
I find the automated tools claim that coverage has gone down rather implausible, but if this is important I can find something else to test :grinning:

itamarst:
As I mentioned, the approach taken by the Perl version is wrong in one case. However, they do have some code in there:

```perl
my $ypad1 = $fontsize * 3;      # pad top, include title
my $ypad2 = $fontsize * 2 + 10; # pad bottom, include labels
my $ypad3 = $fontsize * 2;      # pad top, include subtitle (optional)

...
$imageheight += $ypad3 if $subtitletext ne "";
```

jonhoo:
Haha, no, you're all good. There's some variance in there that sometimes triggers weird false positives. Looks good to me!

And thanks for the perl reference: the change looks sensible :+1:

jonhoo:
Oh, and thanks for the quick fix! <3

jonhoo:
Released as `0.9.4`.

PR #165, opened 2020-03-08
Color diffusion mode

itamarst:
This does a few things:

1. Makes differential mode less saturated red and blue, so it's easier to read.
2. Adds a new color mode, tentatively called "color-diffusion."

The idea is that I want to visually draw the viewer's eyes towards the wider frames. Much like a differential flamegraph pulls you towards the places where things got worse (or better) with red (or blue), this highlights the bad bits in a normal flamegraph, i.e. the wider the frame the redder it gets.

So some questions for you:

1. Do you have a better name?
2. Are you willing to accept this? The alternative is some sort of plugable color generation API, but it would have to take a lot of inputs to cover all cases.


tests/data/flamegraph/perf-vertx-stacks/perf-vertx-stacks-01-collapsed-all.svg
```diff
@@ -1,1844 +0,0 @@
-<?xml version="1.0" standalone="no"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd"><svg version="1.1" width="1200" height="1190" onload="init(evt)" viewBox="0 0 1200 1190" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
jasonrhansen: It looks like the contents of this file got deleted, which is causing some of the tests to fail.

jasonrhansen:
I personally think the color diffusion mode is a nice addition.

I also like the look of the less saturated differential flamegraphs. It makes the text more readable when the sample diffs are larger. On the other hand, I think it makes it harder to see the paler hues. If the number of samples only differs by a small amount, it almost looks like white. At least on my monitor, in the second image below I can barely see the lightest red and blue.

**Before change**
image

**After change**
image

jonhoo:
It would be really good to see some more before/after pictures to judge the effect of this. In general I'm for the change though.

tests/data/flamegraph/perf-vertx-stacks/perf-vertx-stacks-01-collapsed-all.svg
itamarst: Should be fixed now.

itamarst:
Looks like I can get some more example data from https://github.com/corpaul/flamegraphdiff/tree/master/demos, let me see what they look like.

itamarst:
Nope, looks like artificial dataset, or at least just one or two function tweaked.

itamarst:
OK, here is a more realistic example, an optimization I just tried for my Python memory profiler (yes, I'm generating flamegraphs for my flamegraph-generating project :grin: ).

Now, you might say "where did all the blue go!" but if you mouse over the you'll see it's like 0.20% improvement and other similarly tiny changes. 0.20% improvement is mostly noise, so having it be less visible is fine. And maybe it should be _slightly_ more blue.

[comparison.zip](https://github.com/jonhoo/inferno/files/4326614/comparison.zip)

jonhoo:
I think _slightly_ more contrast might be good, but apart from that I'm happy with it :)

src/flamegraph/mod.rs
```diff
@@ -232,6 +232,9 @@ pub struct Options<'a> {
     /// This is only meant to be used in tests.
     #[doc(hidden)]
     pub no_javascript: bool,
+
+    /// Diffusion-based color: the wider the frame, the brighter the color.
```
jonhoo: I think it would be good to document this option more. In particular, it would be good to include what kind of uses it is useful for.

jonhoo:
Beautiful, thank you!

jonhoo:
Ah, could you also add an entry to the "Added" section under "Unreleased" in `CHANGELOG.md` briefly describing the change and with a link to the PR?

Issue #167, opened 2020-03-29
thread 'main' panicked at 'assertion failed: self.event_filter.is_some()


jasonwilliams:
I've moved this into a separate issue.

Steps to reproduce.
1. Clone https://github.com/jasonwilliams/boa
2. Use the Dockerfile provided
3. install inferno and perf 
3. in cargo.toml set debug to `true`
4. Run the commands below

```
root@ab0c08953953:/workspaces/boa# perf record --call-graph dwarf target/release/boa_cli tests/js/test.js 
undefined[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.006 MB perf.data ]
root@ab0c08953953:/workspaces/boa# perf script | inferno-collapse-perf > stacks.folded
thread 'main' panicked at 'assertion failed: self.event_filter.is_some()', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/inferno-0.
9.5/src/collapse/perf.rs:165:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Version:
inferno-flamegraph 0.9.5

jonhoo:
Will be fixed by #168!

jasonwilliams:
Thanks!

PR #168, opened 2020-03-29
perf: fix a number of corner-cases for small traces

jonhoo:
This patch set includes three somewhat orthogonal fixes:

 - It handles combined event/stack lines.
 - It avoids crashing on empty traces (#167)
 - It ensures that we also parse the last sample of a file just before hitting EOF

That last patch is a bit tricky, because upstream flamegraph does _not_ do this, and so we now fail all of their perf collapse tests. I'll file an issue there and we can merge when that gets fixed upstream.


jonhoo:
Upstream issue: https://github.com/brendangregg/FlameGraph/issues/230

jonhoo:
Upstream fix in https://github.com/brendangregg/FlameGraph/pull/231

PR #169, opened 2020-04-01
Mention the default value of Options::text_truncation_direction inline

mineo:
TextTruncateDirection gets its default value from an implementation of the
Default trait, not via an entry in the defaults module. As such, the link to its
default value didn't actually work: on
https://docs.rs/inferno/0.9.5/inferno/flamegraph/struct.Options.html#structfield.text_truncate_direction,
it's a link to `defaults::TEXT_TRUNCATE_DIRECTION` (no http or anything useful).
Similar to Options::direction, put `(the default)` after the default value in
the description.


jonhoo:
Ah, good catch, thanks!

PR #174, opened 2020-05-26
Fix #173 Templated C++ functions cut off after opening parenthesis in template arguments.

### LINKED ISSUE #173: Templated C++ functions cut off after opening parenthesis in template arguments.
Same issue as https://github.com/brendangregg/FlameGraph/pull/215 on the original tool.

SoftwareApe:
Fix too eager truncation of template parameters on opening parentheses in C++

* Add failing tests for C++ with std::function in template arguments.
* Use rfind to find the last opening parenthesis, which is most likely the function aruguments
* Check if the opening parenthesis is inside a C++ function template, by counting opening vs closing angle brackets.


SoftwareApe:
Fixes #173

jonhoo:
I _think_ the failures here are legitimate (you can see the actual error by clicking "Details", then "View more details on Azure Pipelines", then clicking the failing build step in the list on the left.

The style check [fails](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1358&view=logs&jobId=ef6be8fe-0e19-5a3e-9b53-b5954423c060&j=ef6be8fe-0e19-5a3e-9b53-b5954423c060&t=216528f1-1cdc-56ce-f78c-2a0fd5a7e3e3) because you haven't run `rustfmt` (`cargo fmt` should fix it for you).

The main tests [fail](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1358&view=logs&jobId=b0e3b9f2-f861-5d62-5d93-2be3f58ecbf2&j=b0e3b9f2-f861-5d62-5d93-2be3f58ecbf2&t=8bed322d-c384-55e1-f555-9d6311d97c43) because this changes the output of an existing test:
```diff
-"java;start_thread;java_start;GCTaskThread::run;ThreadRootsTask::do_it;JavaThread::oops_do;frame::oops_do_internal;OopMapSet::all_do;PSRootsClosure<false>::do_oop;oopDesc* PSPromotionManager::copy_to_survivor_space<false>;InstanceKlass::oop_push_contents 1"
+"java;start_thread;java_start;GCTaskThread::run;ThreadRootsTask::do_it;JavaThread::oops_do;frame::oops_do_internal;OopMapSet::all_do(frame const*, RegisterMap const*, OopClosure*, void (*);PSRootsClosure<false>::do_oop;oopDesc* PSPromotionManager::copy_to_survivor_space<false>;InstanceKlass::oop_push_contents 1"
```
Specifically:
```diff
-"OopMapSet::all_do;
+"OopMapSet::all_do(frame const*, RegisterMap const*, OopClosure*, void (*);
```

I think what's going wrong is that your new code fails to take into account when there are parentheses inside the function arguments too. In that case, your code will only remove what trails the _last_ `(`, whereas the old code would strip entire argument list (which is what should happen).

jonhoo:
You should also be able to reproduce the test failure locally by running `cargo test`?

SoftwareApe:
cargo test only says I'm missing some files, however it does the same without the change.

```log
running 24 tests
test collapse::dtrace::tests::bench_nstacks_dtrace ... ignored
test collapse::common::tests::fix_partially_demangled_rust_symbol_on_fully_demangled_symbols ... ok
test collapse::dtrace::tests::fuzz_collapse_dtrace ... ignored
test collapse::dtrace::tests::cpp_test ... ok
test collapse::dtrace::tests::test_collapse_multi_dtrace ... FAILED
test collapse::dtrace::tests::test_collapse_multi_dtrace_simple ... FAILED
test collapse::perf::tests::bench_nstacks_perf ... ignored
test collapse::perf::tests::fuzz_collapse_perf ... ignored
test collapse::common::tests::fix_partially_demangled_rust_symbol_on_fully_mangled_symbols ... ok
test collapse::common::tests::fix_partially_demangled_rust_symbols ... ok
test collapse::perf::tests::test_collapse_multi_perf ... FAILED
test flamegraph::attrs::test::func_frame_attrs_map_from_reader ... ok
test flamegraph::color::palette_map::tests::load_from_non_existing_file ... ok
test collapse::perf::tests::test_collapse_multi_perf_simple ... FAILED
test flamegraph::color::palette_map::tests::palette_map_test ... ok
test flamegraph::color::palette_map::tests::parse_line_test ... ok
test flamegraph::color::palettes::tests::java_mod_resolves ... ok
test flamegraph::color::palettes::tests::perl_mod_resolves ... ok
test flamegraph::color::tests::bgcolor_parse_test ... ok
test flamegraph::color::palettes::tests::js_returns_correct ... ok
test flamegraph::tests::top_ypadding_adjusts_for_subtitle ... ok
test flamegraph::color::tests::namehash_test ... ok
test flamegraph::rand::test_rng ... ok
test collapse::dtrace::tests::test_collapse_multi_dtrace_stop_early ... ok

failures:

Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', <::std::macros::panic macros>:5:6

Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', <::std::macros::panic macros>:5:6

Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', <::std::macros::panic macros>:5:6

Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', <::std::macros::panic macros>:5:6


failures:
    collapse::dtrace::tests::test_collapse_multi_dtrace
    collapse::dtrace::tests::test_collapse_multi_dtrace_simple
    collapse::perf::tests::test_collapse_multi_perf
    collapse::perf::tests::test_collapse_multi_perf_simple

test result: FAILED. 16 passed; 4 failed; 4 ignored; 0 measured; 0 filtered out
```

jonhoo:
Yeah, the submodule is only needed for tests. Sorry! I wish there was a way to tell git to always check out the submodule when cloning a repo :'(

SoftwareApe:
So I found the Rust stacks were also mangled due to, e.g. ...FromIterator<(K,V)>>::from_iter being cut off at the last parenthesis.
So it's not only C++ templates that are an issue here.

However the js expectations are also "wrong" now. I'm saying wrong, since they don't get cut off when they have template parameters. My basic question would be why we even want to get rid of the parameter list? After all if it just says Call or Invoke, you know very little about the call stack. With the parameters it's much more informative.

Wouldn't it be better to just leave the parameter lists inside?

SoftwareApe:
For example, without the parentheses this wouldn't be informative at all:
v8::Function::Call(v8::Handle<v8::Value>, int, v8::Handle<v8::Value>*)

jonhoo:
The answer is (perhaps unhelpfully) that we do it because that's what flamegraph did. Now, that's not a very good reason in and of itself, but my _guess_ was that they decided that the argument lists in general only contain redundant information once you have the function name. And I think in general that observation is correct. For the `v8` example you give above, for example, what information does the arguments give that you can't get from the source definition of `v8::Function::Call`?

My preference would be to stick with pruning argument lists for now at least.

src/collapse/perf.rs
```diff
@@ -609,7 +609,14 @@ fn tidy_generic(mut func: String) -> String {
     //    see https://github.com/brendangregg/FlameGraph/pull/72
     //  - C++ anonymous namespace annotations.
     //    see https://github.com/brendangregg/FlameGraph/pull/93
-    if let Some(first_paren) = func.find('(') {
+
+    // Check for C++ template parameters
+    let mut offset = 0usize;
+    if let Some(last_closing_angle_bracket) = func.rfind('>') {
```
jonhoo, line 615: I don't think this will quite work either. Consider, for example:
```rust
foo(Vec<usize>)
```

jonhoo:
Oh, by the way, we should probably add all these interesting weird cases we come across now as unit tests for that function!

src/collapse/perf.rs
SoftwareApe: True, probably actually need to do some recursive descent parsing here.

SoftwareApe:
> The answer is (perhaps unhelpfully) that we do it because that's what flamegraph did. Now, that's not a very good reason in and of itself, but my _guess_ was that they decided that the argument lists in general only contain redundant information once you have the function name. And I think in general that observation is correct. For the `v8` example you give above, for example, what information does the arguments give that you can't get from the source definition of `v8::Function::Call`?
> 
> My preference would be to stick with pruning argument lists for now at least.

I'd say you don't know anything about which function is called here, if there's no argument list.

>Oh, by the way, we should probably add all these interesting weird cases we come across now as unit tests for that function!

I already fixed the broken test cases on the Rust example.

jonhoo:
Right, but the argument list that you printed also tells you nothing more, since those are just the arguments that `Call` takes, not the _values_ that were passed. Unless you're saying that `v8::Function::Call` is special, and the arguments that get printed are the arguments of the underlying JavaScript function that was called? If so, it feels like something we should special-case. In general, the function arguments are at least deterministic from the function name.

> I already fixed the broken test cases on the Rust example.

Ah, yes, what I meant was more that we should add these as unit tests for `tidy_generic` so that a future change doesn't accidentally regress on parsing these things. And it's easier to figure these things out from a unit test than a giant folded file diff.

SoftwareApe:
I now went with the matching of parenthesis. The tidying now doesn't break the old tests (except the faulty Rust tests, which I updated), I added the small unit test for tidy_generic with some edge cases we found here.

src/collapse/perf.rs
```diff
@@ -609,25 +609,59 @@ fn tidy_generic(mut func: String) -> String {
...
+            // Filter chars depending on state
+            if !is_go_function && (end_of_param_list || parentheses_depth > 0) {
+                None
+            } else {
+                Some(c)
             }
-        }
-    }
+        })
+        .collect();
```
jonhoo, line 664: I think instead of filtering and then collecting to make the new string, you can just use `.find` and return `true` on the first `(` that occurs outside of angle brackets (by counting them the way you're currently doing). And then you truncate from that index and out. `char_indices` gives you access to the byte index. Something like:
```rust
let mut angle_depth = 0;
let mut paren_depth = 0;
let arg_start = func.char_indices().find_map(|(i, c)| {
  match c {
    '<' => { angle_depth += 1; },
    '>' => { angle_depth -= 1; }
    '(' if angle_depth == 0 => {
      // TODO: take Go into account here (by counting `()` after `.`)
      return Some(i);
    }
    _ => {}
  }
});
if let Some(i) = arg_start {
  func.truncate(i);
}
```

SoftwareApe: Ok, done.

jonhoo:
This looks great now, thank you!

SoftwareApe:
Thank you for your quick response and helpful comments.

jonhoo:
Thank you for sticking with it even with all my comments :sweat_smile:
Released as 0.9.7!

jonhoo:
Oh, I also took the liberty of squashing your commit and cleaning up the commit message a little. Hope you don't mind. See https://github.com/jonhoo/inferno/commit/2230ac794280b3f109de44ae382a9f307f8f8502.

PR #175, opened 2020-05-30
Fix regression on anonymous namespaces.

SoftwareApe:
I remembered this morning that there was a special case on anonymous namespaces in C++, and indeed I had accidentally removed that. Since there was no unit test for it I added a unit test for anonymous namespaces on tidy_generic so this regression shouldn't happen again.

While at it I saw that there was a different kind of special case stack entry in C++ with lambdas written as {lambda(int, int)#2}, which again got truncated. So I also made the bracket detection in tidy_generic more generic, of course with tests :smile:.


jonhoo:
Oh, nice catch!
This does assume that brackets are balanced, but that seems like an okay assumption to make.

PR #177, opened 2020-06-03
Move details bar to top in icicle mode.

### LINKED ISSUE #176: In icicle mode, the bar at bottom should really be at the top
When you mouseover a frame there's a little bar at the bottom of the diagram that shows info like "33% of samples" or "20% of RAM" or whatever.

In icicle mode most of the interesting info is at the top, and so you never actually get to see this bar if you have a long callstack. This is pretty much always the case for my use case—see e.g. https://pythonspeed.com/products/filmemoryprofiler/memory-graph.svg, where the really tall callstack on the left means you never get to see the bottom bar.

It seems that for icicle mode this bar should always be at the top.

I will attempt to write a PR for this.

itamarst:
Fixes #176.

### DIFF
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -248,7 +248,12 @@ impl<'a> Options<'a> {
         } else {
             0
         };
-        self.font_size * 3 + subtitle_height
+        if self.direction == Direction::Straight {
+            self.font_size * 3 + subtitle_height
+        } else {
+            // Inverted (icicle) mode, put the details on top:
+            self.font_size * 4 + subtitle_height + 4
+        }
     }
 
     /// Calculate pad bottom, including labels
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -862,8 +862,9 @@ fn write_usize(buffer: &mut StrStack, value: usize) -> usize {
 
 #[cfg(test)]
 mod tests {
-    use super::Options;
+    use super::{Direction, Options};
 
+    // If there's a subtitle, we need to adjust the top height:
     #[test]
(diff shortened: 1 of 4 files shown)


src/flamegraph/mod.rs
```diff
@@ -248,7 +248,12 @@ impl<'a> Options<'a> {
         } else {
             0
         };
-        self.font_size * 3 + subtitle_height
+        if self.direction == Direction::Straight {
+            self.font_size * 3 + subtitle_height
+        } else {
+            // Inverted (icicle) mode, put the details on top:
+            self.font_size * 4 + subtitle_height + 4
```
jonhoo, line 255: Hmm, why is the `+4` needed here?
The comment could probably also be a bit more helpful.

src/flamegraph/mod.rs
```diff
@@ -248,7 +248,12 @@ impl<'a> Options<'a> {
...
+        }
```
jonhoo, line 256: I think `ypad2` also has to change to _not_ include padding for the details at the bottom.

itamarst:
Guess I should write a unit test too.

itamarst:
BTW, just realized you're in Cambridge, MA too. Could I get you to sign and share https://bit.ly/3eNdnhK? We're trying to hit 1000 people by the time we submit it tomorrow :grin:

jonhoo:
I'm not currently in a position to test this, but I assume you have tested that both old and new flamegraphs still work? :p

itamarst:
Yeah. I can try to upload the results... 

[newcode.zip](https://github.com/jonhoo/inferno/files/4726221/newcode.zip)

jonhoo:
Looks great, thank you!

PR #180, opened 2020-06-18
Support symbol mangle v0

cfeitong:
Dtrace doesn't demangle symbol mangle v0. Inferno needs to do it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ quick-xml = { version = "0.18", default-features = false }
 rgb = "0.8.13"
 str_stack = "0.1"
 structopt = { version = "0.3", optional = true }
+rustc-demangle = "0.1"
 
 [dev-dependencies]
 assert_cmd = "1"
--- a/src/collapse/dtrace.rs
+++ b/src/collapse/dtrace.rs
@@ -299,6 +299,10 @@ impl Folder {
     fn fix_rust_symbol<'a>(&self, frame: &'a str) -> Cow<'a, str> {
         let mut parts = frame.splitn(2, '`');
         if let (Some(pname), Some(func)) = (parts.next(), parts.next()) {
+            // Support rust symbol mangling (v0)
+            if let Ok(demangled_func) = rustc_demangle::try_demangle(func) {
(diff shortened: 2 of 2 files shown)


jonhoo:
Interesting — are you saying that dtrace doesn't demangle at all, or that its demangling is just incomplete? It'd be good to have some sample files to work with here that we can then use for tests.

jasonrhansen:
Demangling currently works for the legacy (still the default) rustc mangler. On nightly you can use the the new `v0` mangler, which is what @cfeitong wants inferno to be able to demangle.

This PR in its current state should get part of the way there. For dtrace it will need to properly handle the case where the `includeoffset` option is set. Also, we might want to add this to the other collapsers, and make it optional.

jonhoo:
Ohh, I see, interesting. Is there a way for us to detect if a string needs demangling with the v0 demangler? Demangling an already-demangled string for example I can image might do some super strange things to your output.

jasonrhansen:
`rustc_demangle::demangle` will return the original value if it doesn't look like a mangled symbol so I don't think we would need to do any detection. That being said, it might be a good idea to make it optional.

We used to have a `--demangle` option that used `symbolic_demangle` to demangle symbols, but we decided to remove it because the profilers do their own demangling. However, Rust symbols weren't getting demangled properly. They would get partially demangled, but not completely. These partially demangled symbols couldn't be demangled by `rustc_demangle`, so we added a function to inferno that could fix them. But this only works with the legacy mangler.

Since the legacy mangler is still the default, and the plan is to upstream the C implementation of the v0 demangler to the various tools, it might not even be worth adding this. Eventually the profilers should be able to demangle the v0 mangled symbols.

This is the tracking issue for the v0 mangling: https://github.com/rust-lang/rust/issues/60705

jonhoo:
Yeah, I'm inclined to not second-guess the upstream tools, and instead have those tools do the demangling where possible. It may take a bit longer to land for users, but I worry about having "extra" demanglers in the stack.

jasonrhansen:
I would also prefer not to add in a demangler and just wait for all of the upstream tools to support v0.

PR #181, opened 2020-06-20
Mark all Option types non_exhaustive

jonhoo:
This is a breaking change.


jonhoo:
See also https://github.com/rust-lang/rust/issues/70564#issuecomment-647031324

Issue #26, opened 2019-02-03
Adopt changes from upstream pull requests


jonhoo:
The upstream FlameGraph project has a number of outstanding [pull requests](https://github.com/brendangregg/FlameGraph/pulls) that fix bugs and add or improve features. We should take a look through them and see whether we may be able to incorporate some of them here! I propose we take the following approach for each one:

 - First, see whether the change is still relevant in inferno
 - Then, comment on the issue asking the original author whether they mind having their change adopted into inferno (we can't just take without permission -- copyright and licensing is no joke!)
 - If they agree, create a tracking issue on inferno describing the original PR, link to it + the author's agreement to port, and add the `adoption` label. Then, either starting writing a PR if you want to implement it yourself, or just leave it there for others to adopt!
 - If they don't respond, and you believe it to be an important feature, feel free to create a corresponding issue on inferno and write a PR, but note _explicitly_ in both that you don't have the author's approval to port the change. We won't merge those until we do, but we can at least do the work in advance!
 - If they explicitly say they're unwilling to allow the change to be made in inferno, well, then we're our of luck. The exception would be if you can implement the feature _without consulting their changes at all_!

We should also probably take a look through [known FlameGraph issues](https://github.com/brendangregg/FlameGraph/issues) and see which ones also apply to inferno. For those that do, we should add tracking issues here, link to the original issue, and ideally also add a failing test case where possible!

jbcden:
Hi @versable I am trying to help out the [inferno project](https://github.com/jonhoo/inferno/) which is a rewrite of Flamegraph in Rust. Amoung other improvements to the port, [we are looking at existing open changes in this project that we might be able to bring over](https://github.com/jonhoo/inferno/issues/26). Would you be ok with us adapting [your PR for SVG improvements(https://github.com/brendangregg/FlameGraph/pull/189) for inferno? I am not certain all of the changes apply, but at least a few of them seem very promising!

ErichDonGubler:
@jonhoo: It looks like there are currently 2 PRs in the queue, one of which you seem to be actively handling (https://github.com/jonhoo/inferno/pull/32) and the other is waiting on the OP to fix some upstream issues they discovered with the PR (https://github.com/jonhoo/inferno/pull/38). Is there anything left for this issue specifically?

jonhoo:
[...]

PR #183, opened 2020-06-21
Support of AsyncProfiler generated stack trace

AnderEnder:
* based on Async Profiler flamegraph [implementation](https://github.com/jvm-profiling-tools/async-profiler/blob/master/src/flameGraph.cpp#L748-L770)
* add exception for `.` case to support correct behavior with another java traces
* fix test case to support new behaviour
* add more test cases to test new behavior
* add integration test with example async-profiler collapsed file

Related: https://github.com/jonhoo/inferno/issues/26
Backport PR(idea): https://github.com/brendangregg/FlameGraph/pull/234


AnderEnder:
@jonhoo, Async-profiler can produce raw output: traces or already collapsed.  For this case I received a huge file with traces from async-profiler and I did not find a way to convert it to collapsed format. I made a dumb convertor. Do you think it would be valuable to put this convertor implementation into inferno tool set as well?

jonhoo:
Hmm, that's an interesting question.. I don't know enough about the Java async space to really say, but if it's a common enough format, it seems pretty reasonable to add `inferno-collapse-java-async`.

AnderEnder:
@jonhoo I asked my friend and did not received the straight answer: use cases are different. I think this question could be moved to an issue and if anyone will require this use case we can implement it.

src/flamegraph/color/palettes.rs
```diff
@@ -269,11 +267,11 @@ mod tests {
             },
             TestData {
                 input: String::from("jdk/::[ki]"),
-                output: BasicPalette::Green,
+                output: BasicPalette::Yellow,
```
jonhoo, line 270: Interesting.. Shouldn't this still be green?

AnderEnder: This case is contraversal, I think It should be c++ call due to `::`, so yellow looks correct.

jonhoo: Hmm, curious.. My worry is that this is a test that stemmed from some real-world use-case where that popped up in Java code (because of `jdk`). But I could be misremembering its origins, and maybe it was just something we made up?

AnderEnder: I believe it is a synthetic test case to cover the priority matching of previous implementation(https://github.com/jonhoo/inferno/pull/131). In the current implementation the priority was changed. It does not look that a method in java could contain `::` character, but I have not seen `/` in c++ parts too.

Sure we can make an exception for this case in matching to keep this test case to work in the same way, but I believe the test result should be changed instead.

jonhoo: Seems reasonable to me, I just wanted to double-check!

jonhoo:
Released in 0.10.1.

PR #190, opened 2020-09-27
Implement --deterministic, for chaotic but repeatable color selection

saethlin:
From the runs I've tried out with this so far it seems like the color selection is a bit biased towards darker shades, which is unfortunate. Not entirely sure what to do about that.

### DIFF
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -20,9 +20,14 @@ struct Opt {
     #[structopt(long = "cp")]
     cp: bool,
 
-    /// Colors are keyed by function name hash
+    /// Colors are selected by hashing the function name, weighting the first characters more
+    /// heavily
     #[structopt(long = "hash")]
     hash: bool,
+
+    /// Colors are selected chaotically but the color of a function does not change between runs
+    #[structopt(long = "deterministic")]
+    deterministic: bool,
 
     /// Plot the flame graph up-side-down
     #[structopt(short = "i", long = "inverted")]
@@ -208,6 +213,7 @@ impl<'a> Opt {
         options.colors = self.colors;
         options.bgcolors = self.bgcolors;
         options.hash = self.hash;
+        options.deterministic = self.deterministic;
 
         self.set_func_frameattrs(&mut options);
 
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -26,7 +26,7 @@ struct Opt {
     hash: bool,
 
     /// Colors are selected such that the color of a function does not change between runs
-    #[structopt(long = "deterministic")]
+    #[structopt(long = "deterministic", conflicts_with = "hash")]
     deterministic: bool,
 
     /// Plot the flame graph up-side-down
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -22,11 +22,11 @@ struct Opt {
 
     /// Colors are selected by hashing the function name, weighting the first characters more
     /// heavily
-    #[structopt(long = "hash")]
+    #[structopt(long = "hash", conflicts_with = "deterministic")]
     hash: bool,
 
     /// Colors are selected chaotically but the color of a function does not change between runs
-    #[structopt(long = "deterministic")]
+    #[structopt(long = "deterministic", conflicts_with = "hash")]
     deterministic: bool,
 
     /// Plot the flame graph up-side-down
--- a/src/flamegraph/color/mod.rs
+++ b/src/flamegraph/color/mod.rs
@@ -331,6 +331,7 @@ fn rgb_components_for_palette(palette: Palette, name: &str, v1: f32, v2: f32, v3
 pub(super) fn color(
     palette: Palette,
     hash: bool,
+    deterministic: bool,
     name: &str,
     mut rng: impl FnMut() -> f32,
 ) -> Color {
@@ -339,6 +340,13 @@ pub(super) fn color(
         let reverse_name_hash = namehash(name.bytes().rev());
 
         (name_hash, reverse_name_hash, reverse_name_hash)
+    } else if deterministic {
+        use std::hash::Hasher;
+        let mut hasher = ahash::AHasher::default();
+        hasher.write(name.as_bytes());
+        let hash = hasher.finish() as f32 / u64::MAX as f32;
+
+        (hash, hash, hash)
     } else {
         (rng(), rng(), rng())
     };
--- a/src/flamegraph/color/mod.rs
+++ b/src/flamegraph/color/mod.rs
@@ -342,7 +342,8 @@ pub(super) fn color(
         (name_hash, reverse_name_hash, reverse_name_hash)
     } else if deterministic {
         use std::hash::Hasher;
-        let mut hasher = ahash::AHasher::default();
+        // Do not use AHasher::default; it is seeded with a compile-time RNG
+        let mut hasher = ahash::AHasher::new_with_keys(1234, 5678);
         hasher.write(name.as_bytes());
         let hash1 = (hasher.finish() as f64 / std::u64::MAX as f64) as f32;
         hasher.write_u8(0);
--- a/src/flamegraph/color/mod.rs
+++ b/src/flamegraph/color/mod.rs
@@ -342,17 +342,15 @@ pub(super) fn color(
         (name_hash, reverse_name_hash, reverse_name_hash)
     } else if deterministic {
         use std::hash::Hasher;
-        let prefixed_hash = |name: &str, prefix: u8| {
-            let mut hasher = ahash::AHasher::default();
-            hasher.write_u8(prefix);
-            hasher.write(name.as_bytes());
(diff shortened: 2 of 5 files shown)


src/bin/flamegraph.rs
```diff
@@ -20,10 +20,15 @@ struct Opt {
     #[structopt(long = "cp")]
     cp: bool,
 
-    /// Colors are keyed by function name hash
+    /// Colors are selected by hashing the function name, weighting the first characters more
```
jonhoo, line 23:
```suggestion
    /// Colors are selected by hashing the function name, weighting earlier characters more
```

src/bin/flamegraph.rs
```diff
@@ -20,10 +20,15 @@ struct Opt {
...
+    /// heavily
     #[structopt(long = "hash")]
     hash: bool,
 
+    /// Colors are selected chaotically but the color of a function does not change between runs
```
jonhoo, line 28:
```suggestion
    /// Colors are selected such that the color of a function does not change between runs
```

src/bin/flamegraph.rs
```diff
@@ -20,10 +20,15 @@ struct Opt {
...
+    #[structopt(long = "deterministic")]
```
jonhoo, line 29: This probably `conflicts_with = "hash"`.

src/flamegraph/color/mod.rs
```diff
@@ -339,6 +340,13 @@ pub(super) fn color(
         let reverse_name_hash = namehash(name.bytes().rev());
 
         (name_hash, reverse_name_hash, reverse_name_hash)
+    } else if deterministic {
+        use std::hash::Hasher;
+        let mut hasher = ahash::AHasher::default();
+        hasher.write(name.as_bytes());
+        let hash = hasher.finish() as f32 / u64::MAX as f32;
```
jonhoo, line 347: This is probably going to give some weird rounding errors, since `f32`'s coverage over `u64` is quite bad. You may want to use `f64` for the casts, and then cast the resulting fraction to `f32`.

src/flamegraph/color/mod.rs
```diff
@@ -339,6 +340,13 @@ pub(super) fn color(
...
+
+        (hash, hash, hash)
```
jonhoo, line 349: One consequence of setting these three to the exact same value is that the different RGB components are all scaled the same way. You'd get more variance if you generated a second (deterministic) hash and used that as `v2` and `v3`. One option would be to do:
```rust
hasher.write_u8(0);
let hash2 = hasher.finish() ...

(hash, hash2, hash2)
```

jonhoo:
Thanks, this looks pretty good! Left some comments inline. I'm not sure why the colors are generally dark, though the second hash trick I mentioned might help. I'd also like to see a new test case for this.

src/flamegraph/color/mod.rs
saethlin: This fixes the color selection oddness.

saethlin:
I'd love to add some tests, but... how? The existing test suite doesn't run on its own, it seems like maybe it's missing some files?
```
Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /home/ben/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/test/src/lib.rs:192:5

Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /home/ben/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/test/src/lib.rs:192:5

Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /home/ben/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/test/src/lib.rs:192:5

Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /home/ben/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/test/src/lib.rs:192:5
```

src/bin/flamegraph.rs
jonhoo: Even though it's probably unnecessary, I'd also add `conflicts_with = "hash"` to this option as well.

jonhoo:
Ah, yes, it also runs a bunch of tests from the upstream Perl flamegraph repository, which has to be checked out with
```console
$ git submodule update --init
```

src/flamegraph/color/mod.rs
```diff
@@ -339,6 +340,19 @@ pub(super) fn color(
...
 
         (name_hash, reverse_name_hash, reverse_name_hash)
+    } else if deterministic {
+        use std::hash::Hasher;
+        let prefixed_hash = |name: &str, prefix: u8| {
+            let mut hasher = ahash::AHasher::default();
+            hasher.write_u8(prefix);
+            hasher.write(name.as_bytes());
+            (hasher.finish() as f64 / u64::MAX as f64) as f32
+        };
```
jonhoo, line 350: Ah, this is more extensive than what I had in mind, and is probably excessive given how hashing works. It should be sufficient (and faster) to literally do:

```rust
let mut hasher = ahash::AHasher::default();
hasher.write(name.as_bytes());
let hash = (hasher.finish() as f64 / u64::MAX as f64) as f32;
hasher.write_u8(0);
let hash2 = (hasher.finish() as f64 / u64::MAX as f64) as f32;
(hash, hash2, hash2)
```

saethlin: Is there a reason you're suggesting using only two values, when this returns a 3-tuple?

jonhoo: You can do a third too if you want. I'm using the same value for the two last values just because that's the same as the hashing mode does:
https://github.com/jonhoo/inferno/blob/86302c8ddf57281ba982b0da0a1c01aa87eea7fe/src/flamegraph/color/mod.rs#L341

saethlin:
@jonhoo The test I've added passes on my machine, I would really appreciate some assistance here. I don't know why it's not working in the CI or whether this is even the test we want.

jonhoo (dismissed):

jonhoo:
Hmm, looking at the [test output](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1713&view=logs&j=b0e3b9f2-f861-5d62-5d93-2be3f58ecbf2&t=8bed322d-c384-55e1-f555-9d6311d97c43&l=462), it seems like the test run generates a different  color than the committed test file. Even more interestingly, different [runs](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1713&view=logs&j=cc2e323d-a806-545f-bf6b-e1565f70f846&t=1b044aa0-9aef-56b9-bdd7-1f415a470aaf&l=462) of the test produce different values, which suggest that the generated color is not actually deterministic. Are you sure the test is actually being run on your computer? The failure seems to be consistent across all the different OSes.

Separately, you have to use `std::u64::MAX`, not `u64::MAX`, otherwise the code [will not compile on Rust 1.40](https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=1713&view=logs&j=92335eda-b3fc-5c3c-d70a-38a67693cfa7&t=b68c524f-8467-530d-d460-460d24c90ded&l=112).

src/flamegraph/color/mod.rs
```diff
@@ -342,17 +342,15 @@ pub(super) fn color(
...
-            prefixed_hash(name, 0),
-            prefixed_hash(name, 1),
-            prefixed_hash(name, 2),
-        )
+        let mut hasher = ahash::AHasher::default();
+        hasher.write(name.as_bytes());
+        let hash1 = (hasher.finish() as f64 / u64::MAX as f64) as f32;
+        hasher.write_u8(0);
+        let hash2 = (hasher.finish() as f64 / u64::MAX as f64) as f32;
+        hasher.write_u8(1);
```
jonhoo, line 350: You can actually hash `0` again here — the output should still be different.

src/flamegraph/color/mod.rs
```diff
@@ -342,7 +342,8 @@ pub(super) fn color(
         (name_hash, reverse_name_hash, reverse_name_hash)
     } else if deterministic {
         use std::hash::Hasher;
-        let mut hasher = ahash::AHasher::default();
+        // Do not use AHasher::default; it is seeded with a compile-time RNG
```
jonhoo, line 345: Ah, good catch!

jonhoo (approved):

jonhoo:
Excellent, thank you!

jonhoo:
Released in 0.10.1.

Issue #102, opened 2019-03-18
Optimize implementation for when there are many frames


jonhoo:
#99 identified that performance drops significantly when there are many frames in the output (see https://github.com/jonhoo/inferno/pull/99#issuecomment-473726169). We should figure out why that is, and fix it! Maybe it's time to flamegraph inferno ;)

jasonrhansen:
As a start to this, I profiled and generated a flame graph of running `inferno-flamegraph` with `--reverse` to create many frames. It looks like a lot of time is spent in `svg::write_str` and `filled_rectangle`. Both of these spend most of the time adding attributes. I think implementing #50 will reduce the number of attributes written? 

Very little time is spent in `merge::frames`. It's really just all of the attribute creation and formatting/writing the SVG tags that's taking most of the time.

![image](https://user-images.githubusercontent.com/1126521/54729390-dbde0580-4b48-11e9-8278-5265627f5c42.png)
![image](https://user-images.githubusercontent.com/1126521/54729503-8ce4a000-4b49-11e9-8d1b-9d988096dd21.png)
![image](https://user-images.githubusercontent.com/1126521/54729532-c0272f00-4b49-11e9-99c1-dac851bd3758.png)

[flamegraph.svg.zip](https://github.com/jonhoo/inferno/files/2990619/flamegraph.svg.zip)

jonhoo:
Hmm, that `push_attribute` stack looks suspicious. Specifically, it seems to spend a bunch of time reserving additional memory in `Vec::extend`, which I thought we got rid of that in #58 (which fixed #41).  @JordiChauzi any idea why that might be happening?

There's also a decent amount of time spent in number formatting there. I wonder whether we can use [`dtoa`](https://crates.io/crates/dtoa) to significantly speed that up? Or alternatively if @bcmyers' `num-format` crate now has support for floating point, that may be able to take that job?

The buffered writes to output we can't do much about, though you are right that #50 should help a bit with that, since there'll just be less stuff to write overall.

jasonrhansen:
`svg::write_str` isn't reusing a cached `svg::Event` like everything else. I'm going to submit a PR to fix this.

jonhoo:
So I guess now that #108 and #110 have landed, we're probably down to number formatting being the biggest one?

jasonrhansen:
[...]

PR #192, opened 2020-10-06
Faster float formatting

saethlin:
I don't have access to whatever folded stacks were in https://github.com/jonhoo/inferno/issues/102, but I've been experimenting with a 116 MB file and `--minwidth=0`. Looks like a 17% improvement with no other options, and 24% with `--reverse`.

Guidance appreciated, not yet sure how to thread this technique into the code in `from_lines`. I feel like a newtype with a customized `Display` impl might be the thing?

### DIFF
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -844,6 +844,25 @@ fn deannotate(f: &str) -> &str {
     f
 }
 
+fn format_percentage(buffer: &mut StrStack, value: f64) -> usize {
+    if value == 0.0 {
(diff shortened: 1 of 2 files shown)


jonhoo:
Ah, @bcmyers may have some good input here since he's looked at similar things in the past!

saethlin:
Closing this; I've been slowly working on a `perf.data` parser and I think there are much better wins there.

jonhoo:
Oooooh, yeah, I definitely believe that could be the case!

PR #195, opened 2020-11-09
Update cargo dependencies

Mephistophiles:


Mephistophiles:
ahash and crossbeam requires rust > 1.40

jonhoo:
That's fine — I'm okay with also bumping the minimum Rust version to anything released 6+ months ago.

PR #198, opened 2020-11-30
Bump MSRV to 1.42 for const-random-macro

jonhoo:
https://github.com/tkaitchuck/constrandom/commit/2d165d62e92986532e4e548e137d1c62d2340b49#r44645018

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,7 +2,7 @@ jobs:
  - template: default.yml@templates
    parameters:
      codecov_token: $(CODECOV_TOKEN_SECRET)
-     minrust: 1.40.0
+     minrust: 1.42.0
      env:
        RUST_BACKTRACE: 1
      setup:


jonhoo:
@austinabell Huh, it looks like the test you modified now hangs when running under CI — any idea why that might be?

austinabell:
> @austinabell Huh, it looks like the test you modified now hangs when running under CI — any idea why that might be?

Ah I see it now, didn't notice it was slow when I ran tests:
```rust
        let mut input = Vec::new();
        for _ in 0..100 {
            input.extend_from_slice(valid_stack.as_bytes());
        }
        input.extend_from_slice(invalid_stack.as_bytes());
        for _ in 0..(1024 * 1024 * 10) {
            input.extend_from_slice(valid_stack.as_bytes());
        }
```
Is the code, and for some reason the test extended this arbitrarily large slice, because the test exited early before. I will change the value to be smaller since it's being called now. I don't see the benefit in why this second part was added if it was expected to exit

PR #199, opened 2020-11-30
Switch test case to not handle arbitrarily large input

austinabell:
- This test was exiting early because there was invalid utf8 placed in between these statements, but now that it's being processed this test runs for a while to process `1024*1024*10` elements.

My apologies for not noticing this test was really slow when I added support in #196, I didn't consider this was how the test was setup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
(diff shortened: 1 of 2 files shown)


austinabell:
I also noticed you have a CHANGELOG, would you like me to add a point about the change from #196 while I'm here?

jonhoo:
Oh wow, yeah, that does seem like a weird way for it to test things.

And yes, a changelog entry would be awesome :+1:

jonhoo (approved):

jonhoo (approved):

PR #202, opened 2020-12-27
Correctly annotate jitdump stack frames

asherkin:
When using perf with the modern jitdump format, `perf inject` creates an ELF object for each JITted function and replaces the function entries in the recording to point to those SOs. This causes `stackcollapse-perf` to not recognise the functions as the module is no longer displayed as being the perf JIT map for the process.


jonhoo:
Good catch! Could you please include a test case with this new format as well?

src/collapse/perf.rs
```diff
@@ -519,8 +519,8 @@ impl Folder {
                 } else if self.opt.annotate_kernel && is_kernel(module) {
                     func.push_str("_[k]"); // kernel
                 } else if self.opt.annotate_jit
-                    && module.starts_with("/tmp/perf-")
-                    && module.ends_with(".map")
+                    && ((module.starts_with("/tmp/perf-") && module.ends_with(".map"))
+                        || (module.contains("/jitted-") && module.ends_with(".so")))
```
jonhoo, line 523: Should this be `starts_with`, not `contains`?

asherkin: `perf inject` will create the ELF files wherever the jitdump metadata file was located, and unlike the old jit interface it can be anywhere on the filesystem, so `contains` is correct here.

asherkin:
Very happy to add some tests, but it looks like the existing test infrastructure does not test with any of the options enabled (so this would be a no-op) - so I could do with some guidance on how you'd like them added.

jonhoo:
Not sure I follow? The Java flamegraph test for example specifically runs with the java color palette:
https://github.com/jonhoo/inferno/blob/f2f1e9ab07b0f7a669d290c1552024f9b7ddf5ae/tests/flamegraph.rs#L135-L146

asherkin:
Test added - I had only seen the unit tests and completely missed the integration tests 😄

jonhoo:
I also just realized I linked you to the flamegraph tests :facepalm: But glad you found them!

jonhoo (dismissed):

jonhoo:
Would you mind also updating the `CHANGELOG`?

jonhoo (approved):

PR #203, opened 2020-12-27
Add a generic palette to highlight kernel / jitted code

asherkin:
The "java" and "js" colour palettes have a very useful feature of being able to highlight which frames are jitted or kernel functions based on the annotations added to the function makes by the stackcollapse programs. Unfortunately they also make a number of assumptions based on the function names which are very specific to those languages, and cause nonsensical results with other runtimes.

This PR adds a new generic "annotated" colour palette that only uses the function name annotations, and can thus be used regardless of the runtime being profiled.

### DIFF
--- a/src/flamegraph/color/palettes.rs
+++ b/src/flamegraph/color/palettes.rs
@@ -1,3 +1,29 @@
+pub(super) mod annotated {
+    use crate::flamegraph::color::BasicPalette;
+
+    /// Handle annotations (_[j], _[i], ...; which are
+    /// accurate) in a generic way.
+    pub fn resolve(name: &str) -> BasicPalette {
+        if name.ends_with(']') {
+            if let Some(ai) = name.rfind("_[") {
+                if name[ai..].len() == 4 {
+                    match &name[ai + 2..ai + 3] {
+                        // kernel annotation
+                        "k" => return BasicPalette::Orange,
+                        // inline annotation
+                        "i" => return BasicPalette::Aqua,
+                        // jit annotation
+                        "j" => return BasicPalette::Green,
+                        _ => {}
+                    }
+                }
+            }
+        }
+
+        BasicPalette::Red
+    }
+}
+
 pub(super) mod java {
     use crate::flamegraph::color::BasicPalette;
 
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -82,7 +82,7 @@ struct Opt {
         short = "c",
         long = "colors",
         default_value = defaults::COLORS,
-        possible_values = &["aqua","blue","green","hot","io","java","js","mem","orange","perl","purple","red","wakeup","yellow"],
+        possible_values = &["annotated","aqua","blue","green","hot","io","java","js","mem","orange","perl","purple","red","wakeup","yellow"],
         value_name = "STRING"
     )]
     colors: Palette,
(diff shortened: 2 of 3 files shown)


src/flamegraph/color/palettes.rs
```diff
@@ -1,3 +1,29 @@
+pub(super) mod annotated {
+    use crate::flamegraph::color::BasicPalette;
+
+    /// Handle annotations (_[j], _[i], ...; which are
+    /// accurate) in a generic way.
+    pub fn resolve(name: &str) -> BasicPalette {
+        if name.ends_with(']') {
```
jonhoo, line 7: Neat! But since this code is the same as the Java code, maybe move it out into a shared helper function that returns `Option<BasicPalette>`?

asherkin: That looks a lot better, thanks!

jonhoo:
Just out of interest, could you post a flamegraph with this color mode enabled? Would be interesting to see what it looks like!

jonhoo (approved):

asherkin:
> Just out of interest, could you post a flamegraph with this color mode enabled? Would be interesting to see what it looks like!

https://lemon.limetech.io/srcds-startup-sp-only.svg

jonhoo:
For posterity:
2020-12-27-122618_3840x2160_scrot

That's really neat! I almost wonder if we should make this the default behavior for the `Hot` palette, and then have a `LegacyHot` that only uses red as today. @jasonrhansen @bcmyers @itamarst @AnderEnder what do you think?

jasonrhansen:
@jonhoo I wouldn't be opposed to that idea.

itamarst:
The output definitely looks useful. I would suggest documenting the heuristics, so people know what to expect.

jonhoo:
@asherkin I'm pretty happy making this the default. Could you update the PR so that `Hot` uses your new heuristic, and then also document that behavior for the hot color scheme? We'll also then want to add a `LegacyHot` color scheme that is what `Hot` does today.

Oh, and as in the other PR, could you please update the changelog at the end?

asherkin:
@jonhoo Apologies for the late response, I caught the virus that's been going around and have been out of action for a few weeks.

Would love to make this the default! I think the only concern I'd have with that myself is the massively reduced colour range of "red" used here compared to the "hot" palette. I only used these colours as they're what the java one was using, how would you feel about me crafting specifically a new default using these annotations? I'm thinking "hot" for the default (so the output will be the same as currently for anything not-annotated), something muted like greys for the kernel frames, and something bright (probably stick with the green/blue) for the jit/inline frames?

jonhoo:
Ouch, I'm sorry to hear that. Glad to have you back with us!

Absolutely, that sounds like an excellent idea :tada:

asherkin:
I've implemented the proposed new colour scheme and I think it's looking good. The default has changed to be the new annotated palette - let me know if you would prefer that to still be called "hot" from a CLI perspective and I'll make that change, right now I've kept it as its own new thing (internally it'll never be totally straightforward due to BasicPalette vs MultiPalette anyway). I'll update the changelog once that's cleared up.

I could do with some pointers regarding where you'd like documentation related to the annotation behaviour - the adding of annotations and the details around that are handled (and documented) on the stackcollapse side, so I'm not totally sure what's wanted over here on the flamegraph end.

image

jonhoo:
Nice! How about we call the old palette `legacy`, and leave the new one as `hot`?

As for documentation, I think the thing to do is mention in the stack collapsing docs how we detect JIT/kernel things, and then in the flamegraph docs how we expect names to be annotated in order to highlight them as JIT/kernel. How does that sound? I think we'll likely end up with a full API makeover (including docs) at some point in the future when we land a proper API as well :sweat_smile:

Issue #153, opened 2019-11-26
0.7 -> 0.9 upgrade caused reduced svg fidelity


liaden:
I noticed this on rbspy specifically: https://github.com/rbspy/rbspy/issues/239

jonhoo:
Hmm, interesting. I suspect this is due to https://github.com/jonhoo/inferno/pull/136. @jasonrhansen could you take a look? @liaden could you try generating the flamegraph with `min_width` set to a smaller value than the default and see if that fixes the issue?

liaden:
Just removing https://github.com/rbspy/rbspy/blob/master/src/ui/flamegraph.rs#L35 to take the default value, it works as expected.

jonhoo:
Ah, interesting! Yes, the _meaning_ of `min_with` changed in the move from 0.7 to 0.9 (I should probably have documented that in a changelog somewhere). It is not (from memory) a percentage of the image width rather than a pixel value. Should be a straightforward enough change to `rbspy` I think?

liaden:
Yeah, it definitely seems more appropriate to change on rbspy side. 

> It is not (from memory) a percentage of the image width rather than a pixel value.
Did you mean _now_?

jonhoo:
Ah, yes, sorry, now :sweat_smile:

jonhoo:
I'll close this then since it is not an issue on the inferno side of things. Thanks for the report though!

PR #204, opened 2020-12-28
Increase default flamegraph fidelity

asherkin:
See also #153. PR #136 changed the semantics of the `--minwidth` option from output pixel width to percentage width (which is effectively the same as pct of total samples), which greatly reduced the default fidelity of flamegraphs as 0.1% of total time is quite coarse, especially once you start zooming in.

This PR reduces the default to 0.01%, which produces pretty much identical output as flamegraph.pl / old inferno at ~1200px wide and includes a much greater number of bars. I think this change is worth making as it makes it reduces friction for people migrating from flamegraph.pl, and in my tests produces much more useful output by default.

I've also updated the description of the option to match the current behaviour, as it still mentioned pixels and that had me confused for a good while.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Support jit annotations in collapse-perf for runtimes using the jitdump format. [#202](https://github.com/jonhoo/inferno/pull/202)
 
 ### Changed
+ - Decreased default minimum width from 0.1% to 0.01%. [#204](https://github.com/jonhoo/inferno/pull/204)
 
 ### Removed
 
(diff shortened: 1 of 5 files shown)


asherkin:
`flamegraph.pl --colors annotated --hash`:


`inferno-flamegraph --colors annotated --hash --minwidth 0.01 --width 1200`: (i.e. default after this change)


`inferno-flamegraph --colors annotated --hash --minwidth 0.1 --width 1200` (i.e. default before this)


I do miss the rounded corners, I think they helped pick out the edges of samples quite a bit.

asherkin:
Looks like a bunch of the test output files would need updating due to this -- I'll do that if you're happy taking this change, else just pare this down to the description fix.

jonhoo:
I quite like the change! Rounded corners shouldn't be too hard to add back in with CSS if you want them. I think `border-radius` is all you need?

jonhoo (approved):

PR #205, opened 2021-01-04
Support output to files directly instead of just via output redirection

### LINKED ISSUE #193: Support output to files directly instead of just via output redirection
In my profiles I'm seeing 3.3% of cycles spent in `std::memchr::memchr`, because the standard library's stdout wrapper is trying to detect newlines and flush on them. Which is silly.

jasonrhansen:
Closes #193

Adds `--outfile` option to all of the CLI programs so you can output directly to a file instead of STDOUT.


jonhoo:
Hmm, what's the benefit of this over just having the caller do output redirection?

jasonrhansen:
Based on what @saethlin reported in #193, it looks like it was requested for performance reasons because "the standard library's stdout wrapper is trying to detect newlines and flush on them".

jonhoo:
Ah! My proposal then is rather that we be smarter about how we access stdout. The standard trick here is to detect if STDOUT is a TTY, and if it is, then line-buffer, otherwise use a `BufWriter`. This PR is probably a good starting point for that, since we'll then have to write to a writer directly anyway. That means we can continue to use the primitives the OS provides us with, while still supporting both interactive and batch use-cases.

jasonrhansen:
So do you mean something like this?

```Rust
    if atty::is(atty::Stream::Stdout) {
        flamegraph::from_files(&mut options, &infiles, io::stdout().lock())?;
    } else {
        flamegraph::from_files(&mut options, &infiles, io::BufWriter::new(io::stdout().lock()))?;
    }
```

jasonrhansen:
Opened #206 as a better solution.

PR #206, opened 2021-01-06
Use BufWriter when STDOUT is not a TTY

### LINKED ISSUE #193: Support output to files directly instead of just via output redirection

jasonrhansen:
This is an optimization to avoid line buffering when redirecting output.

Close #193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
  - Decreased default minimum width from 0.1% to 0.01%. [#204](https://github.com/jonhoo/inferno/pull/204)
+ - Detect if STDOUT is a TTY and if it's not, use a `BufWriter` to avoid line buffering. [#206](https://github.com/jonhoo/inferno/pull/206)
 
 ### Removed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ nameattr = ["indexmap"]
 
 [dependencies]
 ahash = "0.6"
+atty = "0.2"
 crossbeam-utils = { version = "0.8", optional = true }
 crossbeam-channel = { version = "0.5", optional = true }
 dashmap = { version = "3", optional = true }
--- a/src/bin/collapse-dtrace.rs
+++ b/src/bin/collapse-dtrace.rs
@@ -83,5 +83,5 @@ fn main() -> io::Result<()> {
     }
 
     let (infile, options) = opt.into_parts();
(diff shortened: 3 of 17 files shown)


jonhoo:
My expectation would actually be that the standard library itself implemented this. And it seems like it's _intended_ to, but it's a `FIXME`:
https://github.com/rust-lang/rust/blob/41601ef39442bca19bc6ac8ef72e5d89335e92d7/library/std/src/io/stdio.rs#L491-L493

At the moment, it seems like there is no way to get a handle to a `STDOUT` that isn't line buffered. That is.. unfortunate. The issue to watch appears to be https://github.com/rust-lang/rust/issues/60673, with the fixing PR being https://github.com/rust-lang/rust/pull/78515. There's also https://github.com/rust-lang/rust/issues/58326 which would be an alternative approach.

Ultimately, this means that once std fixes this behavior, this PR will be unnecessary, and performance will immediately improve. Merging this PR will also not improve performance I don't think. Providing a mechanism specifically for writing to a file _would_ help, though I'm a little hesitant to merge a change that complicates both the internal and external API when it will ultimately become unnecessary following a change to `std`...

jasonrhansen:
It will be really nice when the standard library implements this, but I still believe this PR should help with performance. When you write more than a line to the line writer, it will write the buffer directly to the inner writer. This means if we wrap stdio in a `BufWriter`, we'll be writing larger chunks at a time to the underlying line writer and It will skip the line buffer.

Of course it will still call `memrchr` to find the last newline, but it will do this much less frequently since it won't be doing it for each of our small writes.

https://github.com/rust-lang/rust/blob/master/library/std/src/io/buffered/linewritershim.rs
```Rust
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        let newline_idx = match memchr::memrchr(b'\n', buf) {
            // If there are no new newlines (that is, if this write is less than
            // one line), just do a regular buffered write (which may flush if
            // we exceed the inner buffer's size)
            None => {
                self.flush_if_completed_line()?;
                return self.buffer.write(buf);
            }
            // Otherwise, arrange for the lines to be written directly to the
            // inner writer.
            Some(newline_idx) => newline_idx + 1,
        };
```

Once the standard library is updated to handle this, we can remove this workaround.

jonhoo (approved):

jonhoo:
Yeah, that's a good point. Happy to merge. The CI failure appears to be something broken in tarpaulin, not in our code.

jonhoo:
Oh, actually, before I merge, would you mind updating the changelog?

jonhoo (approved):

Issue #207, opened 2021-01-29
ctrl+f search sometimes reports a very incorrect Matched percentage


saethlin:
I just ran into a seriously egregious case in an internal codebase. I'm zoomed into a function that has a a bug (that I just wrote) that causes it to mostly create and extend `Vec`s. If I ctrl+f for `alloc::alloc`, the flamegraph tells me 35.3% of samples match. That looks quite wrong. But this program uses the default allocator, so I ctrl+f for `__GI___libc` which matches what all the allocator functions call. The flamegraph says 72.3% of samples match, and _looks_ correct.

I have no idea what is going on here. I'm trying to reproduce with something that I can upload. No luck just yet.

saethlin:
Bumping up the value in the `toFixed` in `format_percent` fixes this. Is there a way to get around the formatting and re-parsing of integers inside the SVG? I feel like working with the number of samples everywhere should be more effective, no?

I think this is a rounding problem; the flamegraph I've encountered this on has 22 threads and I'm zoomed into a function that is 0.15% of all cycles (on this thread). But there are 5,934 cycles in this function, so I feel like we shouldn't have rounding issues.

jonhoo:
Hmm, I think removing the percentages might be tricky, though you might be able to inject the sample count using a `data-` attribute which the JavaScript code can then pick up on?

saethlin:
I _think_ this was fixed by #210. But I haven't been fiddling with the really complicated flamegraphs that reproduce this. If I find it again I'll reopen this.

jonhoo:
Ah, yes, I think you're right. Thanks for coming to check back in!

PR #210, opened 2021-04-28
Compute rect position and width from samples

saethlin:
This is my attempt at implementing a solution for https://github.com/jonhoo/inferno/issues/207. Recomputing the percentage position and width every time from the integer sample counts should keep floating-point errors to a minimum. Just that change altered the matched % reported during a search.

But I think the bigger discrepancy was coming from the regex search selecting samples that weren't on screen at all. Since I added the check for hidden objects everything seems quite reasonable.

I'm making this PR because a few days ago I zoomed way in to see the dashmap spinlock loop and the rectangles were hanging over each other. I don't have a reproducer of this on hand to verify, but boy does that seem like floating-point imprecision to me.


src/flamegraph/flamegraph.js
```diff
@@ -330,17 +329,20 @@ function search(term) {
     var maxwidth = 0;
     for (var i = 0; i < el.length; i++) {
         var e = el[i];
+        if (e.classList.contains("hide")) {
```
jonhoo, line 332: This feels like an unrelated change?

src/flamegraph/flamegraph.js
```diff
@@ -198,41 +199,42 @@ function update_text(e) {
...
     }
 }
-function zoom_child(e, x, ratio) {
+function zoom_child(e, x, zoomed_width_samples) {
     if (e.attributes != undefined) {
         if (e.attributes.x != undefined) {
-            orig_save(e, "x");
-            e.attributes.x.value = format_percent((parseFloat(e.attributes.x.value) - x) * ratio);
             if (e.tagName == "text") {
                 e.attributes.x.value = format_percent(parseFloat(find_child(e.parentNode, "rect[x]").attributes.x.value) + (100 * 3 / frames.attributes.width.value));
```
jonhoo, line 218: Should this also be updated to use `x_samples`?

src/flamegraph/flamegraph.js
```diff
@@ -380,11 +382,10 @@ function search(term) {
     // Step through frames saving only the biggest bottom-up frames
     // thanks to the sort order. This relies on the tree property
     // where children are always smaller than their parents.
-    var fudge = 0.0001;    // JavaScript floating point
     for (var k in keys) {
-        var x = parseFloat(keys[k]);
+        var x = parseInt(keys[k]);
         var w = matches[keys[k]];
-        if (x >= lastx + lastw - fudge) {
+        if (x > lastx + lastw) {
```
jonhoo, line 388: I think this should remain `>=`, no?

src/flamegraph/mod.rs
```diff
@@ -866,7 +873,9 @@ fn filled_rectangle<W: Write>(
             "y" => &buffer[y],
             "width" => &buffer[width],
             "height" => &buffer[height],
-            "fill" => &buffer[color]
+            "fill" => &buffer[color],
+            "x_samples" => &buffer[x_samples],
+            "width_samples" => &buffer[width_samples]
```
jonhoo, line 878: I hate to recommend this, but given how many instances of this we may end up generating, I wonder if we should choose shorter labels for these. It could have a meaningful impact of file size for the SVG.

Separately, does SVG generally permit random attributes to be placed on XML elements like this? Judging by [this StackOverflow answer](https://stackoverflow.com/a/20265073/472927) I think we'll either want to add a simple flamegraph namespace (`fg`) or at least name these with the `data-` prefix.

src/flamegraph/flamegraph.js
```diff
@@ -198,41 +199,42 @@ function update_text(e) {
 // zoom
 function zoom_reset(e) {
     if (e.attributes != undefined) {
-        orig_load(e, "x");
-        orig_load(e, "width");
+        if (e.attributes.x_samples != undefined) {
```
jonhoo, line 202: Why would this ever be `undefined`?

jonhoo (changes requested):
This is a great change, thank you for taking the time to go through it! I left some comments inline, but in general the approach looks great.

src/flamegraph/flamegraph.js
saethlin: Could be. I was thinking of this PR as a fix for getting incorrect values from ctrl+f, especially when zoomed in. I'm not sure if this is fixing another bug altogether or if I've misunderstood what ctrl+f is supposed to do, but it's in here because I realized that storing the sample counts isn't sufficient to get the matched % I've been expecting.

jonhoo: Oh, I skimmed over your original PR comment too quickly. Yes, I think this does indeed fix a bug, and happy to take it as part of this change :+1:

src/flamegraph/flamegraph.js
saethlin: I'm not sure.

The percentage values for rectangles which are large enough to have text on them will be great enough that floating-point imprecision isn't significant. I think the code to do the computation would be wordy and not change the behavior.

src/flamegraph/flamegraph.js
saethlin: `zoom_reset` is called on every child of `frames`. It's not just passed the rectangles, but also the text and the parent that wraps the rectangle + text. I just learned this in answering your question, and I've tried to clarify the code.

saethlin:
I just realized that I've broken this code, which attempts to access `_orig_x`: https://github.com/jonhoo/inferno/pull/210/files#diff-57033a9a1ddd2ffa82ff282a58541526d380285fdebecc47fca185ef5414798dR75-R80

Unfortunately I have no idea what it's for. I can't find any functionality that's missing. Help?

src/flamegraph/flamegraph.js
jonhoo: I'm more thinking that we might as well make all the computations consistent while we're at it, but happy to leave it either way.

jonhoo:
Looks like GitHub got confused about your link. I _think_ you're referring to
https://github.com/jonhoo/inferno/blob/3e249a3389c8dad8ade0dc729b7e439f2bdef57a/src/flamegraph/flamegraph.js#L75-L80

My read of that code is that it checks that the clicked rectangle is actually a sample box, and not just some other random `rect` on the page. It uses the check that `_orig_x` is set as a proxy for that. And the preceeding checks are just to avoid errors if it lands on a `rect` that, say, doesn't have any attributes. You could probably just check for `fg:x` instead there.

src/flamegraph/flamegraph.js
```diff
@@ -72,9 +72,9 @@ window.addEventListener("click", function(e) {
 
         // set parameters for zoom state
         var el = target.querySelector("rect");
-        if (el && el.attributes && el.attributes.y && el.attributes._orig_x) {
+        if (el && el.attributes && el.attributes.y && el.attributes["fg:x"]) {
             var params = get_params()
-            params.x = el.attributes._orig_x.value;
+            params.x = el.attributes["fg:x"].value;
```
jonhoo, line 77: Given that this gets saved into the URL (with `replaceState` below), I wonder if we also need to update the code that parses the URL to automatically restore the zoom appropriately (based on sample count rather than x position).

saethlin: I've come up with a commit that fixes the selector. I'm really quite new to JS and all this, and I don't understand why `fg\\:x` and `fg|x` don't work, but `*|x` does. I suppose what I came up with is potentially not the semantics we want, but I can't figure out anything else that works.

jonhoo: Oh, that's bizarre. Doing some digging, I found [6.3.3 Attribute selectors and namespaces in the CSS spec](https://drafts.csswg.org/selectors-3/#attrnmsp), which states that:
> An attribute selector with an attribute name containing a namespace prefix that has not been previously declared is an invalid selector. 

That makes me think that you can't query for `fg|x` since `fg` hasn't been declared in the CSS "world", only in the HTML "world" (through the namespace tag). And with some more digging, it seems like [there isn't a way to specify a CSS namespace from JS](https://stackoverflow.com/a/23047888/472927). So, your solution is the only one it appears — just match independent of the namespace. Weird stuff.

If you've tested this and it works, then I'm happy with it! Should just be a matter of zooming through a bunch of different elements, and then using your browser back and forward arrows and seeing that the zoom changes as expected.

saethlin: > If you've tested this and it works, then I'm happy with it! Should just be a matter of zooming through a bunch of different elements, and then using your browser back and forward arrows and seeing that the zoom changes as expected.

I'm not sure forward and back have ever worked. As far as I can tell, they do not currently work on the latest-released version. What I broke then fixed with the latest commit was the ability to copy a zoomed URI to a new browser window and get the correct zoomed-in view upon pasting it.

jonhoo: Ah, I see — I think that's because we're using `replaceState` rather than `pushState`, but that's not something you have to fix here given that it's already broken!

jonhoo (approved):
Will release this either today or tomorrow, thanks! 👍

jonhoo:
Released in 0.10.4

PR #211, opened 2021-05-01
Add Rust language coloring

zthompson47:
This PR is part of an effort to provide more coloring options for `cargo-flamegraph` in flamegraph-rs/flamegraph#19.

It looks for Rust functions in the stack (i.e. contains "::"), and applies separate color palettes for Rust system functions and user functions.


src/flamegraph/color/palettes.rs
```diff
@@ -99,6 +99,25 @@ pub(super) mod wakeup {
     }
 }
 
+pub(super) mod rust {
+    use crate::flamegraph::color::BasicPalette;
+
+    // Partition into user code and Rust system code
+    pub fn resolve(name: &str) -> BasicPalette {
+        if name.starts_with("core::") || name.starts_with("std::") {
```
jonhoo, line 107: Let's include `alloc::` here too.

src/flamegraph/color/palettes.rs
```diff
@@ -99,6 +99,25 @@ pub(super) mod wakeup {
...
+            return BasicPalette::Orange;
+        } else if name.contains("as core::")
+            || name.contains("as std::")
+            || name.contains("as alloc::")
```
jonhoo, line 111: Do these show up in practice in debug symbols? If so, I think I'd prefer more robust parsing here, as I feel like this parsing might end up yielding false positives down the line. Can you narrow down exactly what kind of items we expect to hit this case?

src/flamegraph/color/palettes.rs
```diff
@@ -99,6 +99,25 @@ pub(super) mod wakeup {
...
+        {
+            return BasicPalette::Orange;
+        } else if name.contains("::") {
+            return BasicPalette::Yellow;
```
jonhoo, line 115: What case is this for? That is, what's the rationale for coloring these differently than things without `::`?

src/flamegraph/color/palettes.rs
```diff
@@ -99,6 +99,25 @@ pub(super) mod wakeup {
...
+        }
+        BasicPalette::Red
```
jonhoo, line 117: I wonder if we want a different color here since the base color is `Orange`, just to improve the contrast. Might be helpful to take a look at an actual rendering of the resulting SVG for comparison.

jonhoo (changes requested):
This is a great addition, thank you! I left some notes inline.

src/flamegraph/color/palettes.rs
zthompson47: Yes, I saw things like the following in a local test:
```
<alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
<core::iter::adapters::enumerate::Enumerate<I> as core::iter::traits:iterator::Iterator>::fold
```
I'll search around for a more definitive answer, but these should probably match on something like `name.starts_with("<alloc::")`, etc.

src/flamegraph/color/palettes.rs
zthompson47: Yes, good point, I'll add that in.

src/flamegraph/color/palettes.rs
zthompson47: Good point - I'll try something with more contrast.

src/flamegraph/color/palettes.rs
zthompson47: That's a catch-all for any rust function that isn't part of `core`, `alloc`, or `std`.  So I'm assuming it's a function from the user's code being tested.  I see now that it would also match on functions from dependency crates, but generally I see it as a way for a user to locate code that they wrote in the flamegraph.

Is there a better way to identify Rust functions?  From scanning my test flamegraphs, I don't see any "::" in the lower-level functions (e.g. `cpal_alsa_in`, `start_thread`), so it's been working on my limited samples.  Doesn't seem robust, though.

Other options include using `regex` patterns, and possibly even parsing `Cargo.toml` to find crate names to match on, so dependencies could look different from project crates.

zthompson47:
Thanks for the review and your time!  I can start another revision tomorrow morning or so, and I left some responses up above.

src/flamegraph/color/palettes.rs
jonhoo: Yeah, I think `<alloc::` and `<core::` is probably a better match here then :+1:

src/flamegraph/color/palettes.rs
jonhoo: Ohh, I read this entirely backwards then. `::` is "any Rust function", and the _else_ case is for libc/C/syscall type stuff. I totally missed that. Would be good to call that out with a comment I think.

As for whether it's right, I think it depends on things like what language you're interfacing _with_. For example, if Rust calls into C++ code (or vice-versa), a bunch of C++ methods would also be colored with the "Rust color" here. But maybe that's okay. I guess what I was getting at with my question was more "should we just make anything that _isn't_ core/alloc/std have the same color?".

src/flamegraph/color/palettes.rs
zthompson47: Sounds good.  I found [the name-mangling rfc](https://github.com/rust-lang/rfcs/blob/master/text/2603-rust-symbol-name-mangling-v0.md) where it says, "The RFC therefore proposes to keep symbol names close to how methods are represented in error messages".  My naive interpretation is that what we have here should work..

src/flamegraph/color/palettes.rs
zthompson47: I feel like it would be useful to partition the Rust functions into user/system, mostly so it's easy for a user to find their own code in the flamegraph.  I don't have the experience to make the call, but it seems like these false positives would be okay.  I think code from different languages would appear grouped in distinct regions of the flamegraph, so it wouldn't reduce readability.

I'll have time later this evening to test more examples.  I'm currently just using a pure Rust example for testing.

src/flamegraph/color/palettes.rs
zthompson47: Here's the nicest color combo I've found so far.  I'm starting with the idea that the Rust code has to be orange so it looks "oxidized".  :)

flamegraph

zthompson47:
I pushed another commit.  I hope that's the right way to go at this point, but I thought it would be most useful to have my latest code up here for review.  I'm still kind of new to the collaboration end of this process so I'm hoping I didn't confuse things by introducing new changes at this stage.

src/flamegraph/color/palettes.rs
jonhoo: Right, but which "Rust" code is more clearly Rust, the user's code or the std/core/alloc code? :p
I think what I'd do is make std/core/alloc red, make user code (`::`) orange, and make everything else either green or yellow. I think that might give a nice "all of this is Rust-related, and these bits are not".

jonhoo:
Not at all — the right way to go is to continuously improve the PR incrementally until we get to it being ready to land :)

src/flamegraph/color/palettes.rs
jonhoo: Yeah, that makes sense — I buy it.

src/flamegraph/color/palettes.rs
zthompson47: It's fun playing with these combos.  Here's the latest:
flamegraph

jonhoo: I worry that with the brightness adjustments, red and orange are just too hard to distinguish 😢 Maybe your instinct to use `Yellow` was better in the first place. I'm honestly torn here!

Also, just after suggesting it, I realized that red+green is a bad choice, since it's the most common form of color-blindness. We should probably use basically any color other than green for the non-Rust things. Like... blue. I guess that brings us back to your original proposal 😅 

This is very much bikeshedding at this point. But the fun kind! My overall aims with the colors are:
 - Make user code very visible.
 - Make std code distinguishable from user code.
 - Make user and std colors mesh well, since they're likely to be interspersed (like in your posting).
 - Make non-Rust frames highly distinguishable from Rust frames.
 - Make non-Rust frames more muted than Rust frames, since we're trying to _highlight_ Rust code with this palette.

Working backwards from that, how about:
 - Rust user: Aqua
 - Rust std: Orange
 - Non-Rust: Yellow

zthompson47: Heh, I hear you on all that, and I like this latest palette.  The aqua pops right out but matches well with the orange (it's complementary color).  The yellow stands back a bit but also pairs with it's nearby orange to anchor down the non-user lower-level functions (rust std + non-rust).

As an aside, I'm also grappling with a rather short experience with flamegraphs so far, so I'm glad you wanted to explore these choices a bit more.  Also, I'm realizing that I've been favoring certain color palettes because of the font color contrast (the red is a bit hard for me), but I'm drifting off-topic, and I might have another idea about that for another day..

Here's another visual:
flamegraph-aqua

jonhoo: This looks great! I'm not a huge fan of the "grenness" of the yellow color, but that's a problem for another day. And so is font contrast :)

jonhoo:
This looks great, now we just need to get the tests to pass!

jonhoo (dismissed):

zthompson47:
Excellent!  I just tried to merge with upstream/master, and it looks like there are still errors in the tests.

Ok, just realized that _my_ test was failing and fixed it.

FWIW, there were some already-broken tests on my system before this feature which I attributed to not have the right profiling software installed on my end.

jonhoo (approved):

jonhoo:
Looks great, thanks!

jonhoo:
Released in 0.10.5 :tada:

zthompson47:
Thanks so much for your help on this!

jonhoo:
Thanks for sticking with it!

PR #212, opened 2021-05-22
Improve differential flame graph by delta width

unpluggedcoder:
Hi, all: 

I change the way generate differential flame graph by using delta as frame's width, and wrote an article to explain the new diff-graph, you can see it below:
https://unpluggedcoder.me/2021/05/22/Improved%20Differential%20Flame%20Graph/

⚠️NOTICE: This PR may have a big impact on current differential flame graph.

And round the rect of frame for better view.

### DIFF
--- a/src/flamegraph/merge.rs
+++ b/src/flamegraph/merge.rs
@@ -16,11 +16,13 @@ pub(super) struct TimedFrame<'a> {
     pub(super) start_time: usize,
     pub(super) end_time: usize,
     pub(super) delta: Option<isize>,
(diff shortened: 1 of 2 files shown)


jonhoo:
Hi! Thanks for this, but I'm going to close this as even though the new type of diff you propose exposes some useful information compared to the existing one, it also reduces other useful information. So, between those, I'd prefer to stick with something that's close to what the upstream flamegraph project has, rather than inferno having a different, custom implementation.

unpluggedcoder:
👌

PR #213, opened 2021-05-25
Color user-defined async functions correctly for Rust coloring

dmarriner:
The current Rust coloring is a little confusing because it colors desugared user-defined async functions in the same color as system functions.  This change will cause the Rust coloring to use the user coloring instead of the system coloring for those functions.

Before:

before

After:

after


jonhoo (approved):
This looks great, thank you!

jonhoo:
Released in 0.10.6 :tada:

PR #216, opened 2021-08-31
Update dependencies

djc:
While improving the flamegraph-rs repo setup I noticed that quite a few of inferno's dependencies are outdated. From quickly skimming the docs, I don't think any of these affect the public API, so this seems fairly straightforward.

The quick-xml upgrade changed the output for the test cases slightly
(omitting whitespace between the <script> tags and CDATA start/end),
so I solved that by updating the test data with nicer formatting.

I wrote a Python script to automatically rewrite test data for failing tests, I can submit that too if you like.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,25 +27,25 @@ multithreaded = ["dashmap", "crossbeam-utils", "crossbeam-channel", "num_cpus"]
 nameattr = ["indexmap"]
 
 [dependencies]
-ahash = "0.6"
+ahash = "0.7"
 atty = "0.2"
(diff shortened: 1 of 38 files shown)


djc:
This turned out to require an MSRV bump. Given that the new MSRV is 1.46, which was released a year ago, that doesn't seem like a big deal?

jonhoo (approved):
Seems entirely reasonable — thanks!

jonhoo:
Published as 0.10.7 :tada:

djc:
Thanks!

PR #218, opened 2021-09-15
Add collapser for built in Visual Studio profiler

### LINKED ISSUE #217: Add collapser for the Visual Studio built in profiler
Recently, I wanted to create a FlameGraph of the output of the Visual Studio profiler, but as it turns out, there is no stackcollapser for that  yet. Since the original project does have a [stackcollapser](https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-vsprof.pl) for this profiler, I was wondering if this would be a welcome contribution? I implemented my own stackcollapser before I found out there was an original, so I'd need to check that I haven't missed anything, but besides some polishing (and some tests) I think it should be pretty much good to go.

KnapSac:
This adds a stack collapser for the built in Visual Studio profiler.

Closes #217


src/collapse/vsprof.rs
```diff
@@ -0,0 +1,198 @@
...
+
+impl Collapse for Folder {
+    fn collapse<R, W>(&mut self, mut reader: R, writer: W) -> io::Result<()>
+    where
+        R: std::io::BufRead,
+        W: std::io::Write,
+    {
+        // Skip the header
+        let mut line = Vec::new();
+        if reader.read_until(0x0A, &mut line)? == 0 {
```
jonhoo, line 27:
```suggestion
        if reader.read_until(b'\n', &mut line)? == 0 {
```
;)

KnapSac:
@jonhoo I'm not entirely sure what to do about the code coverage test failing, looking at the diff it seems to me as if they're erroneously marking some lines as not covered. Do you have any ideas?

I only tested the output of a C# program, as I don't have any C++ projects to profile. Additionally, I don't know what other languages the built in Visual Studio profiler supports, which there are probably quite a few of. I suspect that the output for all of them is the same, and as such they shouldn't have any issues with this implementation.

In C# the namespace is included in the function name output (so the function name is actually the fully qualified name of the function), this makes some function names incredibly long, especially if the function also has several arguments with long fully qualified names. One idea I had to make the output more readable was to introduce an option to specify the number of namespaces which should be kept and trim any extra namespaces. Another idea could be to add an option to keep the fewest namespaces possible, while still being unique. Both might also not be a bad idea. Is this something you'd like to discuss/incorporate into this PR, or should a discussion about that be moved to a new issue? The latter would have my preference, as this is, although related, not a requirement for a working implementation. What are your thoughts on this?

jonhoo:
I wouldn't worry too much about the code coverage — it tends to be overly sensitive. Usually it's because error paths aren't covered, but I'm okay with that for a first PR :)

I _did_ notice that the [output `svg` in `tests/data/collapse-vsprof/results/sample-default.svg`](https://github.com/jonhoo/inferno/blob/73d54d6daa8a0e843a478f312af55980acf88dc1/tests/data/collapse-vsprof/results/sample-default.svg) doesn't appear to be a valid SVG file though, which doesn't seem right?

I'm also not entirely sure why you chose to not support lines with a depth deeper than 1000? It seems fairly simple to parse the contained numbers? Just find the next `"`, strip all `,`, and then parse as an integer?

For namespaces, I agree with you that such trimming functionality is something we can discuss separately. No need to implement that here, even if it may be a good idea to add later!

I'm a bit short on time, so may take a bit until I have a chance to give this a thorough review, but hopefully the above unblocks you for a little while at least. Thanks for taking the time to implement this!

KnapSac:
The svg output indeed does not appear to be correct, when I have some time later this week I'll look into it. As to not supporting lines with a depth over 1000, I don't have any good reason for that. On second thought, whether or not it is actually possible to have a call stack over 1000 deep, we shouldn't introduce an arbitrary restriction on that. I'll also fix that.

For now I'll mark this PR as a draft, when I've resolved these issues I'll mark it as ready. Don't worry about being short on time, as a student I know the feeling ;).

src/collapse/mod.rs
```diff
@@ -35,6 +35,13 @@ pub mod sample;
 ///   [crate-level documentation]: ../../index.html
 pub mod vtune;
 
+/// Stack collapsing for the output of the built in [`Visual Studio profiler`](https://docs.microsoft.com/en-us/visualstudio/profiling/profiling-feature-tour?view=vs-2019).
```
jonhoo, line 38: This is a nit, but don't want to suggest that the Visual Studio profiler is built in to _this_ project :p
```suggestion
/// Stack collapsing for the output of the [Visual Studio built-in profiler](https://docs.microsoft.com/en-us/visualstudio/profiling/profiling-feature-tour?view=vs-2019).
```

jonhoo: Also note that the backticks aren't needed for links. They're there in the other backends because those are intended to represent commands, which are typically typeset in `monospace` (which is what the backticks do).

tests/collapse-vsprof.rs
```diff
@@ -0,0 +1,118 @@
+mod common;
```
jonhoo, line 1: Just as a sanity check — can you verify that you actually get a reasonable SVG at the end as well when you run the flamegraph CLI locally with this change on the `sample-default.txt` file? We should really have an end-to-end test from stack-collapse-input to flamegraph SVG for _every_ collapser, but that's not something that's currently in there. Would be happy for you to add the first (probably to `tests/flamegraph.rs`), but it's not a requirement. I just at least want a manual check that you get a good SVG.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+static START_LINE: &str = "Level,Function Name,Number of Calls,Elapsed Inclusive Time %,Elapsed Exclusive Time %,Avg Elapsed Inclusive Time,Avg Elapsed Exclusive Time,Module Name,";
+
+/// A stack collapser for the output of the built in Visual Studio profiler.
+#[derive(Default)]
+pub struct Folder {
+    /// Function entries on the stack in this entry thus far.
+    stack: Vec<(String, usize)>,
+}
+
+impl Collapse for Folder {
```
jonhoo, line 19: If you want a slight challenge (definitely not required for the PR to land!), try implementing the `CollapsePrivate` trait instead:
https://github.com/jonhoo/inferno/blob/410018b46983b1ee732f8e6f7f94f843b9de09d5/src/collapse/common.rs#L53-L59

That will give you multi-core file processing, which can speed up collapsing quite a lot.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+    fn collapse<R, W>(&mut self, mut reader: R, writer: W) -> io::Result<()>
+    where
+        R: std::io::BufRead,
+        W: std::io::Write,
+    {
+        // Skip the header
```
jonhoo, line 25: This doesn't actually check that `line == START_LINE`. I feel like I'm missing something?

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+                return Some(false);
+            }
+        } else {
+            return Some(false);
+        }
+
+        // Some files may start with the <U+FEFF> character (zero width no-break space). This
+        // causes the call to `starts_with` to return false, which in this case isn't what we want.
+        // As this character has no influence on the rest of the file, we can safely ignore it.
+        if line.starts_with(START_LINE) || line.starts_with(&("\u{feff}".to_owned() + START_LINE)) {
```
jonhoo, line 70: This saves you the extra string allocation, and I think reads better to:
```suggestion
        if line.trim_start_matches('\u{feff}') == START_LINE {
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+/// Gets the number from the start of the line. This can either be a number <1000, in which case the
+/// line doesn't contain double quotes, or the number can be >1000, in which case the line does
+/// contain double quotes. In both cases `line` may start with a leading comma, which will be
+/// ignored.
+fn get_next_number(line: &str) -> io::Result<(usize, &str)> {
+    // Trim the leading comma, if any
+    let line = line.trim_start_matches(',');
```
jonhoo, line 158: Up to you which one you prefer. This one will only remove _one_ comma.
```suggestion
    let line = line.strip_prefix(',').unwrap_or(line);
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    if let Some('"') = line.chars().next() {
```
jonhoo, line 160:
```suggestion
    if let Some(rest) = line.strip_prefix('"') {
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+        // number >1000
+        // Trim the leading double quote and then split on the next double quote, which leaves
+        // the number as the first part of the result.
+        let mut split = line.trim_start_matches('"').splitn(2, '"');
```
jonhoo, line 164:
```suggestion
        let split = rest.split_once('"');
```
Crucially, this will return `None` if there is no second `"` in the string, whereas `.splitn` will return the remainder of the string if it contains no `"`.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+        if let Some(num) = split.next() {
```
jonhoo, line 165:
```suggestion
        if let Some((num, rest)) = split {
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+            if !(num.contains('.') && num.contains(',')) {
+                // Remove both commas and periods from the number, this means we can parse the
+                // number culture independently.
+                if let Ok(num) = num.replace(",", "").replace(".", "").parse::<usize>() {
+                    return Ok((num, split.next().unwrap_or("")));
+                }
+            }
+        }
+    } else {
+        let mut split = line.splitn(2, ',');
```
jonhoo, line 178: Here too I'd use `split_once`.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+    if let Some('"') = line.chars().next() {
+        // number >1000
+        // Trim the leading double quote and then split on the next double quote, which leaves
+        // the number as the first part of the result.
+        let mut split = line.trim_start_matches('"').splitn(2, '"');
+        if let Some(num) = split.next() {
+            // If `num` contains both a comma and a period, it is definitely a fractional
+            // number, regardless of the current culture. As a fractional number of calls is
+            // not possible, we don't accept them and fallback to the error returned below.
+            if !(num.contains('.') && num.contains(',')) {
```
jonhoo, line 169: This condition confuses me. A number that contains _neither_ is accepted here, as is a number like `"100.53"`, which doesn't seem right? Is `,` _always_ used as the thousands separator in this file format, or only sometimes?

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+                // Remove both commas and periods from the number, this means we can parse the
+                // number culture independently.
+                if let Ok(num) = num.replace(",", "").replace(".", "").parse::<usize>() {
+                    return Ok((num, split.next().unwrap_or("")));
```
jonhoo, line 173:
```suggestion
                    return Ok((num, rest));
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+impl Folder {
+    // Parse lines like:
+    // Level,Function Name,Number of Calls,...
+    // 6,"System.String.IsNullOrEmpty(string)",4,0.00,0.00,0.00,0.00,"mscorlib.dll",
+    fn on_line(&mut self, line: &str, occurences: &mut Occurrences) -> io::Result<()> {
+        let (depth, remainder) = get_next_number(line)?;
+        // After the depth is parsed from `line`, it still has a leading comma and double
+        // quote. By trimming those, we know that the first double quote is the closing double
```
jonhoo, line 85: This assumes that the function name is _always_ in double quotes. Is that the case? I could imagine that maybe it isn't if it doesn't contain spaces or something, but don't know for sure?

If it is the case, I'd use `.strip_prefix(",\"")` here over `trim_start_matches`, since the latter will remove multiple occurrences.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+        // quote surrounding the function name. By splitting on this double quote, we get the
+        // function name and the remainder of the line.
+        let x: &[_] = &[',', '"'];
+        let mut split = remainder.trim_start_matches(x).splitn(2, '"');
```
jonhoo, line 89: Here, too, `split_once`. It also simplifies your `match` below slightly, and you can just make it
```rust
if let Some((function_name, remainder)) = split
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+        match (split.next(), split.next()) {
+            (Some(function_name), Some(remainder)) => {
+                let (number_of_calls, _) = get_next_number(remainder)?;
+
+                let prev_depth = self.stack.len();
+                // There are 3 separate cases to handle regarding the depth:
```
jonhoo, line 95: Love this!

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+                //    outer function
+                // 2. prev_depth == depth -> the previous function call was a leaf node, so we
+                //    need to save the current stack and replace the top node with our node
+                //    call
+                // 3. prev_depth > depth -> the previous function call was a leaf node, so we
+                //    need to save the current stack and than we need to pop the top nodes
+                //    until the top node is our parent
+                match prev_depth.cmp(&depth) {
+                    // Case 1
+                    Ordering::Less => {
```
jonhoo, line 107: We'll hit this if `prev_depth` is, say, `depth - 2` as well. Can that happen? Should it be handled differently?

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+                    }
+                    // Case 3
+                    Ordering::Greater => {
+                        self.write_stack(occurences);
+
+                        // Pop (prev_depth - depth - 1) nodes of the stack, this leaves our
+                        // parent as the top node.
+                        for _ in 0..=prev_depth - depth {
+                            self.stack.pop();
+                        }
```
jonhoo, line 126: I would probably use [`Vec::truncate`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.truncate) here instead.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+            _ => {
+                return invalid_data_error!("Unable to parse function name from line:\n{}", line);
+            }
+        }
+
+        Ok(())
+    }
+
+    // Store the current stack in `occurences`
+    fn write_stack(&self, occurences: &mut Occurrences) {
```
jonhoo, line 142:
```suggestion
    fn write_stack(&self, occurrences: &mut Occurrences) {
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+        let (functions, samples): (Vec<_>, Vec<_>) = self.stack.iter().cloned().unzip();
+        if let Some(number_of_calls) = samples.last() {
+            if *number_of_calls > 0 {
+                occurences.insert(functions.join(";"), *number_of_calls);
+            }
+        }
```
jonhoo, line 148:
```suggestion
        if let Some(nsamples) = self.stack.last().map(|(_, n)| *n) {
            if nsamples > 0 {
                let functions: Vec<_> = self.stack.iter().cloned().map(|(f, _)| f).collect();
                occurences.insert(functions.join(";"), nsamples);
            }
        }
```
This avoids doing the `unzip` work if there are no samples, and avoids allocating a `Vec` for samples, which isn't really needed.

Technically the `functions` `Vec` is wasted, since we're ultimately just collecting into a `String`, but unless you specifically want to I wouldn't worry about optimizing that :)

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,188 @@
...
+                        self.stack.pop();
+                        self.stack
+                            .push((function_name.to_string(), number_of_calls));
+                    }
+                    // Case 3
+                    Ordering::Greater => {
+                        self.write_stack(occurences);
+
+                        // Pop (prev_depth - depth - 1) nodes of the stack, this leaves our
+                        // parent as the top node.
```
jonhoo, line 123: Something seems off here. Let's say the stack depth goes from 5 to 3. Presumably I should register the sample counts for both the node at 5 and the one at 4? But I think the single call to `write_stack` above will _just_ register the sample count for the stack at depth 5, but not the one at 4, which then gets popped (and thus forgotten) just below?

jonhoo (changes requested):
Excellent! Left some more comments :)

src/collapse/vsprof.rs
KnapSac: I was using `split_once` originally, which was much more convenient, but it has only been stabilized in `1.52.0`, so this made the CI check for compilation on `1.46.0` fail. Unless the minimum compiler version for this project is changed to `1.52.0`, I think this isn't possible.

jonhoo: I'm okay with bumping to 1.52.0. It was released May 6th, which is now 5 months ago. Assuming the next release of inferno with this change goes out in, say, a month, that's a 6 month window, which I'm comfortable with.

KnapSac: Should I bump the version in this PR, or would you prefer a separate PR for that? When running clippy I get quite a few warnings (in other files), which may or may not be new since 1.46.0. With a separate PR I feel like that would present a good opportunity to include fixes for those warnings.

jonhoo:
Given that we've now bumped the MSRV (in https://github.com/jonhoo/inferno/pull/227), you can use `split_once` again!

src/collapse/vsprof.rs
KnapSac: I checked, the function name is always in double quotes, regardless of the presence of any spaces.

src/collapse/vsprof.rs
KnapSac: What I was trying to do here (in maybe not the best way possible), was to disallow fractional numbers, but allow numbers with thousands separators, regardless of the current culture. I have looked, but I'm unable to find whether VS always outputs with `,`s as thousands separators, or whether they'll use `.`s if that is required by the current culture. They do seem to always use thousands separators if the number is >1000.

Right now I'm only removing the `,`s, because apparently Rust doesn't support them during parsing (probably has to do with culture stuff). Do you know of a more idiomatic way of handling numbers which contain thousands separators, regardless of culture? Or are you fine with just assuming that VS always uses `,`s?

src/collapse/vsprof.rs
KnapSac: As far as I can tell, this can never happen.

tests/collapse-vsprof.rs
KnapSac: This produces a valid SVG.

KnapSac:
Wow, this has taken me a lot longer to get back to than I would have liked, my apologies.

I believe I have addressed all the review comments you left during your first review, and all my tests pass. The produced flamegraphs also look correct to me visually (I just created a simple program with only a few functions to make it easier to check correctness visually).

Since this branch has fallen behind `master` quite a bit, it's probably wise to update it. I feel like the PR would benefit from rebasing onto `master` and then squashing most of the commits, but I just wanted to know whether you're ok with that, or whether you'd rather I merge with `master`. Especially the commits regarding the MSRV change create a bit of unnecessary noise, which I'd like to get rid of.

Updating will also fix the build I believe.

jonhoo:
All good, thanks for coming back to it! Yes, please do squash and rebase onto master and I'll do another review :+1:

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+        // Skip the header
+        let mut line = Vec::new();
+        if reader.read_until(b'\n', &mut line)? == 0 {
+            warn!("File ended before start of call graph");
+            return Ok(());
+        };
+
+        let header = String::from_utf8_lossy(&line).to_string();
+        if !line_matches_start_line(header.clone()) {
+            return invalid_data_error!("Incorrect header:\n{}", header);
```
jonhoo, line 34: Let's be a little more helpful here. How about
```suggestion
            return invalid_data_error!("Expected first line to be header line\n    {}\nbut instead got\n    {}", START_LINE, header);
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn is_applicable(&mut self, input: &str) -> Option<bool> {
+        let mut input = input.as_bytes();
+        let mut line = String::new();
+        if let Ok(n) = input.read_line(&mut line) {
+            if n == 0 {
+                return Some(false);
+            }
+        } else {
+            return Some(false);
+        }
```
jonhoo, line 72: I think you can do this instead:
```suggestion
        let line = input.lines().next().expect("there is always at least one line (even if empty)");
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+impl Folder {
+    // Parse lines like:
+    // Level,Function Name,Number of Calls,...
+    // 6,"System.String.IsNullOrEmpty(string)",4,0.00,0.00,0.00,0.00,"mscorlib.dll",
+    fn on_line(&mut self, line: &str, occurences: &mut Occurrences) -> io::Result<()> {
+        let (depth, remainder) = get_next_number(line)?;
+
+        // Function names are always wrapped in spaces. By trimming the leading double quote, we
```
jonhoo, line 85: Do you mean "wrapped in quotes" here?

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+        // know that the next double quote is the double quote closing the function name. Splitting
```
jonhoo, line 86: Heh, that's unfortunately only true _if_ the generator doesn't ever generate `\"` with the expectation that it _doesn't_ end the string. But I'm guessing it's exceedingly rare that a function name has a `"` in it.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+                occurrences.insert(functions.join(";"), nsamples);
+            }
+        }
+    }
+}
+
+/// Gets the number from the start of the line. This can either be a number <1000, in which case the
+/// line doesn't contain double quotes, or the number can be >1000, in which case the line does
+/// contain double quotes. In both cases `line` may start with a leading comma, which will be
+/// ignored.
```
jonhoo, line 181: Can you give some example inputs here? I also think this function would be very worth writing some simple unit tests for with the kinds of inputs you're expecting it to handle.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+fn get_next_number(line: &str) -> io::Result<(usize, &str)> {
+    // Trim the leading comma, if any
+    let line = line.strip_prefix(',').unwrap_or(line);
+
+    let mut remove_leading_comma = false;
+    let split = if let Some(line) = line.strip_prefix('"') {
```
jonhoo, line 187:
```suggestion
    let field = if let Some(line) = line.strip_prefix('"') {
```
This would make the code easier to read for me at least, since at first I couldn't figure out why you were _splitting_ on `"`. But it makes sense given that `line` is actually a _record_ that has many fields in it.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+        line.split_once('"')
+    } else {
+        line.split_once(',')
+    };
+
+    if let Some((num, remainder)) = split {
+        // Remove any thousands separators which may be present
+        // TODO: If Visual Studio were to output numbers with a `.` as thousands separators, this
+        //       would fail
+        let num = num.replace(',', "");
```
jonhoo, line 198: This will also allocate a string, which seems sad. Here's a counter-proposal:
```rust
fn parse(field: &str) -> std::io::Result<usize> {
    let thousands = field.split(|c: char| !c.is_ascii_digit());
    let mut n = 0;
    for thousand in thousands {
        // return invalid data error here instead if you prefer
        if n == 0 {
            assert!(thousand.len() <= 3);
        } else {
            assert_eq!(thousand.len(), 3);
        }
        n *= 1000;
        n += thousand.parse::<usize>()?;
    }
    Ok(n)
}
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    invalid_data_error!("Invalid number in line:\n{}", line)
+}
+
+/// Some files may start with the <U+FEFF> character (zero width no-break space). This
+/// causes the call to `starts_with` to return false, which in this case isn't what we want.
+/// As this character has no influence on the rest of the file, we can safely ignore it.
+fn line_matches_start_line(line: String) -> bool {
+    line.trim().trim_start_matches('\u{feff}') == START_LINE
```
jonhoo, line 221:
```suggestion
    line.trim().trim_start_matches('\u{feff}').starts_with(START_LINE)
```
To handle the case where new fields are added in the future?

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+            //    store the function name and the number of times it is called from the
+            //    outer function
+            // 2. prev_depth == depth -> the previous function call was a leaf node, so we
+            //    need to save the current stack and replace the top node with our node
+            //    call
+            // 3. prev_depth > depth -> the previous function call was a leaf node, so we
+            //    need to save the current stack and than we need to pop the top nodes
+            //    until the top node is our parent (i.e. the function which called us)
+            match prev_depth.cmp(&depth) {
+                // Case 1
```
jonhoo, line 109: Should we sanity check that the depth is only greater _by 1_?

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+                    // that `A()` only calls `B()` less than 50% of the time, while in fact it is
+                    // called more than 50% of the time.
+                    //
+                    // To prevent this from happening, we instead subtract the number of calls from
+                    // the previous node (in this case `B()`) from the current node (in this case
+                    // `A()`. This leaves `A()` with the correct number of 500 samples.
+                    //
+                    // If the previous number of calls is equal to the current number of calls, we
+                    // don't want to write the current top node, because that would duplicate the
+                    // number of samples for the current node.
```
jonhoo, line 137: I feel like I'm missing something here. If A is called 500 times, and B is called 500 times _from A_, then won't this result in B not getting included in the flamegraph at all?

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+            return invalid_data_error!("Unable to parse function name from line:\n{}", line);
+        }
+
+        Ok(())
+    }
+
+    // Store the current stack in `occurences`
+    fn write_stack(&self, occurrences: &mut Occurrences) {
+        if let Some(nsamples) = self.stack.last().map(|(_, n)| *n) {
+            if nsamples > 0 {
```
jonhoo, line 170: If you want you can simplify this somewhat:
```suggestion
        if let Some(nsamples) = self.stack.last().map(|(_, n)| *n).filter(|n| n > 0) {
```

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+                let functions: Vec<_> = self.stack.iter().cloned().map(|(f, _)| f).collect();
```
jonhoo, line 171:
```suggestion
                let functions: Vec<_> = self.stack.iter().map(|(f, _)| f).cloned().collect();
```
No need to also clone the counts. If that fails to compile you can try:
```rust
                let functions: Vec<_> = self.stack.iter().map(|&(ref f, _)| f).cloned().collect();
```

jonhoo: Actually, given that _just_ use this to `join` below, I don't think you need the `.cloned()`, since the `join` should work fine with `&str`. You _may_ need to map with `&*` or something to go from `&String` to `&str` though.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        Some(line_matches_start_line(line))
+    }
+}
+
+impl Folder {
+    // Parse lines like:
+    // Level,Function Name,Number of Calls,...
+    // 6,"System.String.IsNullOrEmpty(string)",4,0.00,0.00,0.00,0.00,"mscorlib.dll",
```
jonhoo, line 81: What are the rest of the fields?

jonhoo (changes requested):
Phew, finally found some time to review! Thanks again for coming back to this. I've left some more notes inline, though I think they're all fairly minor :+1:

src/collapse/vsprof.rs
KnapSac: In C# a function name cannot have a `"` inside it, and I doubt that there are many (if any) languages which do allow them.

According to the [docs], the Visual Studio profiler only supports `.NET`, `C/C++` and `Javascript`. None of these languages support `'"`s in identifiers, as far as I know, so I feel like it is safe to assume that any `"` encountered is not part of the function name itself.

[docs]: https://docs.microsoft.com/en-us/windows/apps/performance/choose-between-tools

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,222 @@
+use std::{
+    cmp::Ordering,
+    io::{self, BufRead},
+};
+
+use log::warn;
+
+use crate::collapse::{common::Occurrences, Collapse};
+
+static START_LINE: &str = "Level,Function Name,Number of Calls,Elapsed Inclusive Time %,Elapsed Exclusive Time %,Avg Elapsed Inclusive Time,Avg Elapsed Exclusive Time,Module Name,";
```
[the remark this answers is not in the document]
KnapSac: I looked around, but I could not find anything indicating that the order/presence of these fields can be changed.

src/collapse/vsprof.rs
KnapSac: I could not get this to compile:
```rust
let function: Vec<_> = self.stack.iter().map(|&(ref f, _)| &*f).collect();
```

But this does compile, without needing `.cloned()`:
```rust
let function: Vec<_> = self.stack.iter().map(|(f, _)| &f[..]).collect();
```

src/collapse/vsprof.rs
KnapSac: Ah, these are simply the fields specified in the header. I've updated the comment to reflect that.

src/collapse/vsprof.rs
KnapSac: I tested this, and it seems to work. If I call A 500 times, and B 500 times *from* A (so B actually is called 500 * 500 = 250.000 times), B still shows up on the output. B also has 250.000 samples, so that also seems to be correct.

jonhoo: Ah, no, I meant the case where every call to A is also calls B (once each)

KnapSac: I checked, this also works.

The reason this works is that `prev_number_of_calls` is initialized to 0, and the top node will never have 0 calls, because than it would not be present in the output. This makes the if statement on line 137 always return true, so the initial top node is always written.

I have updated the comment, hopefully this clears up your confusion.

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,288 @@
...
+fn get_next_number(line: &str) -> io::Result<(usize, &str)> {
+    // Trim the leading comma, if any
+    let line = line.strip_prefix(',').unwrap_or(line);
+
+    let mut remove_leading_comma = false;
+    let field = if let Some(line) = line.strip_prefix('"') {
+        remove_leading_comma = true;
+        line.split_once('"')
+    } else {
+        line.split_once(',')
```
jonhoo, line 191: I think this would fail if the number was the last item on the line. That may not happen in the inputs we expect, but maybe worth handling?

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,288 @@
...
+            }
+        }
+
+        if remove_leading_comma {
+            // `remainder` still has a leading comma, because the number is >1000. We need to
+            // remove it so we are consistent regardless of whether the number was wrapped in
+            // double quotes or not.
+            if let Some(remainder) = remainder.strip_prefix(',') {
+                return Ok((n, remainder));
+            }
```
jonhoo, line 230:
```suggestion
            remainder = remainder.strip_prefix(',').unwrap_or(remainder);
```

jonhoo (changes requested):
Something definitely still tugs at me about the A calls B case, but I don't think it's worth blocking this on. It may just be me failing to read in enough detail. And if someone finds an issue later, we'll find out about it then!

I found two more nits, and with those fixed I think we're finally ready to merge!

src/collapse/vsprof.rs
```diff
@@ -0,0 +1,288 @@
...
+    let field = if let Some(line) = line.strip_prefix('"') {
+        remove_leading_comma = true;
+        line.split_once('"')
+    } else {
+        line.split_once(',')
+    };
+
+    if let Some((num, remainder)) = field {
+        // Parse the number
+        let thousands = num.split(|c: char| !c.is_ascii_digit());
```
jonhoo, line 196: Oh, here's maybe one more. Should this perhaps read
```suggestion
        let thousands = num.split(|c| c == ',');
```

KnapSac: Wouldn't that then present a problem with different output cultures? I'm not sure whether Visual Studio always outputs the numbers in the `en-US` culture (so with `,` as thousands separators and `.` as the decimal symbol), or whether they'll output using the current culture, which may use `,` and `.` in the opposite way (so `.` as thousands separator and `,` as the decimal symbol).

To handle that, and to be a bit more explicit, maybe this would be more appropriate:
```suggestion
        let thousands = num.split(|c: char| c == ',' || c == '.');
```

jonhoo: Hmm, it's an interesting point. What I was trying to get at was that there are some characters we probably shouldn't be splitting on because it can make errors worse (e.g., imagine if we get a field that's a word not a number). I think I'd be on board with splitting on `,`, `.`, and ` `.

KnapSac: Why also split on ` `? A space would never occur in an actual number, so why would we split on that?

To get a better error message when the field is a word instead of a number, should we perhaps verify that the field only contains digits, `.`s and `,`s? Although that might introduce some (significant) overhead, as `get_next_number` is called twice for every line in the profile output (once to get the depth and once to get the number of calls), and a large profile could have many lines.

Right now a word instead of a number returns `Expected thousands separators for number bigger than 1000, found '{}'`, which is not exactly a great error.

jonhoo: Space-separated thousands is actually [fairly common](https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html).

Here's a counter-proposal: parse one digit (char) at a time. So instead of going by thousands, we go by tens. At each character:
- if it's a digit, `*= 10; += c - '0';`
- if it's `,`, `.`, or ` `, skip
- otherwise error out and print the whole field

How does that sound?

KnapSac: > Space-separated thousands is actually [fairly common](https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html).

Huh, you learn something new every day I guess haha.

Your proposal sounds good, when I have some time this week I'll give that a shot.

jonhoo (approved):
Excellent, thank you for sticking with it!

jonhoo:
Published as 0.11.2 :tada:

PR #219, opened 2021-10-06
Update MSRV and fix `clippy` warnings

KnapSac:


djc:
I'm not sure why this bumped the MSRV version proactively? It didn't seem to be necessary? The small fixes here don't seem like a very good reason to force downstream code to a relatively fresh MSRV.

PR #224, opened 2021-12-06
Make behavior of skip_after more consistent.

romange:
Specifically, omits process name as a root frame if `skip_after` condition was trigerred.
Also, I added "skip-after" flag to collapse-perf.rs in order to actually be able to use
and test this flag on real stack traces.

### DIFF
--- a/src/bin/collapse-perf.rs
+++ b/src/bin/collapse-perf.rs
@@ -82,8 +82,9 @@ struct Opt {
     infile: Option<PathBuf>,
 
     #[structopt(long = "skip-after", value_name = "STRING")]
-    /// If set, will omit all the stack frames that follow the frame with matched function name.
-    /// Has not effect on the stack trace if no function is matched.
+    /// If set, will omit all the parent stack frames of the frame with matched function name.
+    ///
+    /// Has no effect on the stack trace if no function is matched.
     skip_after: Option<String>,
 }
 
--- a/src/bin/collapse-perf.rs
+++ b/src/bin/collapse-perf.rs
@@ -80,6 +80,11 @@ struct Opt {
     #[structopt(value_name = "PATH")]
     /// Perf script output file, or STDIN if not specified
     infile: Option<PathBuf>,
+
+    #[structopt(long = "skip-after", value_name = "STRING")]
+    /// If set, will omit all the stack frames that follow the frame with matched function name.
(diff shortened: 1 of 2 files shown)


src/bin/collapse-perf.rs
```diff
@@ -80,6 +80,9 @@ struct Opt {
     #[structopt(value_name = "PATH")]
     /// Perf script output file, or STDIN if not specified
     infile: Option<PathBuf>,
+
+    #[structopt(long = "skip-after", value_name = "STRING")]
```
jonhoo, line 84: We should document what this flag does :)

src/collapse/perf.rs
```diff
@@ -568,13 +568,18 @@ impl Folder {
...
+            if self.stack_filter != StackFilter::SkipRemaining {
+                // add the comm name
+                stack_str.push_str(&self.pname);
                 stack_str.push(';');
+            }
+            for e in self.stack.drain(..) {
                 stack_str.push_str(&e);
+                stack_str.push(';');
             }
+            // It's guaranteed that self.stack and therefore stack_str are not empty.
```
jonhoo, line 581: This comment doesn't read right to me?

jonhoo (changes requested):
Just two minor nits. Thanks!

src/bin/collapse-perf.rs
romange: fixed

src/collapse/perf.rs
romange: Changed the wording.

romange:
@jonhoo  addressed your comments. Please advise if the comment above is not clear.

src/bin/collapse-perf.rs
```diff
@@ -80,6 +80,11 @@ struct Opt {
...
+    /// If set, will omit all the stack frames that follow the frame with matched function name.
+    /// Has not effect on the stack trace if no function is matched.
```
jonhoo, line 86:
```suggestion
    /// If set, will omit all the parent stack frames of the frame with matched function name.
    ///
    /// Has no effect on the stack trace if no function is matched.
```

jonhoo (approved):
Thanks!

jonhoo:
Released in 0.10.9 :tada:

PR #227, opened 2022-01-04
Update from structopt to clap v3

Sciencentistguy:
Clap v3 has merged structopt directly into clap, and structopt itself is now in maintenance-only mode.

This PR switches all use of structopt over to clap's new `#[derive(Parser)]`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ codecov = { repository = "jonhoo/inferno", branch = "master", service = "github"
 
 [features]
 default = ["cli", "multithreaded", "nameattr"]
-cli = ["structopt", "env_logger"]
+cli = ["clap", "env_logger"]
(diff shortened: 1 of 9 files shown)


jonhoo:
Thanks for taking the time to do this! It looks like this ends up bumping the MSRV, which I think is okay for this change, but we should update CI to match the new required version :+1:
https://github.com/jonhoo/inferno/blob/502c8f33cb2dff8c08bf806f053a0d0a8659c8a6/azure-pipelines.yml#L5

jonhoo (approved):
Excellent, thanks!

jonhoo:
Actually, before I merge, cc @djc for `cargo-flamegraph` to see how problematic this would be there. I think the move to clap3 is pretty worthwhile.

djc:
Should be fine, we already ported flamegraph (which now disables the cli feature in inferno). I'm discussing the MSRV issue upstream.

PR #231, opened 2022-03-01
Support multiple `--skip-after`

bsilver8192:
perf is stopping short on some of my backtraces (I think it's related to
complex C++ code in the stack), which fragments the relevant functions
in the flamegraph. `--skip-after` works great to avoid this, but I want
to do it for multiple functions.

### DIFF
--- a/src/collapse/perf.rs
+++ b/src/collapse/perf.rs
@@ -71,11 +71,11 @@ pub struct Options {
     /// Default is the number of logical cores on your machine.
     pub nthreads: usize,
 
-    /// If a stack function name is equal to the specified string it will omit all the
+    /// If a stack function name is equal to any of the specified strings it will omit all the
     /// following stackframes for that event.
     /// In case no function is matched the whole stack is returned.
-    /// Default is `None`.
-    pub skip_after: Option<String>,
+    /// Default is not omitting any.
+    pub skip_after: Vec<String>,
 }
 
 impl Default for Options {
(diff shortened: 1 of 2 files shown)


bsilver8192:
> This looks great, and definitely sounds useful! Could you add a test with multiple skip-afters though? Or just update the current one to have multiple filters?

Done. I went with adding a new one to make sure all the branches are covered (I assume that's what codecov is saying? I haven't dug into what numbers it's reporting).

src/collapse/perf.rs
```diff
@@ -815,6 +815,37 @@ mod tests {
...
+        };
+        let mut buf_actual = Vec::new();
+        <Folder as Collapse>::collapse(&mut folder, &bytes[..], &mut buf_actual)?;
+        let lines = std::str::from_utf8(&buf_actual[..]).unwrap().lines();
+
+        for line in lines {
+            // without `skip_after` some collapsed lines would look like:
+            // go;[unknown];x_cgo_notify_runtime_init_done;runtime.main;main.init;...
+            if line.contains("main.init") {
+                assert!(line.contains("main.init;")); // we removed the frames above "main.init"
```
jonhoo, line 839: Actually, reading this test again. I think this should be `line.starts_with`, not `line.contains`. Otherwise it doesn't seem to test much?

bsilver8192: Good point. Fixed them both, and confirmed that the tests fail if I remove things from `skip_after`.

jonhoo (approved):
Excellent, thanks!

PR #234, opened 2022-03-16
Identified external dependencies with `js` color palette

daniellockyer:
- identifying external dependencies is useful because it enables you to
  differentiate between code in your project, and code that you import
  in
- this commit adds an if-statement to check for the `node_modules/`
  string, and colors it purple if found

---

Hey @jonhoo - ❤️ Inferno 🙂 

I've tested this locally with a tool I'm writing for NodeJS and it produces the following: 

CleanShot 2022-03-16 at 10 15 20@2x

### DIFF
--- a/src/flamegraph/color/palettes.rs
+++ b/src/flamegraph/color/palettes.rs
@@ -78,7 +78,9 @@ pub(super) mod js {
         } else if name.contains(':') {
             return BasicPalette::Aqua;
         } else if let Some(ai) = name.find('/') {
-            if (&name[ai..]).contains(".js") {
+            if (&name[ai..]).contains("node_modules/") {
+                return BasicPalette::Purple;
+            } else if (&name[ai..]).contains(".js") {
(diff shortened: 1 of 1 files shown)


daniellockyer:
As a reference for colors, [0x](https://www.npmjs.com/package/0x) colors external dependencies in purple, and I think it looks good here too 🙂 

CleanShot 2022-03-16 at 10 26 17@2x

daniellockyer:
I've added a test case but it seems GitHub are having issues updating the PR 🤔

jonhoo (approved):
This is great, I love it. Thanks!

jonhoo:
Hmm, not entirely sure why CI isn't running. I think it's because I need to "approve and run", but that button doesn't appear. Seems to be some kind of GitHub bug. I'm running the test suite locally, and then I'll update the changelog and publish!

jonhoo:
Released in 0.11.1 :tada:

Issue #237, opened 2022-05-01
Filter out "comments" in inferno-flamegraph input


mrob95:
The [austin](https://github.com/P403n1x87/austin) Python profiler includes some metadata as "comments" in its output, e.g.:
```
# austin: 3.3.0
# interval: 100
# mode: wall
# python: 3.9.10

P247;T247;<frozen importlib._bootstrap>:_install_external_importers:1187;<frozen importlib._bootstrap>:_find_and_load:1007;<frozen importlib._bootstrap>:_find_and_load_unlocked:986;<frozen importlib._bootstrap>:_load_unlocked:680;<frozen importlib._bootstrap>:exec_module:838;<frozen importlib._bootstrap_external>:<module>:35;<frozen importlib._bootstrap>:_find_and_load:1007;<frozen importlib._bootstrap>:_find_and_load_unlocked:986;<frozen importlib._bootstrap>:_load_unlocked:680;<frozen importlib._bootstrap>:exec_module:768;<frozen importlib._bootstrap>:_call_with_frames_removed:228 53
...
P247;T247;/home/mike/profiling/test.py:<module>:6;/usr/lib/python3.9/dis.py:dis:79;/usr/lib/python3.9/dis.py:_disassemble_recursive:373;/usr/lib/python3.9/dis.py:disassemble:369;/usr/lib/python3.9/dis.py:_disassemble_bytes:404 180

# duration: 695345
# sampling: 6,15,159
# saturation: 10/3898
# errors: 4/3898
```

Most of these are correctly ignored as invalid lines by `inferno-flamegraph`, but some (like `# duration: 695345`) look enough like a valid sample to end up in the output svg. I think the original perl version correctly identifies all of them as invalid, since it uses a strict regex to parse lines.

It's fairly trivial to work around this with a `sed '/^#/d'`, but it would be nice for `austin` and `inferno-flamegraph` to work together smoothly.

Would you accept a PR to filter out all lines which begin with `# ` in `inferno-flamegraph`?

Thanks

jonhoo:
Absolutely — that seems reasonable!

PR #239, opened 2022-05-01
Filter out lines which begin with '# ' in inferno-flamegraph

### LINKED ISSUE #238: flamegraph panics on empty lines with --reverse set
Found while I was investigating #237 

Repro:
* Change `tests/data/flamegraph/bad-lines/bad-lines.txt` to the following (adding a blank line):
```
cksum;_start;__libc_start_main;main;cksum 31
cksum;cksum 6
THIS IS A BAD LINE
cksum;cksum;__GI___fread_unlocked;_IO_file_xsgetn;_IO_file_read;entry_SYSCALL_64_fastpath_[k];sys_read_[k];vfs_read_[k];__vfs_read_[k];ext4_file_read_iter_[k] 1
cksum;main;cksum 19
THIS IS A BAD FRACTIONAL LINE 12V.43
noploop;[unknown] 2
noploop;main 274

noploop;[unknown] 2
```
* Run `cargo test flamegraph_should_warn_about_bad_input_lines_with_reversed_stack_ordering`

Failure:
```
---- flamegraph_should_warn_about_bad_input_lines_with_reversed_stack_ordering stdout ----
thread 'flamegraph_should_warn_about_bad_input_lines_with_reversed_stack_ordering' panicked at 'attempt to subtract with overflow', /home/mike/inferno/src/flamegraph/mod.rs:415:60


failures:
    flamegraph_should_warn_about_bad_input_lines_with_reversed_stack_ordering
```

The issue is here: https://github.com/jonhoo/inferno/blob/master/src/flamegraph/mod.rs#L411

`samples_idx` is 0 when we have an empty line, and we are trying to subtract from it.

A simple fix would be to add a check for `line.is_empty()` at the top of that loop. Alternatively, it seems like the reason we are subtracting 1 from sample_idx is to remove a trailing space from the line, so replacing with `&line[..samples_idx].trim_end()` would fix it.

mrob95:
implements #237 

also fixes #238 by moving the check for empty/ignored lines out of `merge` and into `from_lines`. I think this makes more sense as the block in `from_lines` which handles `opt.reverse_stack_order` pre-processes the samples before calling `merge`, and seems to assume that the lines it deals with are not empty. I'm not very experienced with rust though, so please let me know if there is a better way!

I've added an empty line to the `bad-lines` test file as a regression test for #238, and a couple of test files for input from the austin profiler.

### DIFF
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -394,7 +394,10 @@ where
     W: Write,
 {
     let mut reversed = StrStack::new();
-    let lines = lines.into_iter().filter(|line| !(line.is_empty() || line.starts_with("# ")));
+    let lines = lines
+        .into_iter()
+        .map(|line| line.trim())
+        .filter(|line| !(line.is_empty() || line.starts_with("# ")));
 
     let (mut frames, time, ignored, delta_max) = if opt.reverse_stack_order {
         if opt.no_sort {
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -394,6 +394,8 @@ where
     W: Write,
 {
     let mut reversed = StrStack::new();
+    let lines = lines.into_iter().filter(|line| !(line.is_empty() || line.starts_with("# ")));
+
     let (mut frames, time, ignored, delta_max) = if opt.reverse_stack_order {
         if opt.no_sort {
             warn!(
--- a/src/flamegraph/merge.rs
+++ b/src/flamegraph/merge.rs
@@ -124,9 +124,6 @@ where
     let mut prev_line = None;
     for line in lines {
         let mut line = line.trim();
-        if line.is_empty() {
-            continue;
-        }
 
         if !suppress_sort_check {
             if let Some(prev_line) = prev_line {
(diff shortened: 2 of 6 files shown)


src/flamegraph/mod.rs
```diff
@@ -394,6 +394,8 @@ where
     W: Write,
 {
     let mut reversed = StrStack::new();
+    let lines = lines.into_iter().filter(|line| !(line.is_empty() || line.starts_with("# ")));
```
jonhoo, line 397: I think this is missing a `.trim()` call. Without that, a line that's empty except for spaces would not get skipped whereas it does today.

jonhoo (changes requested):
Just one small nit and then I think we're good to land this. Thanks for taking the time to fix the issue too!

src/flamegraph/mod.rs
mrob95: Ah yes, fixed :)

jonhoo (approved):

jonhoo:
Thanks! I'll fix the MSRV issue separately — that's not caused by your change.

jonhoo:
Published as 0.11.3 :tada:

mrob95:
Awesome! Thanks 😊

PR #240, opened 2022-05-04
Add colour palette for python traces

mrob95:
Adds a new colour palette for python. Aiming to distinguish four cases:
* User python code
* python code imported from an installed package
* python code imported from the standard library
* native calls

I've tested it out in a few scenarios and it definitely makes the output more understandable at a glance:
image
image


mrob95:
I have mainly been testing this with the `austin` profiler.

@benfred Do you think this could be adapted to work with [py-spy](https://github.com/benfred/py-spy) as well? It looks to me like py-spy removes a lot of the info which would allow us to tell the difference between user code and library code, but it would be great if we could make it work.

src/flamegraph/color/palettes.rs
```diff
@@ -59,6 +59,21 @@ pub(super) mod perl {
     }
 }
 
+pub(super) mod python {
+    use crate::flamegraph::color::BasicPalette;
+
+    pub fn resolve(name: &str) -> BasicPalette {
+        if name.starts_with("native@") { // austin-specific format for native calls
```
jonhoo, line 66: I worry about encoding generator-specific detection into here. Is there no other way to detect native calls that isn't austin-specific?

src/flamegraph/color/palettes.rs
```diff
@@ -59,6 +59,21 @@ pub(super) mod perl {
...
+            return BasicPalette::Aqua;
+        } else if name.contains("site-packages/") {
```
jonhoo, line 68:
```suggestion
        } else if name.contains("/site-packages/") {
```
is probably more robust

src/flamegraph/color/palettes.rs
```diff
@@ -59,6 +59,21 @@ pub(super) mod perl {
...
+            return BasicPalette::Yellow;
+        } else if name.contains("/python") || name.starts_with("<frozen importlib") { // stdlib
```
jonhoo, line 70: This feels quite brittle. For example, if I do all my Python development under
```
~/dev/python/
```
then _every_ Python file I have will be considered part of the Python standard library. Is there a pattern with lower risk of false positives we can use here?

jonhoo (changes requested):
This looks great! I left some inline notes.

src/flamegraph/color/palettes.rs
mrob95: I agree it's a bit nasty to be doing profiler specific checks here. Unfortunately since getting profiles which combine Python stack traces with native calls is [pretty involved](https://www.benfrederickson.com/profiling-native-python-extensions-with-py-spy/), we are at the mercy of how the profiler decides to format things.

As far as I know the only profilers which support native calls are austin and py-spy, so if we can support both of those then maybe it's fine.

All austin native calls are prefixed with `native@`, e.g.
```
native@7f8b5cbb69a6:ufunc_generic_fastcall:12022
```

py-spy looks like this:
```
ufunc_generic_fastcall (numpy/core/_multiarray_umath.cpython-39-x86_64-linux-gnu.so)
```

I've tagged the author of py-spy in this PR, it would be good to get his input on the best way of parsing this. We could easily do a check for `.so` or `.dll`, but I suspect that if the native extension is built with debug symbols then we would get a reference to a source file, which could have any extension.

Alternatively, I'm happy to remove this branch and ignore native calls for now. For me at least it is very rare to need to go down into native calls when profiling Python.

src/flamegraph/color/palettes.rs
mrob95: You're right. I don't think there is going to be a perfect way of doing it but we can probably do better than this. I've done a couple more experiments:

On a standard Python installation on Windows the stdlib lives in e.g.:
```
C:\Users\Mike\AppData\Local\Programs\Python\Python39\lib\concurrent\futures\thread.py
```

On a standard Ubuntu installation it lives in e.g.:
```
/usr/lib/python3.9/concurrent/futures/thread.py
```

Based on [this](https://stackoverflow.com/questions/2927993/where-are-the-python-modules-stored) SO thread this looks fairly standard, though given the number of ways there are to install Python I would be surprised if it's universal.

So I think a reasonable way of identifying stdlib files would be:
```
name.contains("/python") && name.contains("/lib/")
```

This would need to come after the check for `"/site-packages/"` since those also live under `/lib/python3.9`..

We also need to handle backslashes for windows traces. Would you prefer to just replace all `\` with `/` at the start of this function or duplicate each condition? Or is there a smart way to conditionally compile the function based on the OS?

src/flamegraph/color/palettes.rs
mrob95: Fixed 👍

src/flamegraph/color/palettes.rs
mrob95: I've updated this to look for `/lib/` as well, and also added equivalent checks for Windows paths

jonhoo: Here's another proposal:
```rust
if Path::new(name).iter().any(|c| c.strip_prefix("python").map_or(false, |version| version.chars().all(|c| c.is_digit() || c == '.')))
```
It's not beautiful, but I think it's unlikely to produce false positives, and I _think_ it should map onto both the OS examples you gave above?

jonhoo: Ah, and as far as `/` vs `\` is concerned, use https://doc.rust-lang.org/std/path/constant.MAIN_SEPARATOR.html

jonhoo: Or, actually, for that too you could use:
```rust
if Path::new(name).iter().any(|c| c == "site-packages")
```

src/flamegraph/color/palettes.rs
jonhoo: I'm really torn here. I'd love to hear from the py-spy maintainer too, but I think given the state of the landscape I'm then okay with putting in place checks specifically for austin/py-spy's markers for native calls.

src/flamegraph/color/palettes.rs
mrob95: Very nice! I had to convert from `OsString` to `str` and added a `to_ascii_lowercase` to handle the uppercased path on Windows

src/flamegraph/color/palettes.rs
```diff
@@ -59,6 +59,33 @@ pub(super) mod perl {
...
+        if name.starts_with("native@") {
+            // austin-specific format for native calls
+            return BasicPalette::Aqua;
+        } else if Path::new(name).iter().any(|c| c == "site-packages") {
+            return BasicPalette::Yellow;
+        } else if Path::new(name).iter().any(|c| {
+            c.to_str()
+                .unwrap_or("")
+                .to_ascii_lowercase()
+                .strip_prefix("python")
```
jonhoo, line 76: This allocates, which I'd like to avoid. I think you can use this instead:
```rust
c.strip_prefix("python").or_else(|| c.strip_prefix("Python"))
```

src/flamegraph/color/palettes.rs
```diff
@@ -437,6 +464,43 @@ mod tests {
...
+            TestData {
+                input: String::from("<frozen importlib._bootstrap>:_load_unlocked:680"),
+                output: BasicPalette::Green,
+            },
+            TestData {
+                input: String::from(".venv/lib/python3.9/time.py:12"),
+                output: BasicPalette::Green,
+            },
+            TestData {
+                input: String::from("C:/Users/User/AppData/Local/Programs/Python/Python39/lib/concurrent/futures/thread.py"),
```
jonhoo, line 481: I'd like to see some Windows-style path tests here too :+1:

jonhoo (changes requested):

src/flamegraph/color/palettes.rs
mrob95: Difficulty with that is that `Path` seems to only separate by backslash if you are actually on Windows - so the test fails on Linux. I'm not sure if there is a way to say "pretend you are on windows when running this test"

jonhoo: We run the test suite on Windows too though, so then these tests should fail on Windows, no?

Actually, thinking some more about this, this would mean that if you captured a perf report on Linux and tried to debug it on Windows, it wouldn't work, which is a little concerning..

I wonder if we want to add a helper like:

```rust
fn split_any_path(path: &str) -> impl Iterator<Item = &str> {
    path.split(|c| c == '/' || c == '\')
}
```

mrob95: Ah I hadn't thought of that but yes - if you had a trace from Windows with backslashes in then you couldn't generate a graph on Linux.

I've added the helper you suggested - this also avoids having to deal with `OsString` so is better all round I think :)

mrob95: > We run the test suite on Windows too though, so then these tests should fail on Windows, no?

From the `Path` docs:
> This type supports a number of operations for inspecting a path, including breaking the path into its components (separated by / on Unix and by either / or \ on Windows)

So it makes sense that the current tests pass on windows - it's only going the other way which would be problematic

jonhoo:
Was it intentional that you also made changes to the `flamegraph` submodule?

src/bin/flamegraph.rs
```diff
@@ -82,7 +82,7 @@ struct Opt {
         short = 'c',
         long = "colors",
         default_value = defaults::COLORS,
-        possible_values = &["aqua","blue","green","hot","io","java","js","mem","orange","perl","purple","red","rust","wakeup","yellow"],
+        possible_values = &["aqua","blue","green","hot","io","java","js","mem","orange","perl","python","purple","red","rust","wakeup","yellow","gray"],
```
jonhoo, line 85: Would you mind doing `gray` as a separate PR, just so we don't hold up landing one because of discussion on the other?

src/flamegraph/color/palettes.rs
```diff
@@ -59,6 +59,31 @@ pub(super) mod perl {
...
+        } else if split_any_path(name).any(|part| {
+            part.strip_prefix("python")
+                .or_else(|| part.strip_prefix("Python"))
+                .map_or(false, |version| {
+                    version.chars().all(|c| c.is_digit(10) || c == '.')
+                })
+        }) || name.starts_with("<")
+        {
+            // stdlib
+            return BasicPalette::Gray
```
jonhoo, line 81: Why gray for the stdlib?

src/flamegraph/color/palettes.rs
```diff
@@ -59,6 +59,31 @@ pub(super) mod perl {
...
+pub(super) mod python {
+    use crate::flamegraph::color::BasicPalette;
+
+    fn split_any_path(path: &str) -> impl Iterator<Item = &str> {
+        path.split(|c| c == '/' || c == '\\')
+    }
+
+    pub(in super::super) fn resolve(name: &str) -> BasicPalette {
+      if split_any_path(name).any(|part| part == "site-packages") {
+            return BasicPalette::Mem;
```
jonhoo, line 71: Usually it's better to refer to color palettes here rather than "semantic" palettes. Makes it easier to ensure the colors chosen for different things don't overlap.

Also, I'd probably skip the `return` here and below, and just stick the last line of the function in an `} else {`.

src/flamegraph/color/palettes.rs
```diff
@@ -59,6 +59,31 @@ pub(super) mod perl {
...
+        } else if split_any_path(name).any(|part| {
+            part.strip_prefix("python")
+                .or_else(|| part.strip_prefix("Python"))
+                .map_or(false, |version| {
+                    version.chars().all(|c| c.is_digit(10) || c == '.')
+                })
+        }) || name.starts_with("<")
```
jonhoo, line 78: Why is this clause here? I see the test string below, but are those _always_ standard library? Is there something more we can test on than _just_ `<`?

src/flamegraph/color/palettes.rs
```diff
@@ -59,6 +59,31 @@ pub(super) mod perl {
...
+        {
+            // stdlib
+            return BasicPalette::Gray
+        }
+        BasicPalette::Hot
```
jonhoo, line 83: Same here as for `Mem` above — usually better to stick with single-color schemes.

jonhoo (changes requested):
A few more thoughts

mrob95:
> Was it intentional that you also made changes to the `flamegraph` submodule?

Whoops, nope. Reverted that. To be honest I had forgotten about this PR!

src/flamegraph/color/palettes.rs
mrob95: It was looking a bit garish/fruit salad as it was and I wanted to draw attention away from stdlib calls, as you aren't going to be able to optimise calls into the standard library.

jonhoo:
Heh, yeah, sorry, I've been falling really behind on OSS stuff!

src/flamegraph/color/palettes.rs
mrob95: The angle bracket syntax is used internally by cpython, generally for built-ins, methods on built-in objects, import machinery, etc. For normal user code the name should be the path which should never hit this case. See e.g. here:
https://github.com/python/cpython/blob/main/Python/import.c#L1469

src/bin/flamegraph.rs
mrob95: Will do

src/flamegraph/color/palettes.rs
jonhoo: Sort of — you might still be able to optimize by choosing different calls for example. Gray tends to mean "ignore this", which I don't think applies to "this code spent much of its time in this place in the standard library".

src/flamegraph/color/palettes.rs
jonhoo: Well, so, for example, is it always `<frozen ...`? If so, could we match on that?

mrob95: I'll do a bit of digging - the two I see often are "<frozen ..." and "<built-in ..." so potentially we could just match on those two

src/flamegraph/color/palettes.rs
mrob95: I've switched this to Yellow, which is a python-y colour and doesn't stand out too much

src/flamegraph/color/palettes.rs
mrob95: Narrowed it down to `<frozen`, `<built-in` and `<method`. I've updated the PR to target these specifically

mrob95:
BTW, you may be glad to hear that inferno helped us track down and report an O(n^2) in `boto3` :)

src/bin/flamegraph.rs
jonhoo:
```suggestion
        possible_values = &["aqua","blue","green","hot","io","java","js","mem","orange","perl","python","purple","red","rust","wakeup","yellow"],
```

jonhoo (approved):
Looks great, thanks!

Also, that's so cool to hear :D

jonhoo:
Published in 0.11.11 :tada:

PR #242, opened 2022-05-15
Workaround for Firefox bug: clear viewbox later

### LINKED ISSUE #241: Firefox rendering bug for SVGs
In Firefox there appears to be a race condition where sometimes the height of the flamegraph is not updated fast enough, and so the browser renders it with too large of a height. Here you see two graphs stacked above each other; notice the whitespace between them that shouldn't be there:

Screenshot from 2022-05-15 15-36-23

If I reload the page, the gap (usually) disappears.

The solution I came up with, which I need to test further (it's an intermittent bug...) is to move the viewpoint DOM manipulation until _after_ the new width was created, i.e. `svg.removeAttribute("viewBox");` gets moved to the end of `update_for_width_change()`. I will do some more verification, and if it really fixes the bug, submit a PR.

itamarst:
Fixes #241

### DIFF
--- a/src/flamegraph/flamegraph.js
+++ b/src/flamegraph/flamegraph.js
@@ -42,10 +42,6 @@ function init(evt) {
             var svgWidth = svg.width.baseVal.value;
             searchbtn.attributes.x.value = svgWidth - xpad - 100;
             matchedtxt.attributes.x.value = svgWidth - xpad - 100;
-
-            if (!isEdge) {
-                svg.removeAttribute("viewBox");
-            }
         };
         window.addEventListener('resize', function() {
             update_for_width_change();
@@ -55,6 +51,9 @@ function init(evt) {
             unzoom();
             update_for_width_change();
             restore_state();
+            if (!isEdge) {
+                svg.removeAttribute("viewBox");
+            }
         }, 0);
     } else {
         restore_state();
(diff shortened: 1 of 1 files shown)


jonhoo:
Huh, very interesting. I wonder if this'd be better suited to go in the `setTimeout(function() {` block slightly lower down, just so it doesn't run on _every_ resize (after all, it doesn't do anything after the first time it's invoked). What do you think?

itamarst:
Other points, beyond the location:

* Someone pointed out my HTML was in quirks mode, so need to check if fixing that solves this problem.
* Need to test in Safari now that my new Mac is set up.

I'm trying to remember if I tried it in setTimeout. Possibly not. But there are definitely some rendering race conditions in there that mean setTimeout might not work, will try.

itamarst:
Just validated that disabling quirks mode in my HTML page does not fix the original bug.

itamarst:
And Safari seems happy regardless, both with and without the fix.

itamarst:
Moving to setTimeout seems to also fix the Firefox problem, so I will update PR.

itamarst:
(Going to do a bit more manual testing even after I push the change, I'll let you know when I'm completely convinced.)

jonhoo (approved):
Excellent, thank you!

jonhoo:
Release in 0.11.4 :tada:

PR #252, opened 2022-06-20
Don't quote generic font families in CSS

### LINKED ISSUE #244: Not possible to set font in flamegraphs to default fonts
In CSS I can do:

```
body {
     font-family: monospace;
}
```

but I can't do that with flamegraph, because it turns it into:

```
body {
     font-family: "monospace";
}
```

which means "the font called `monospace`" rather than "default monospace font".

jasonrhansen:
Fixes #244 

When passing a font type to use for the generated flamegraph,
if it is one of the following we shouldn't put quotes around it:

* cursive
* fantasy
* monospace
* serif
* sans-serif

### DIFF
--- a/src/flamegraph/svg.rs
+++ b/src/flamegraph/svg.rs
@@ -9,6 +9,9 @@ use str_stack::StrStack;
 
 use super::{Direction, Options, TextTruncateDirection};
 
+/// The generic font families should not have quotes around them in the CSS.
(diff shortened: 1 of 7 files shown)


jonhoo (approved):
Looks good, thanks!

jonhoo (changes requested):
Oh, actually, can you add to `CHANGELOG.md` while you're at it?

jonhoo (approved):
Thanks!

jonhoo:
Released in 0.11.6 :tada:

PR #253, opened 2022-06-21
Add support for collapsing PHP Xdebug traces

MaartenStaa:
Inspired by and continued from  #126 by @daniellockyer 

This PR is a bit different, because in the meantime Xdebug 3 was released, which apparently introduced a new tracing format as far as I can tell based on the samples I've taken, and the old code. One caveat is that Xdebug actually produces gzipped files, which is not (yet) handled here, as it seems a bit out of scope for this tool.

I went through the comments on the old PR and think I've tackled them all, but please let me know if you have any other feedback.

Also, I love your streams and your book, keep up the great work :)

### DIFF
--- a/src/bin/collapse-xdebug.rs
+++ b/src/bin/collapse-xdebug.rs
@@ -10,7 +10,7 @@ use once_cell::sync::Lazy;
 static NTHREADS: Lazy<String> = Lazy::new(|| format!("{}", *DEFAULT_NTHREADS));
 
 #[derive(Debug, Parser)]
-#[clap(name = "inferno-collapse-xdebug", author = "")]
+#[clap(name = "inferno-collapse-xdebug")]
 struct Opt {
     /// Silence all log output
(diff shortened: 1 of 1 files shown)


lucc:
Hello @MaartenStaa are you still interested in working on this? I would be interested in having this in an inferno release as I am still stuck on some old PHP version (PHP 7.4 on Rocky 8) and that does not have the xdebug.trace_format=3 (I assume your PR is for the trace_format=1).

@jonhoo are you still interested in including this code in inferno? Note that new PHP versions can natively produce a proper flamegraph trace with [the xdebug.trace_format=3](https://xdebug.org/docs/all_settings#trace_format). But it might still be valuable for people that can not yet use xdebug >= 3.3.0.

lucc:
@jonhoo can I help to move  this forward in some way?  I could reopen a new version of this PR to fix the merge conflicts and try to address further comments you might have (in case @MaartenStaa does not return to this PR).

Issue #247, opened 2022-06-16
Consider using text-anchor for right-aligned strings


nagisa:
I’ve used a somewhat modified version of the `inferno` crate to construct flamegraphs for a blog post of mine. I have made a couple of changes that I think some contributor here may be interested in implementing properly in the project proper.

The flamegraph has two strings that are expected to be aligned to the right side of the image: the “Search”/“Reset search” at the top and “Matched: 42%” at the bottom. Right now they are aligned by giving the 100px of “space” from the side, which can fail to fit the string if the font size is increased or a wider font is used.

Instead consider using a `text-anchor: end` and placing the text node at a constant location where the strings should end:

```diff
diff --git a/src/flamegraph/mod.rs b/src/flamegraph/mod.rs
index 4ba2290..a90e742 100644
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -228,8 +199,8 @@ text {{ font-family:{}; font-size:{}px; fill:rgb(0,0,0); }}
         svg,
         &mut buf,
         TextItem {
-            x: Dimension::Pixels(image_width as usize - super::XPAD - 100),
-            y: (opt.font_size * 2) as f64,
+            x: Dimension::Pixels(image_width as usize - super::XPAD),
+            y: (opt.font_size + 5) as f64,
             text: "Search".into(),
             extra: vec![("id", "search")],
         },
@@ -239,7 +210,7 @@ text {{ font-family:{}; font-size:{}px; fill:rgb(0,0,0); }}
         svg,
         &mut buf,
         TextItem {
-            x: Dimension::Pixels(image_width as usize - super::XPAD - 100),
+            x: Dimension::Pixels(image_width as usize - super::XPAD),
             y: (style_options.imageheight - (opt.ypad2() / 2)) as f64,
             text: " ".into(),
             extra: iter::once(("id", "matched")),
diff --git a/src/flamegraph/flamegraph.css b/src/flamegraph/flamegraph.css
index 0bc4184..d47a8b9 100644
--- a/src/flamegraph/flamegraph.css
+++ b/src/flamegraph/flamegraph.css
@@ -1,7 +1,13 @@
+#matched { text-anchor:end; }
+#search { text-anchor:end; }
diff --git a/src/flamegraph/flamegraph.js b/src/flamegraph/flamegraph.js
index cf53d99..fae5f55 100644
--- a/src/flamegraph/flamegraph.js
+++ b/src/flamegraph/flamegraph.js
[...]

PR #254, opened 2022-07-17
Use text-anchor instead of hardcoded offsets

Urgau:
This was proposed by `@nagisa` in https://github.com/jonhoo/inferno/issues/247, I just applied the suggestions and tested that the result was as expected.

I tried the suggestion of `dominant-baseline="hanging"`, but it will need further work to make it useful as applying instead will just increase the size of the CSS without reducing enough in the SVG.

Fixes https://github.com/jonhoo/inferno/issues/247

### DIFF
--- a/tests/flamegraph.rs
+++ b/tests/flamegraph.rs
@@ -63,7 +63,11 @@ fn test_flamegraph_multiple_files(
     }
     // and then compare
     result.set_position(0);
-    compare_results(result, expected, expected_result_file);
+    if std::env::var("INFERNO_BLESS_TESTS").is_ok() {
+        fs::write(expected_result_file, result.get_ref()).unwrap();
+    } else {
+        compare_results(result, expected, expected_result_file);
+    }
     Ok(())
 }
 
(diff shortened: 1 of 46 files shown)


tests/flamegraph.rs
```diff
@@ -63,7 +63,11 @@ fn test_flamegraph_multiple_files(
     }
     // and then compare
     result.set_position(0);
-    compare_results(result, expected, expected_result_file);
+    if std::env::var("FLAMEGRAPH_BLESS").is_ok() {
```
jonhoo, line 66: Oh, yeah, this is a good idea! I wonder if it should be called `INFERNO_BLESS_TESTS` instead though. Also, would you mind updating `compare_results` so that it mentions this envvar if a test output file doesn't match the stored one?

jonhoo (changes requested):
This looks great! Just one note about the new `FLAMEGRAPH_BLESS` envvar.

tests/flamegraph.rs
Urgau: Renamed to `INFERNO_BLESS_TESTS` and added a msg/hint as asked.

tests/flamegraph.rs
```diff
@@ -76,30 +76,33 @@ where
     R: BufRead,
     E: BufRead,
 {
+    const BLESS_MSG: &str =
+        "If you want to bless the output, set `INFERNO_BLESS_TESTS=1` before the `cargo test`";
```
jonhoo, line 80: I think "bless the output" isn't going to be super obvious to most people. How about:

> If you have modified the code and specifically want to update the expected test output, set  `INFERNO_BLESS_TESTS=1` before the `cargo test`.

Urgau: I used the same terminology as `rust-lang/rust` but you're right it isn't really understandable for someone new. Changed to your proposition.

jonhoo (approved):
Perfect, thanks!

tests/flamegraph.rs
```diff
@@ -52,7 +52,7 @@ fn test_flamegraph_multiple_files(
 
     let expected_len = metadata.len() as usize;
     let mut result = Cursor::new(Vec::with_capacity(expected_len));
-    flamegraph::from_files(&mut options, &input_files, &mut result)?;
+    flamegraph::from_files(&mut options, &input_files, &mut result).unwrap();
```
jonhoo, line 55: I think the test failure are due to this change.

Urgau: You're right. Reverted.

jonhoo (approved):

jonhoo:
Released in 0.11.7 :tada:

Issue #248, opened 2022-06-16
Consider adding strokes to the rectangles representing a stack frame


nagisa:
This is leaning towards a stylistic preference somewhat, but adding a stroke to all of the rectangles in a flamegraph should serve as a great way to separate the boxes from each other, especially when they are touching together. Today colour sets include multiple colours and one of the reasons for this I think to differentiate these frames that are touching together.

A stroke would also enable colour sets that are flatter, where colours are used to mark different types of frames, like proposed in #203.

Here's an example of what it might end up looking like. Note especially the `rustix::io::mmap::mprotect` and frames around it: 

![A screenshot of a flamegraph with some boxes touching](https://user-images.githubusercontent.com/679122/174092586-04e78a25-ede8-40df-a74b-9eeb6ed2c7aa.png)

jonhoo:
That's an interesting idea. I'm not a huge fan of it myself, but I could see it making certain graphs (much) easier to read. Maybe an option for this would be a good addition?

PR #255, opened 2022-07-17
Add frame stroke option

Urgau:
This was proposed by `@nagisa` in https://github.com/jonhoo/inferno/issues/248.

The new option will be `--stroke-color` and default to `none` and accept any color.
It use a non-configurable stroke width of 1.

Fixes https://github.com/jonhoo/inferno/issues/248

### DIFF
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -2,7 +2,7 @@ use std::io;
 use std::path::{Path, PathBuf};
 
 use env_logger::Env;
-use inferno::flamegraph::color::{BackgroundColor, PaletteMap, SearchColor};
+use inferno::flamegraph::color::{BackgroundColor, PaletteMap, SearchColor, StrokeColor};
 use inferno::flamegraph::{self, defaults, Direction, Options, Palette, TextTruncateDirection};
 
 #[cfg(feature = "nameattr")]
@@ -174,6 +174,14 @@ struct Opt {
     )]
     search_color: SearchColor,
 
+    /// Adds an outline to every frame
+    #[clap(
+        long = "stroke-color",
+        default_value = defaults::STROKE_COLOR,
+        value_name = "STRING"
+    )]
+    stroke_color: StrokeColor,
+
     /// Second level title (optional)
     #[clap(long = "subtitle", value_name = "STRING")]
     subtitle: Option<String>,
@@ -255,6 +263,7 @@ impl<'a> Opt {
         options.negate_differentials = self.negate;
         options.factor = self.factor;
         options.search_color = self.search_color;
+        options.stroke_color = self.stroke_color;
         (self.infiles, options)
     }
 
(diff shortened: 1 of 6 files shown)


src/flamegraph/mod.rs
```diff
@@ -532,6 +539,12 @@ where
         ]),
     ))?;
 
+    let stroke_color = match opt.stroke_color {
+        StrokeColor::Color(c) => Some(c.to_string()),
+        StrokeColor::None => None,
+    };
```
jonhoo, line 545: Hmm, why is this necessary given you immediately do `as_deref` below?

src/bin/flamegraph.rs
```diff
@@ -174,6 +174,14 @@ struct Opt {
     )]
     search_color: SearchColor,
 
+    /// Stroke color
```
jonhoo, line 177: We should document how this affects the design ("adds an outline to every frame") as well as the format here too I think. Probably also the stroke width.

src/bin/flamegraph.rs
```diff
@@ -174,6 +174,14 @@ struct Opt {
...
+    #[clap(
+        long = "stroke-color",
+        default_value = defaults::STROKE_COLOR,
+        value_name = "STRING"
```
jonhoo, line 181: Why `STRING` and not `COLOR`?

src/flamegraph/mod.rs
```diff
@@ -881,6 +902,13 @@ fn filled_rectangle<W: Write>(
             "fg:x" => &buffer[x_samples],
             "fg:w" => &buffer[width_samples]
         ));
+
+        if let Some(stroke_color) = stroke_color {
+            bytes_start.extend_attributes(args!(
+                "stroke" => stroke_color,
+                "stroke_width" => "1"
+            ));
+        }
```
jonhoo, line 911: It feels unfortunate to add this attribute to every single frame — can we instead just add it to the style sheet instead only when the option is set?

jonhoo (changes requested):
Thanks for taking this on! I left a few comments.

src/bin/flamegraph.rs
Urgau: Same as https://github.com/jonhoo/inferno/blob/c9c07d2f0248e646438b4cba9eae15ef587791d5/src/bin/flamegraph.rs#L170-L175 and also because "none" is not a color but a valid entry for the stroke color.

src/bin/flamegraph.rs
Urgau: I'm not sure it's necessary to define the format here, it isn't done for the other options. As for the stroke width I also don't think we should mention it either, it doesn't add anything in my opinion.

Changed the text to `Adds an outline to every frame`.

src/flamegraph/mod.rs
Urgau: Yes we can. I added them to `#frames > g > rect` like `@nagisa` was doing in their blog.

jonhoo (approved):
Nice, thanks!

jonhoo:
Released in 0.11.7 :tada:

PR #258, opened 2022-09-13
update dependendies (ahash, quick-xml)

crepererum:
Hey :wave: 

We're using your nice crate over at https://github.com/influxdata/influxdb_iox (indirectly via `pprof`) and I've seen that some of our the dependencies that `inferno` pulls are slightly outdated and hence get compiled twice. So I've decided to update them. :slightly_smiling_face:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -668,9 +668,9 @@ dependencies = [
 
 [[package]]
 name = "quick-xml"
-version = "0.25.0"
+version = "0.26.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "58e21a144a0ffb5fad7b464babcdab934a325ad69b7c0373bcfef5cbd9799ca9"
+checksum = "7f50b1c63b38611e7d4d7f68b82d3ad0cc71a2ad2e7f61fc10f1328d917c93cd"
 dependencies = [
  "memchr",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,10 +10,11 @@ checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
 name = "ahash"
-version = "0.7.6"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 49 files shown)


jonhoo (approved):
Excellent, thanks for taking the time!

jonhoo:
Unfortunately, it looks like some of the CI jobs fail :thinking: I'm guessing some of the formatting from `quick-xml` may have changed.

Separately, I noticed that `criterion` has also had a new release — would you mind including that in the PR?

crepererum:
Working on it (probably needs a few iterations because I cannot get the tests to run locally (the `perf` and `dtrace` tests fail all the time and I don't really know why).

jonhoo (approved):
Excellent, thank you!

jonhoo:
Looks like we have a couple of CI failures:

 - The `minimal-versions` one seems to be because we somehow now pull in `itoa 0.4.0`, which is _super_ old. Not sure why. If you run `cargo update -Zminimal-versions` and then `cargo tree -i itoa:0.4.0`, you may get some pointers to what's happened. That said, I'm fine merging even if that one is red. It _could_ be that merging from `main` will fix the problem for you.
 - The MSRV CI step (1.59.0) is because of a MSRV regression in `quick-xml`. It's been fixed upstream, but there hasn't been a release yet: https://github.com/tafia/quick-xml/pull/481#issuecomment-1262799455. I'd actually like to hold off on merging this PR until that lands (at which point we should bump the `quick-xml` dependency to `0.25.1`).
 - The `check / stable /fmt` CI step seems to complain about the lack of `cargo fmt` being applied to a line you changed — should be an easy fix!

crepererum:
This is now solely blocked by the quick-xml release.

jonhoo:
We should also do a bump of clap to major version 4, but I suppose it's no rush (and can be done in a separate PR).

crepererum:
Updated to quick-xml 0.26 -- which fixes the MSRV. Let's do clap in another PR, also because this would mean that we compile two version of clap (at least for testing) because criterion still uses V3.

jonhoo (approved):

jonhoo:
Ah, forgot the changelog! Would you mind updating that too?

crepererum:
> Ah, forgot the changelog! Would you mind updating that too?

#273.

Issue #256, opened 2022-07-21
Slow rendering of flamegraphs


itamarst:
A user complained about slow rendering of flamegraphs, and showed me a video; it was quite slow. Loading the same data into speedscope (https://www.speedscope.app/) was not slow, so the problem was with the way Inferno rendered it.

I tried doing a little performance profiling in Chrome; for context, I'm using 12th-gen i7 desktop, so it's probably got almost double the single-core speed of many (most?) computers.

* Looking at rendering time for a complex SVG (35K lines in input), there's 90ms of time spent in JavaScript.
* If I make `update_text()` a no-op, that goes down to 10ms.

This suggests that coming up with a faster alternative to `update_text()` would improve rendering time for everyone, and hopefully fix rendering problems for people whose computers are extra slow.

Chrome's profiler suggested as a hint that maybe the problem had something to do with https://web.dev/avoid-large-complex-layouts-and-layout-thrashing/

itamarst:
Reading the implementation it also seems not ideal, e.g. growing the string character by character until it's too long.

itamarst:
If it is layout thrashing, fastdom is proposed by the linked article and does seem like it would allow fixing things without a massive refactoring.

itamarst:
From an example:

```javascript
    function reset(done) {
      n = input.value;
      divs = [];

      fastdom.measure(function() {
        var winWidth = window.innerWidth;

        fastdom.mutate(function() {
          container.innerHTML = '';

          for (var i = 0; i < n; i++) {
            var div = document.createElement('div');
            div.style.width = Math.round(Math.random() * winWidth) + 'px';
            container.appendChild(div);
            divs.push(div);
          }

          if (done) done();
        });
      });
    }
```

itamarst:
Coincidentally I just found an example input file that triggers visible slowness on my computer with only 2000 lines of input.
[memory.zip](https://github.com/jonhoo/inferno/files/9168771/memory.zip)

itamarst:
Increasing `flamegraph::Options::min_width` helps a lot for the example above. So my guess: long text + tall stacks + extremely thin columns means a whole lot of looping to shrink the text until it fits.

Perhaps then one improvement would be increasing text, instead of shrinking it? By their nature you'd expect many more thin columns than large columns, so that should reduce useless spinning. And/or maybe use parent column's text width as starting point, since child by its nature is going to be narrower?

jonhoo:
[...]

PR #262, opened 2022-09-27
Speed up reflow when resizing flamegraphs with a monospace font

itamarst:
This addresses #256 partially; variable-width fonts aren't addressed.

By having two passes, one for reading attributes and one for writing attributes, we avoid refreshing the screen a bazillion times, as recommended by https://web.dev/avoid-large-complex-layouts-and-layout-thrashing/.

I tested this by rendering the input file attached in one of the comments on #256 and resizing the window, in both Chrome and Firefox, and with both variations of text truncation. On my very fast desktop, with a monospace font the resize is in real-time, and with variable-width font there is a noticeable pause before refresh happens.

Before merging I should probably at least:

* [x] Make known_font_size calculation only happen once
* [x] Add some comments (based on above) explaining why code is structured the way it is


itamarst:
See attached files to get a sense of the impact, specifically you should resize the window and know how responsive refreshes are

flamegraph-mono
flamegraph-variable

itamarst:
Added `zoom()` support too, so now the refresh on startup is faster.
[result.zip](https://github.com/jonhoo/inferno/files/9685556/result.zip)

src/flamegraph/flamegraph.js
```diff
@@ -162,12 +161,91 @@ function g_to_func(e) {
...
+        var newX = format_percent((parseFloat(r.attributes.x.value) + (100 * 3 / frames.attributes.width.value)));
+
+        // Smaller than this size won't fit anything
+        if (w < 2 * known_font_width) {
+            textElemNewAttributes.push([newX, ""]);
+            continue;
+        }
+
+        // Fit in full text width
+        if (/^ *\$/.test(txt) || txt.length * known_font_width < w) {
```
jonhoo, line 211: Why the empty-text check? Are very-long-but-all-spaces text fields common?

src/flamegraph/flamegraph.js
```diff
@@ -162,12 +161,91 @@ function g_to_func(e) {
...
+            textElemNewAttributes.push([newX, txt.substring(0, substringLength) + ".."]);
+            continue;
+        } else {
+            // Truncate the left side of the text.
+            textElemNewAttributes.push([newX, ".." + txt.substring(txt.length - substringLength, txt.length)]);
+            continue;
+        }
+    }
+    if (textElemNewAttributes.length !== elements.length) {
+        console.log("UH OH", textElemNewAttributes.length, elements.length);
```
jonhoo, line 228: `console.warn` maybe? And let's be a bit more descriptive than "UH OH" :p

jonhoo: Also, should this fall back to the `for` loop?

src/flamegraph/flamegraph.js
```diff
@@ -37,9 +38,7 @@ function init(evt) {
 
             // Text truncation needs to be adjusted for the current width.
             var el = frames.children;
-            for(var i = 0; i < el.length; i++) {
-                update_text(el[i]);
-            }
+            update_text_for_elements(frames.children);
```
jonhoo, line 41: Can't we just pass in `el` here?

src/flamegraph/flamegraph.js
```diff
@@ -162,12 +161,91 @@ function g_to_func(e) {
...
+        return;
+    }
+
+    // Now that we know new textContent, set it all in one go so we don't refresh a bazillion times.
+    for (var i = 0; i < elements.length; i++) {
+        var e = elements[i];
+        var values = textElemNewAttributes[i];
+        var t = find_child(e, "text");
+        t.attributes.x.value = values[0];
+        t.textContent = values[1];
```
jonhoo, line 238: Entirely idle speculation, but I wonder if it'd draw even faster if we _first_ updated all the widths, and _then_ updated all the text content.

src/flamegraph/flamegraph.js
```diff
@@ -162,12 +161,91 @@ function g_to_func(e) {
...
+    // Fall back to inefficient calculation, if we're variable-width font.
+    // TODO This should be optimized somehow too.
+    if (known_font_width === 0) {
+        for (var i = 0; i < elements.length; i++) {
+            update_text(elements[i]);
+        }
+        return;
+    }
+
+    var textElemNewAttributes = [];
```
jonhoo, line 195: This is going to be one big array. Hopefully the memory use spike won't be too bad. Have you observed how big it is (the "performance" tab may tell you)?

jonhoo (commented):
This generally looks good to me, just had a couple of questions

src/flamegraph/flamegraph.js
itamarst: A lot of this code was just copy/pasted unmodified from the original code. Including this. I guess I will... delete this and also delete it from the original code path?

src/flamegraph/flamegraph.js
itamarst: More in the way of an assertion, really. I don't think it _can_ happen at this point.

src/flamegraph/flamegraph.js
itamarst: For my test data, 370KB. So I don't think it's too bad.

src/flamegraph/flamegraph.js
itamarst: :shrug: that doesn't quite match my mental model of how the browser reflows, but it's not necessarily a correct mental model.

itamarst: I tried, didn't seem to make a difference.

itamarst:
Given this only addresses monospace fonts, perhaps the default font should be monospace for now?

itamarst:
I can't tell why tests are failing at this point :(

jonhoo (dismissed):
Could you post a screenshot of a decently-busy flamegraph with and without monospace fonts? I'd like to see how much of a detriment it is to the appearance.

Also, would you mind updating the changelog?

itamarst:
Screenshot from 2022-10-23 17-44-41

I will update changelog once you decide whether you want to change to monospace by default or not.

jonhoo:
Yeah, looks reasonable to me — let's update the default!

jonhoo (approved):
Thanks! Will cut a release once I get a chance.

jonhoo:
Released as 0.11.12 :tada:

PR #267, opened 2022-09-29
Upgrade to version 4 of clap

jonhoo:
Replaces #266 

This increases MSRV to 1.60.0 (because of clap) _when the `cli` feature is enabled_.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -89,7 +89,7 @@ jobs:
     # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


djc:
Another reason to adopt 1.60 (or 1.59) might be to get rid of the num_cpus dependency in favor of `std::thread::available_parallelism()`.

jonhoo:
Yep, good call-out. I'll merge this and then land that and a fix to #278 separately.

jonhoo:
Done in https://github.com/jonhoo/inferno/pull/279

jonhoo:
Released in 0.11.15 :tada:

PR #269, opened 2022-10-09
Upgrade num-format to 0.4.3 and criterion to 0.4

bcmyers:
I neglected to maintain `num-format` (a dependency of `inferno`) for a long time (years, basically); so it was sorely out-of-date. I intend to start maintaining it going forward.

I just published an updated version of `num-format` to crates.io. It's public API remains the same, but under the hood the following dependencies have been updated...

* arrayvec 0.4 -> 0.7.2
* itoa 0.4 -> 1.0.4

This diff ensures `inferno` is using at least `num-format` 0.4.3.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,8 @@ pretty_assertions = "1"
 rand = { version = "0.8", features = ["small_rng"] }
 # Force criterion to pull in regex 1.6 instead of 1.5 during minimal version CI
 regex = { version = "1.6", default-features = false, features = ["std"] }
+# Force criterion to pull in serde 1.0.145 instead of 1.0.0 during minimal version CI
+serde = { version = "1.0.145" }
 testing_logger = "0.1.1"
 
 [profile.release]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ once_cell = "1.12.0"
 
 [dev-dependencies]
 assert_cmd = "2"
-criterion = "0.3"
+criterion = "0.4"
 libflate = "1"
 maplit = "1.0.1"
 pretty_assertions = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,8 @@ libflate = "1"
 maplit = "1.0.1"
 pretty_assertions = "1"
 rand = { version = "0.8", features = ["small_rng"] }
+# Force criterion to pull in regex 1.6 instead of 1.5 during minimal version CI
+regex = { version = "1.6", default-features = false, features = ["std"] }
 testing_logger = "0.1.1"
 
 [profile.release]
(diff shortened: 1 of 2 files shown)


jonhoo:
Ah, so that's where the old itoa came from! I think you may also have to run something like `cargo tree` and `git add` the resulting modifications to the lockfile 👍

jonhoo:
Hmm, it looks like you've somehow pulled in a bunch of changes from the main branch as well?

bcmyers:
Just fixed.

jonhoo (dismissed):

bcmyers:
oh, oops. gotta push lockfile change too... One second.

bcmyers:
Let me try one more time to fix the lockfile...

bcmyers:
Looks like criterion is now the culprit for the fact that itoa 0.4 is sticking around...
```
├── criterion v0.3.6
│   ├── atty v0.2.14 (*)
│   ├── cast v0.3.0
│   ├── clap v2.34.0
│   │   ├── bitflags v1.3.2
│   │   ├── textwrap v0.11.0
│   │   │   └── unicode-width v0.1.10
│   │   └── unicode-width v0.1.10
│   ├── criterion-plot v0.4.5
│   │   ├── cast v0.3.0
│   │   └── itertools v0.10.4 (*)
│   ├── csv v1.1.6
│   │   ├── bstr v0.2.17 (*)
│   │   ├── csv-core v0.1.10
│   │   │   └── memchr v2.5.0
│   │   ├── itoa v0.4.8
│   │   ├── ryu v1.0.11
│   │   └── serde v1.0.144
```

bcmyers:
Perhaps we should upgrade criterion in a separate diff?

bcmyers:
Nevermind. Upgrading criterion to 0.4 was easy. Now we've completely eliminated `itoa 0.4`

Cargo.toml
```diff
@@ -41,11 +41,15 @@ once_cell = "1.12.0"
 
 [dev-dependencies]
 assert_cmd = "2"
-criterion = "0.3"
+criterion = "0.4"
 libflate = "1"
 maplit = "1.0.1"
 pretty_assertions = "1"
 rand = { version = "0.8", features = ["small_rng"] }
+# Force criterion to pull in regex 1.6 instead of 1.5 during minimal version CI
```
jonhoo, line 49: I think it'd be useful to include a brief comment about *why* we do this in each instance. Forcing a higher version isn't a goal in and of itself. What problem does each of these forcings solve?

bcmyers: Yep. Makes sense. I added some more information in the form of comments.

jonhoo (approved):
Thanks!

jonhoo:
Released in 0.11.10 :tada:

PR #271, opened 2022-10-23
Add uicolor option

TheZoq2:
I wanted to embed the flamegraphs in a html document with a dark theme. While setting the background works fine, I couldn't find a way to the text color of the "UI elements", so this is an attempt to add that option.

image
Here is an example with a dark background but white uicolor.

The search button is quite faint here, maybe I should do something about that

Should I also add these options to the CLI part? For my personal use, the library part is enough, but I suppose others may want it via CLI.

I'm very unfamiliar with this code base, so let me know if this is not the right way to do it. Also, feel free to disregard if this is out of scope

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 
-- Add flag `--uicolor` to specify the color to draw UI element text, like the Reset view button
+- Add flag `--uicolor` to specify the color to draw UI element text, like the Reset view button. [#271](https://github.com/jonhoo/inferno/pull/271)
 
 ### Changed
 - Updated `ahash` to version 0.8
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 
+- Add flag `--uicolor` to specify the color to draw UI element text, like the Reset view button
+
 ### Changed
 - Updated `ahash` to version 0.8
 - Updated `quick-xml` to version 0.26
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated `ahash` to version 0.8
 - Updated `quick-xml` to version 0.26
 
+- Bumped `clap` to 3.2.0
+
 ### Removed
 
 ## [0.11.12] - 2022-10-24
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ num-format = { version = "0.4.3", default-features = false }
 quick-xml = { version = "0.26", default-features = false }
 rgb = "0.8.13"
 str_stack = "0.1"
-clap = { version = "3.0.1", optional = true, features = ["derive"] }
+clap = { version = "3.2.0", optional = true, features = ["derive"] }
 once_cell = "1.12.0"
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ num-format = { version = "0.4.3", default-features = false }
 quick-xml = { version = "0.26", default-features = false }
 rgb = "0.8.13"
 str_stack = "0.1"
-clap = { version = "3.2.0", optional = true, features = ["derive"] }
+clap = { version = "3.2.1", optional = true, features = ["derive"] }
 once_cell = "1.12.0"
 
 [dev-dependencies]
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -139,7 +139,7 @@ struct Opt {
         default_value = defaults::UI_COLOR,
         value_parser = |s: &str| {
             parse_hex_color(s)
-                .ok_or_else(|| format!("unknown ui color: {}", s))
+                .ok_or_else(|| format!("Expected a color in hexadecimal format, got: {}", s))
         },
         value_name = "STRING"
     )]
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -141,7 +141,7 @@ struct Opt {
             parse_hex_color(s)
                 .ok_or_else(|| format!("Expected a color in hexadecimal format, got: {}", s))
         },
-        value_name = "STRING"
+        value_name = "#RRGGBB"
     )]
     uicolor: Color,
 
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 
 use env_logger::Env;
 use inferno::flamegraph::color::{
-    parse_hex_color, BackgroundColor, Color, PaletteMap, SearchColor, UiColor,
+    parse_hex_color, BackgroundColor, Color, PaletteMap, SearchColor, StrokeColor,
 };
 use inferno::flamegraph::{self, defaults, Direction, Options, Palette, TextTruncateDirection};
 
@@ -191,10 +191,10 @@ struct Opt {
     /// Adds an outline to every frame
     #[clap(
         long = "stroke-color",
-        default_value = defaults::UI_COLOR,
+        default_value = defaults::STROKE_COLOR,
         value_name = "STRING"
     )]
-    stroke_color: UiColor,
+    stroke_color: StrokeColor,
 
     /// Second level title (optional)
     #[clap(long = "subtitle", value_name = "STRING")]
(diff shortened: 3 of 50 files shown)


TheZoq2:
Updated to also pass tests now. I hope I did that part correctly.

Should I also add a test to make sure that these style options are applied?

src/flamegraph/mod.rs
```diff
@@ -102,6 +102,9 @@ pub struct Options<'a> {
     /// If `None`, the background color will be selected based on the value of `colors`.
     pub bgcolors: Option<color::BackgroundColor>,
 
+    /// The color of UI text such as the search and reset view button. Defaults to black
+    pub uicolor: color::Color,
```
jonhoo, line 106: Given what it does, should this be called `textcolor` instead?

jonhoo (changes requested):
Thanks for taking the time to contribute!

You're on the right track, though I wonder if we should make the change in a slightly different place (see comments). And yes, a test would be great!

Separately, you may also want to add a command-line argument option for this in `src/bin/flamegraph.rs`.

src/flamegraph/svg.rs
```diff
@@ -129,7 +130,7 @@ where
     let titlesize = &opt.font_size + 5;
     svg.write_event(Event::Text(BytesText::from_escaped_str(&format!(
         "
-text {{ font-family:{}; font-size:{}px; fill:rgb(0,0,0); }}
+text {{ font-family:{}; font-size:{}px }}
```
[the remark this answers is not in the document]
TheZoq2: I think I tried that, but that also changes the color of the text inside the flamegraph blocks which I think (at least with the default color scheme) looks better in black.

src/flamegraph/mod.rs
TheZoq2: The answer is similar to the above. This doesn't change all the text, only the free-standing text outside the actual flame graph

TheZoq2:
After adding the binary flag in the latest commit, this is now failing on my local machine with

```
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/test/src/lib.rs:184:5

  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/test/src/lib.rs:184:5

  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/test/src/lib.rs:184:5

  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52/library/test/src/lib.rs:184:5
 ```
It feels like those things shouldn't be related, right?

File: src/bin/flamegraph.rs [hunk 3]
```diff
@@ -131,6 +133,18 @@ struct Opt {
     )]
     fontwidth: f64,
 
+    /// Color of UI text such as the search and reset zoom buttons
+    #[clap(
+        long = "uicolor",
+        default_value = defaults::UI_COLOR,
+        value_parser = |s: &str| {
+            parse_hex_color(s)
+                .ok_or_else(|| format!("unknown ui color: {}", s))
```
jonhoo, line 142: Let's be a little more helpful here. Something along the lines of "expected a color in hexadecimal format"

src/bin/flamegraph.rs
```diff
@@ -131,6 +133,18 @@ struct Opt {
...
+        },
+        value_name = "STRING"
+    )]
+    uicolor: Color,
```
jonhoo, line 146: Why isn't this also the new `UiColor` type? Seems weird for `stroke_color` to be `UiColor`, but _not_ `uicolor`.

src/flamegraph/svg.rs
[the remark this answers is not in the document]
jonhoo: Somehow I read straight past the part where you want the text color inside each of the boxes to remain black! Yes, doing the change this way seems reasonable then.

jonhoo:
> It feels like those things shouldn't be related, right?

I agree. Our tests don't generally fail spuriously so it's a bit concerning, but unlikely to be related to your change :thinking:

Also, looks like there's a conflict with `main`

src/bin/flamegraph.rs
```diff
@@ -177,10 +191,10 @@ struct Opt {
     /// Adds an outline to every frame
     #[clap(
         long = "stroke-color",
-        default_value = defaults::STROKE_COLOR,
+        default_value = defaults::UI_COLOR,
```
jonhoo, line 194: This ends up changing the default from `none` to `#000000`. Was that intentional?

jonhoo (changes requested):

TheZoq2:
Turns out the reason those tests were failing was that I forgot `git update --init` :slightly_smiling_face: 

The UiColor struct was not intentional, I initially started out trying to use it instead of just using `Option<Color>`. Then I accidentally replaced the wrong thing with the LSP before getting rid of its use. Should be better now

src/bin/flamegraph.rs
```diff
@@ -131,6 +133,18 @@ struct Opt {
...
 
+    /// Color of UI text such as the search and reset zoom buttons
+    #[clap(
+        long = "uicolor",
+        default_value = defaults::UI_COLOR,
+        value_parser = |s: &str| {
+            parse_hex_color(s)
+                .ok_or_else(|| format!("Expected a color in hexadecimal format, got: {}", s))
+        },
+        value_name = "STRING"
```
jonhoo, line 144: Feels weird to specify `value_name` as `STRING`. It means that the help text will read:
```
--uicolor STRING
```
which doesn't seem particularly helpful. How about setting it to something like `COLOR` or `#RRGGBB`?

src/bin/flamegraph.rs
```diff
@@ -263,7 +277,7 @@ impl<'a> Opt {
         options.negate_differentials = self.negate;
         options.factor = self.factor;
         options.search_color = self.search_color;
-        options.stroke_color = self.stroke_color;
```
jonhoo, line 266: It doesn't seem right for us to no longer be setting `stroke_color`?

jonhoo (changes requested):

jonhoo:
Oh, and would you mind also updating the CHANGELOG file?

src/bin/flamegraph.rs
TheZoq2: Ah, I thought that was some sort of type checking thing, changing it to #RRGGBB

TheZoq2:
Updated with all three changes

jonhoo (dismissed):

tests/flamegraph.rs
```diff
@@ -730,6 +730,18 @@ fn stroke_color_non_default() {
...
 }
 
+#[test]
+fn ui_color_non_default() {
+    let input_file =
+        "./tests/data/flamegraph/differential/perf-cycles-instructions-01-collapsed-all-diff.txt";
+    let expected_result_file = "./tests/data/flamegraph/options/uicolor_color.svg";
+
+    let mut options = flamegraph::Options::default();
+    options.uicolor = rgb::RGB8{r: 255, g: 0, b: 0};
```
jonhoo, line 740:
```suggestion
    options.uicolor = rgb::RGB8 { r: 255, g: 0, b: 0 };
```

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 
+- Add flag `--uicolor` to specify the color to draw UI element text, like the Reset view button
```
jonhoo, line 10:
```suggestion
- Add flag `--uicolor` to specify the color to draw UI element text, like the Reset view button. [#271](https://github.com/jonhoo/inferno/pull/271)
```

src/bin/flamegraph.rs
```diff
@@ -131,6 +133,18 @@ struct Opt {
     )]
     fontwidth: f64,
 
+    /// Color of UI text such as the search and reset zoom buttons
+    #[clap(
+        long = "uicolor",
+        default_value = defaults::UI_COLOR,
+        value_parser = |s: &str| {
```
jonhoo, line 140: `value_parser` was added in [`clap 3.2.0`](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#320---2022-06-13), so you'll have to update `Cargo.toml` to have that be the required version, then this should work :+1:

jonhoo (dismissed):

Cargo.toml
```diff
@@ -36,7 +36,7 @@ num-format = { version = "0.4.3", default-features = false }
 quick-xml = { version = "0.26", default-features = false }
 rgb = "0.8.13"
 str_stack = "0.1"
-clap = { version = "3.0.1", optional = true, features = ["derive"] }
+clap = { version = "3.2.0", optional = true, features = ["derive"] }
```
jonhoo, line 39:
```suggestion
clap = { version = "3.2.1", optional = true, features = ["derive"] }
```
Looks like there was a same-day fix release.

jonhoo (approved):

jonhoo:
Published as 0.11.13 :tada:

PR #286, opened 2023-03-26
`xctrace` support

### LINKED ISSUE #246: Use `xctrace` on macOS
This is a crosspost from the flamegraph repo: https://github.com/flamegraph-rs/flamegraph/issues/192

> Right now flamegraph uses dtrace on macOS but this runs into some limitations:
> 
> Inline frames are not properly handled (https://github.com/flamegraph-rs/flamegraph/issues/113)
Running dtrace requires root rights
> dtrace is generally considered deprecated functionality on macOS and will likely disappear longer term
> One potential solution is to switch to xctrace and then implement a reader for the .trace / .run files. speedscope appears to have partial support for these files but I haven't been able to make it work at all yet. There is an issue from a while ago that it's not working (https://github.com/jlfwong/speedscope/issues/196) but the code there might still be useful.

Although not familiar with the codebase I'd be ready to give it a go, but would it falls into the scope of `inferno` or do you not want to support an additional profiler?

ldm0:
~~Basically working. Required to be done:~~

- [x] `quick-xml` update #332
- [x] Add more logs and tests
- [x] Documentation and changelog update

Fixes #246

`xctrace` is a tracer used internally by Apple Instruments. It generates a `.trace` folder containing a bunch of files in an unknown binary format. People have [reverse engineerd](http://jamie-wong.com/post/reverse-engineering-instruments-file-format/) it but it's [internal format changed](https://x.com/KacperHarasim/status/1275463460298477570) after XCode 12.

There's a `xctrace export` command, which can convert `.trace` into a readable XML file. However, it's output can't be used to generate flamegraphs until XCode 14.3, [which introduces VM address and symbol name to the XML output](https://developer.apple.com/documentation/xcode-release-notes/xcode-14_3-release-notes#New-Features).

In this PR, I added a stack collapser for XML files generated by `xctrace export`, making it possible to convert `Instruments` or `xctrace`‘s output into flamegraph.


Cargo.toml
```diff
@@ -23,6 +23,7 @@ nameattr = ["indexmap"]
 
 [dependencies]
 ahash = "0.8"
+anyhow = "1.0.70"
```
jonhoo, line 26: Why do we now need a dependency on `anyhow`?

ldm0: Since there are some common functions which can emit error. I thought anyhow can make the error output more informative easility(rather than cascading`return invalid_data_error!(...)`). 

e.g. When there is no ref id in a <backtrace/> tag, we can get a error like:
```
Get ref id of backtrace failed.

Caused by:
    No ref id found
```
just by adding two `.context(` in parsing code.

The error type is eventually transformed into io error `InvalidData`, just like other collapser does.

However if there are some limitation on dependency introducing, I'll remove it.

jonhoo: Anyhow is nice in general, but I'd like to avoid adding it in this particular PR :)

ldm0: Removed

kraktus:
Hey, I was going to start working on integrating this only to see the work was already done! Thanks @ldm0!

I can help add the documentation if needed. Also not sure if this feature need to be tied to the `quick-xml` update?

ldm0:
> I can help add the documentation if needed. Also not sure if this feature need to be tied to the `quick-xml` update?

Thanks! I've written some documentation already, will submit it tomorrow. I think introducing multiple versions of the same dependency is bad(in terms of artifact size and complexity), so I'm waiting for the `quick-xml` update.

BTW, I've forked `flamegraph` with this `inferno` branch and you can try it: `cargo install --git https://github.com/ldm0/flamegraph --branch xctrace`, I replaced the default tracer(`dtrace`) with `xctrace` on macOS.

ldm0:
@jonhoo

Hi, now that #332 has been merged, this PR is ready for review.
I've personally used it for quite a long time and it has worked well. Please take a look.

kraktus:
Same, I've been using it locally without issue

src/collapse/guess.rs
```diff
@@ -109,6 +110,7 @@ impl Collapse for Folder {
             try_collapse_impl!(sample, 2);
             try_collapse_impl!(vtune, 3);
             try_collapse_impl!(vsprof, 4);
+            try_collapse_impl!(xctrace, 5);
             try_collapse_impl!(ghcprof, 5);
```
jonhoo, line 114:
```suggestion
            try_collapse_impl!(ghcprof, 5);
            try_collapse_impl!(xctrace, 6);
```

src/lib.rs
```diff
@@ -70,6 +70,23 @@
...
+//! ### xctrace (macOS)
+//!
+//! ```console
+//! $ target/release/mybin &
+//! $ pid=$!
+//! # xctrace record --template 'Time Profiler' --attach $pid --output out.trace
+//! $ xctrace export --input out.trace --xpath '/trace-toc/*/data/table[@schema="time-profile"]' | inferno-collapse-xctrace | inferno-flamegraph > flamegraph.svg
+//! ```
+//!
+//! If you want demangled output(xctrace won't demangle rust symbols for you), you can use rustfilt:
```
jonhoo, line 82: Let's make rustfilt a link to <https://github.com/luser/rustfilt>

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+    state_backtrace: TagBacktrace,
+    // --------- per-xml caches below -----------
+    /// backtrace_id <--> BackTrace
+    backtraces: BTreeMap<u64, Arc<Backtrace>>,
+    /// backtrace_id <--> Frame
+    frames: BTreeMap<u64, Arc<Frame>>,
+}
+
+// Note that sometimes same Backtrace have different id, we need to merge them
+// before writing folded file.
```
jonhoo, line 39: Do you know why this happens? It could be the kind of thing that's needed to support over-time mode ("flamechart mode") perhaps, I'm not sure.

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+                if n == 0 {
+                    break;
+                }
+            } else {
+                return Some(false);
+            }
+
+            let trimmed = line.trim();
+            if !trimmed.is_empty() {
+                return Some(trimmed.contains(r#"<?xml version="1.0"?>"#));
```
jonhoo, line 261: Is there something more distinct we can look for than just whether it's an XML file?

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+        W: std::io::Write,
+    {
+        let mut context = CollapseContext::default();
+        let mut buf = Vec::new();
+        let nodes = loop {
+            let event = match reader.read_event_into(&mut buf) {
+                Ok(x) => x,
+                Err(e) => return invalid_data_error!("Read xml event failed: {:?}", e),
+            };
+            match event {
```
jonhoo, line 282: It would be really helpful to have a comment here that shows some example XML we're trying to parse to get a sense of the layout.

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+                            CurrentTag::Backtrace(BacktraceState {
+                                id,
+                                frames: Vec::new(),
+                            })
+                        }
+                        (Some(CurrentTag::Backtrace(_)), FRAME) => {
+                            let (id, name) = attributes_to_frame(&attributes)?;
+                            CurrentTag::Frame(FrameState { id, name })
+                        }
+                        _ => CurrentTag::Other(String::from_utf8_lossy(name).into_owned()),
```
jonhoo, line 309: Should this ever happen? If it happens, should we warn and `continue` instead (perhaps with some special logic to not err in the `End`)?

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+    backtrace: Vec<CurrentTag>,
+}
+
+impl TagBacktrace {
+    fn push_back(&mut self, tag: CurrentTag) {
+        self.backtrace.push(tag);
+    }
+
+    /// Remove the top tag and return it. Return `Err(())` on tag mismatching.
+    fn collide(&mut self, name: &[u8]) -> Result<CurrentTag, ()> {
```
jonhoo, line 59: The name and description of this don't quite match. `collide` doesn't imply `pop`, and it's not clear what mismatching/colliding "means". Can I suggest: `pop_with_name`? And then have it return an `Option` instead.

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+            FRAME => matches!(self, Self::Frame(_)),
+            other => matches!(self, Self::Other(tag) if tag.as_bytes() == other),
+        }
+    }
+}
+
+#[derive(Default)]
+struct TraceQueryResultState {
+    nodes: Vec<Node>,
+}
```
jonhoo, line 104: Any particular reason to have all these defined as separate types instead of just defining the fields directly inline on each enum variant of `CurrentTag`?

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+};
+
+const REF: &[u8] = b"ref";
+const ID: &[u8] = b"id";
+const NAME: &[u8] = b"name";
+const TRACE_QUERY_RESULT: &[u8] = b"trace-query-result";
+const NODE: &[u8] = b"node";
+const ROW: &[u8] = b"row";
+const BACKTRACE: &[u8] = b"backtrace";
+const FRAME: &[u8] = b"frame";
```
jonhoo, line 24: It'd be really useful to have a very brief comment for each of this for what that kind of element _means_ in xctrace.

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+            folded.extend(frame_name.as_bytes().iter().copied());
+        }
+        folded
+    }
+}
+
+/// Unescapes the text in xml exported from xctrace.
+fn unescape_xctrace_text(x: Cow<'_, [u8]>) -> Vec<u8> {
+    // xctrace shouldn't give us invalid xml text here, therefore
+    // we don't expect the error branch being hit:
```
jonhoo, line 167: If we don't expect to hit the error case, should we have it actually error instead?

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+    // `quick_xml::escape::unescape` will error out if the input is not a valid xml text:
+    // https://github.com/tafia/quick-xml/blob/0793d6a8d006cb5dabf66bf2a25ddbf198305b46/src/escape.rs#L253
+    if let Ok(x) = quick_xml::escape::unescape(&String::from_utf8_lossy(&x)) {
+        x.into_owned().into_bytes()
+    } else {
+        x.into_owned()
+    }
+}
+
+fn get_ref_id_from_attributes(attributes: &Attributes) -> io::Result<u64> {
```
jonhoo, line 178: I really wish there was a way for us to use `quick-xml`'s [serde support](https://docs.rs/quick-xml/latest/quick_xml/de/index.html) to make this nicer, but sadly I think that only supports whole-document deserialization.

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+        None => return invalid_data_error!("No ref id found in attributes"),
+    };
+    let ref_id = String::from_utf8_lossy(&ref_id);
+    match ref_id.parse() {
+        Ok(x) => Ok(x),
+        Err(e) => invalid_data_error!("Unrecognized ref id: {}: {:?}", ref_id, e),
+    }
+}
+
+fn get_id_from_attributes(attributes: &Attributes) -> io::Result<u64> {
```
jonhoo, line 194: I think you could merge this and the above with a single function that also takes a `key` argument (that's then passed in as either `REF` or `ID`).

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,455 @@
...
+    let id = attributes
+        .clone()
+        .filter_map(|x| x.ok())
+        .find_map(|x| (x.key.into_inner() == ID).then_some(x.value));
+    let id = match id {
+        Some(x) => x,
+        None => return invalid_data_error!("No id found in attributes"),
+    };
```
jonhoo, line 202: I'd probably use let-else here to make this read slightly nicer.
```suggestion
    let Some(id) = id else {
        return invalid_data_error!("No id found in attributes");
    };
```
There are a couple of other places too where this pattern could work nicely.

jonhoo (changes requested):
That's exciting, thanks! Took a first pass now. Nothing too deep to change, though some stuff that'll make this easier to maintain longer-term :)

src/collapse/xctrace.rs
ldm0: This is a [xml declaration](https://www.geeksforgeeks.org/xml-declarations/), I thought it's distinct enough for detecting xctrace output. Any suggestions?

src/collapse/xctrace.rs
ldm0: This is expected to happen. The XML file exported from xctrace often contains more information than necessary for generating flamegraph.

For example, in [tests/data/collapse-xctrace/simple_frame_without_binary_info.xml](https://github.com/jonhoo/inferno/pull/286/files#diff-06dd714fe46a7ecb6fca121c1d6d14a4dd67b92c3ec62fe6a5797b63e37cebe4)
```xml
<row><sample-time id="9426" fmt="00:09.448.362">9448362666</sample-time><thread ref="2"/><process ref="4"/><core ref="20"/><thread-state ref="8"/><weight ref="9"/><backtrace ref="154"/></row>
```

The tag `<sample-time>` is extraneous.

The collapser in this PR walks through all of the XML tags, filters out the XML tags we want and extracts information from them. As for extraneous tags, this collapser just wrap them with `CurrentTag::Other(_)`, push them onto the stack and pop them out after visiting their end tags.

src/collapse/xctrace.rs
ldm0: Fixed. I was thinking of [a game](https://play.google.com/store/apps/details?id=org.jjcfbgzs.ddxpro&hl=en&pli=1) while writing this, where identical blocks collide and disappear. XD

src/collapse/xctrace.rs
jonhoo: Right, but _all_ XML files have that declaration at the top. Could we look for `<trace-query-result>` instead for example?

src/collapse/xctrace.rs
jonhoo: Right, I suppose my hope was that we could avoid pushing + popping them (and thus allocating for them) if we know we're going to skip them anyway. But I don't know if `quick-xml` has a way to declare that we want to completely skip a tag when we see its opening event?

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,439 @@
...
+
+/// Unescapes the text in xml exported from xctrace.
+fn unescape_xctrace_text(text: Cow<'_, [u8]>) -> Vec<u8> {
+    // xctrace shouldn't give us invalid xml text here, therefore
+    // we don't expect the error branch being hit:
+    //
+    // `quick_xml::escape::unescape` will error out if the input is not a valid xml text:
+    // https://github.com/tafia/quick-xml/blob/0793d6a8d006cb5dabf66bf2a25ddbf198305b46/src/escape.rs#L253
+    quick_xml::escape::unescape(&String::from_utf8_lossy(&text))
+        .expect("Invalid xml text from xctrace, which is not expected")
```
jonhoo, line 172: Ah, we probably don't want to panic here (which is what `expect` will do). Instead, I think we should make this function return a `Result` and propagate an `Err` here.

kraktus: Do we really want to propagate the error if we have no plan of handling it anyway?

jonhoo: Yes — we'd rather surface it to the caller who can then choose what to do (including panic), rather than _force_ a program abort.

ldm0: This situation should only occur if there is a bug in xctrace, which is rare, so I choose to panic here.

From inferno's perspective, it might not want to panic over abnormal input(which is not a bug in inferno), I am okay with it.

src/collapse/xctrace.rs
ldm0: Yeah, you're right, the pushing + popping can be skipped.

I choose to push them onto stack to check if tags are matched and bail out on mismatching. But I found that quick_xml already did that by default:
- https://docs.rs/quick-xml/latest/quick_xml/reader/struct.Config.html#structfield.allow_unmatched_ends
- https://docs.rs/quick-xml/latest/quick_xml/reader/struct.Config.html#structfield.check_end_names

src/collapse/xctrace.rs
ldm0: No particular reason, just a style preference though, fixed.

src/collapse/xctrace.rs
ldm0: Fixed, I originally thought nobody else was insane enough to put sample result into XML. XD

src/collapse/xctrace.rs
ldm0: Hmm, I don't know why this happens. xctrace is a black box like many other Apple components.

ldm0: Ah, I think I found out one of the reason, xctrace's sample backtrace is address based. Two identical backtraces might have different addresses. For example:

```rust
fn foo() {
    bar(); // 1
    bar(); // 2
}
```

Then there will be two identical symbolized backtraces with different address set:

```xml
<backtrace id="20">
  <frame id="21" name="rust_test2::bar::h2640131654657f56" addr="0x102af5c30"></frame>
  <frame id="22" name="rust_test2::foo::ha31fba0d06a8a3eb" addr="0x102af5d95"></frame>
</backtrace>
```

```xml
<backtrace id="23">
  <frame ref="21"/>
  <frame id="24" name="rust_test2::foo::ha31fba0d06a8a3eb" addr="0x102af5d99"></frame>
</backtrace>
```

src/collapse/xctrace.rs
ldm0: Added at the top.

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,500 @@
...
+            .last()
+            .map(|t| t.matches(name))
+            .unwrap_or_default()
+        {
+            // There is at least one element in backtrace, hence this unwrap is safe.
+            Some(self.backtrace.pop().unwrap())
+        } else {
+            None
+        }
+    }
```
kraktus, line 138:
```suggestion
    /// If `name` matches the top tag of the stack, the tag is popped.
    fn pop_with_name(&mut self, name: &[u8]) -> Option<CurrentTag> {
        self
            .backtrace
            .last()
            .is_some_and(|t| t.matches(name))
            .then(|| self.backtrace.pop())?
    }
```

ldm0: Thanks!

kraktus: I should be thanking you! Thanks for the great work! I've been using your fork of flame graph for years now

src/collapse/xctrace.rs
jonhoo: Nice find! Could you include that in the comments describing the algorithm here?

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,492 @@
...
+                        }
+                        (
+                            Some(CurrentTag::Row { backtrace }),
+                            CurrentTag::Backtrace { id, frames },
+                        ) => {
+                            let new_backtrace = Arc::new(Backtrace { id, frames });
+                            let ret = context
+                                .backtraces
+                                .insert(new_backtrace.id, new_backtrace.clone());
+                            assert!(ret.is_none());
```
jonhoo, line 388:
```suggestion
                            assert_eq!(ret, None);
```
Gives better errors if the assetion fails.
Although I wonder if this should instead return an `Err` saying "two backtraces with the same ID encountered"?

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,492 @@
...
+                            *backtrace = Some(new_backtrace);
+                        }
+                        (
+                            Some(CurrentTag::Backtrace { id: _, frames }),
+                            CurrentTag::Frame { id, name },
+                        ) => {
+                            let frame = Arc::new(Frame { id, name });
+                            let ret = context.frames.insert(frame.id, frame.clone());
+                            assert!(ret.is_none());
```
jonhoo, line 397: Ditto here as above.

src/collapse/xctrace.rs
```diff
@@ -0,0 +1,492 @@
...
+        self.backtrace.last_mut()
+    }
+}
+
+/// The tag we are scanning, with additional states.
+enum CurrentTag {
+    TraceQueryResult { nodes: Vec<Node> },
+    Node { rows: Vec<Row> },
+    Row { backtrace: Option<Arc<Backtrace>> },
+    Backtrace { id: u64, frames: Vec<Arc<Frame>> },
```
jonhoo, line 142: Do we actually need the `Arc`s here? My sense is that we can just pass around backtrace/frame IDs everywhere, and look them up when we need their contents?

jonhoo (changes requested):
Nice, it's already looking a lot better, thank you!

src/collapse/xctrace.rs
ldm0: Fixed

src/collapse/xctrace.rs
ldm0: Fixed.

src/collapse/xctrace.rs
ldm0: Done

jonhoo (approved):
Sorry about that taking forever! Looks good to me now — let's get this landed and released! :tada:

jonhoo:
Releasing as 0.12.2 in https://github.com/jonhoo/inferno/pull/340

ldm0:
Thanks for reviewing this!

PR #291, opened 2023-05-06
Port stackcollapse-recursive.pl

### LINKED ISSUE #275: Support for simplifying recursive function calls as stackcollapse perl scripts
Is there a way of collapsing recursive calls? The original perl scripts have some support here
https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-recursive.pl

Its different from --skip-after. I dont want to ignore recursion, I want to "compress" it to a single layer.

jacksonriley:
Original script -  https://github.com/brendangregg/FlameGraph/blob/master/stackcollapse-recursive.pl

I was a bit unsure whether this fit better as a standalone script or as an option to the existing collapse scripts, but for a first pass I ported this as a standalone script, as
 - this seemed closer to the function of the original code
 - this allows the script to be run on some pre-collapsed input, rather than requiring you to go back and pass in a `--squash-recursive` option to one of the other collapse scripts after you decide you want to collapse recursive calls.

I can definitely also see benefits to adding this as an option to the existing collapse scripts, so I'm happy to switch to that if you'd prefer!

Fixes #275.

### DIFF
--- a/src/collapse/recursive.rs
+++ b/src/collapse/recursive.rs
@@ -1,0 +1,171 @@
+use super::common::{self, CollapsePrivate};
+use std::{borrow::Cow, io};
+
+/// Recursive backtrace folder configuration options.
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub struct Options {
+    /// The number of threads to use.
+    ///
+    /// Default is the number of logical cores on your machine.
+    pub nthreads: usize,
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Self {
+            nthreads: *common::DEFAULT_NTHREADS,
+        }
+    }
+}
+
+/// A "middleware" folder that receives and outputs the folded stack format
+/// expected by [`crate::flamegraph::from_lines`], collapsing direct recursive
+/// backtraces.
+#[derive(Clone)]
+pub struct Folder {
+    /// The number of stacks per job to send to the threadpool.
+    nstacks_per_job: usize,
+
+    // Options...
+    opt: Options,
+}
+
+impl From<Options> for Folder {
+    fn from(mut opt: Options) -> Self {
+        if opt.nthreads == 0 {
+            opt.nthreads = 1;
+        }
+        Self {
+            nstacks_per_job: common::DEFAULT_NSTACKS_PER_JOB,
+            opt,
+        }
+    }
+}
+
+impl Default for Folder {
+    fn default() -> Self {
+        Options::default().into()
+    }
+}
+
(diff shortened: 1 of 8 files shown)


src/collapse/recursive.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    fn set_nthreads(&mut self, n: usize) {
+        self.opt.nthreads = n;
+    }
+}
+
+impl Folder {
+    fn line_parts(line: &str) -> Option<(&str, usize)> {
+        let mut parts = line.rsplitn(2, ' ');
```
jonhoo, line 117: This can just be `rsplit_once(' ')` now I think?

src/collapse/recursive.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    #[test]
+    fn test_line_parts() {
+        assert_eq!(
+            Folder::line_parts("foo;bar;baz 42"),
+            Some(("foo;bar;baz", 42))
+        );
+        assert_eq!(
+            Folder::line_parts("something; including spaces  42"),
+            Some(("something; including spaces ", 42))
```
jonhoo, line 210: The trailing space here is questionable — I genuinely don't know whether the collapsed format dictates that the count is only ever offset by a _single_ space. Maybe avoid testing that specifically here, and just keep a single space before 42?

jonhoo (changes requested):
This looks great! Two minor nits, and then I think we're ready to land this :tada:

src/collapse/recursive.rs
```diff
@@ -0,0 +1,215 @@
...
+        if !Self::is_recursive(&stack) {
+            return stack;
+        }
+
+        // There is recursion, so we can't get away without allocating a new
+        // String.
+        let mut result = String::with_capacity(stack.len());
+        let mut last = None;
+        for frame in stack.split(';') {
+            match last {
```
jonhoo, line 136: _Technically_ you could reduce the duplication here with something like:
```rust
if last.map_or(true, |l| l != frame) {
```
Not terribly important though

src/collapse/recursive.rs
jacksonriley: Oh nice, thanks!

src/collapse/recursive.rs
jacksonriley: Why not 😄

jonhoo (approved):
Thank you!

jonhoo:
Released in 0.11.16 :tada:

PR #294, opened 2023-05-26
A better explanation for change in differentials

### LINKED ISSUE #293: Differential output tooltips are confusing
Consider these two inputs.

Baseline:

```
base_only 50
both1 25
both2 25
```

Modified:

```
both1 25
both2 50
changed_only 125
```

The backwards looking difference flamegraph has explanatory tooltips as follows:

* `both1 (25 my counts, 12.50%; 0.00%)`
* `both2 (50 my counts, 25.00%; +12.50%)`
* `changed_only (125 my counts, 62.50%; +62.50%)`

The forwards looking difference flamegraph:

* `base_only (50 other counts, 50.00%; -50.00%)`
* `both1 (25 other counts, 25.00%; 0.00%)`
* `both2 (25 other counts, 25.00%; +25.00%)`

The main issue is the difference (+/-) percentages. They are really bad at explaining what happened.

Consider `both2`: it went from 25 to 50. In the backwards looking diff that's described as `+12.50%` even though it doubled. In the forwards looking diff that's described as `both2 (25 other counts, 25.00%; +25.00%)`.

1. This is inconsistent.
2. It's talking about % of overall time, basically, but that's not what most people care about, they care about _absolute_ change. "This function got faster" or "this function uses more memory", not "this function is higher percentage of the whole". If the latter is really waht you care about that's why there's a normalization option when creating the diff.

Consider `changed_only`: it didn't exist in the base profile, only in the changed profile. `+62.50%` does _not_ convey that important piece of information, it just makes it seem like it grew 62% from the base.

itamarst:
An example of how to fix #293

### DIFF
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -626,21 +626,27 @@ where
                     "{} ({} {}, {:.2}%)",
                     function, samples_txt, opt.count_name, pct
                 ),
-                // Special case delta == 0 so we don't format percentage with a + sign.
-                Some(delta) if delta == 0 => write!(
-                    buffer,
-                    "{} ({} {}, {:.2}%; 0.00%)",
-                    function, samples_txt, opt.count_name, pct,
-                ),
-                Some(mut delta) => {
-                    if opt.negate_differentials {
-                        delta = -delta;
-                    }
-                    let delta_pct = (100 * delta) as f64 / (timemax as f64 * opt.factor);
+                Some(delta) => {
+                    let samples = frame.end_time as isize - frame.start_time as isize;
+                    let (old, new) = if opt.negate_differentials {
+                        (samples, samples - delta)
+                    } else {
+                        (samples - delta, samples)
+                    };
+                    let change = match (old, new) {
+                        (0, _) => "↑ added in new".to_string(),
+                        (_, 0) => "↓ removed in new".to_string(),
+                        (x, y) if x == y => "unchanged".to_string(),
+                        (old, new) => {
+                            let ratio = new as f64 / old as f64;
+                            let direction = if ratio > 1.0 { '↑' } else { '↓' };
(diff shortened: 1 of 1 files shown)


itamarst:
Any thoughts on this approach?

src/flamegraph/mod.rs
```diff
@@ -626,21 +626,30 @@ where
...
+                Some(delta) => {
+                    let samples = frame.end_time as isize - frame.start_time as isize;
+                    let (old, new) = if opt.negate_differentials {
+                        (samples, samples - delta)
+                    } else {
+                        (samples - delta, samples)
+                    };
+                    let change = match (old, new) {
+                        (0, _) => "all newly added".to_string(),
+                        (_, 0) => "were all removed".to_string(),
```
jonhoo, line 638: I wonder if these labels still make sense if `negate_differentials` is on.

src/flamegraph/mod.rs
```diff
@@ -626,21 +626,30 @@ where
...
+                        (x, y) if x == y => "".to_string(),
+                        (old, new) => {
+                            let (ratio, compared_to) = if opt.negate_differentials {
+                                (old as f64 / new as f64, "new")
+                            } else {
+                                (new as f64 / old as f64, "old")
+                            };
+                            format!(" = {ratio:.3} × {compared_to} {}", opt.count_name)
```
jonhoo, line 646: I'm not sure I follow the phrasing here. If, for example, we spend 2x as many samples in some frame, this will look like (I think):

> = 2 × old cycles

Which I think is maybe overly concise. It doesn't include the function name (is that printed elsewhere?), and the old _or_ new sample counts. I think we'll want to include at least the function name and one of the two absolute numbers. I'm also not sure if "old" and "new" are understandable labels in this instance (esp. for negate_differentials) — they may even be unnecessary. I think "2×cycles" may actually be sufficient.

PR #300, opened 2023-08-13
add inferno-collapse-ghcprof to handle GHC's prof files

dten:
Because this format is not directly rust related I'm not sure if you would prefer this be part of this repo or be a separate crate listing

Basically GHC (the Haskell compiler) has its own format for profiling traces and this adds collapsing for it.

https://downloads.haskell.org/ghc/latest/docs/users_guide/profiling.html


src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,222 @@
...
+                return Ok(());
+            };
+            let l = String::from_utf8_lossy(&line);
+
+            if l.split_whitespace()
+                .take(START_LINE.len())
+                .eq(START_LINE.iter().cloned())
+            {
+                let cost_centre = 0;
+                let module = l.find(START_LINE[2]).unwrap_or(0);
```
jonhoo, line 78: I would just use `"MODULE"` here I think. `START_LINE[2]` is mostly just obscuring what this does.

src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,222 @@
...
+                // "%time %alloc   %time %alloc"
+                // `ticks` and `bytes` columns might appear on the end
+                // ticks header is right aligned
+                // bytes header tries to be right aligned but has a max width
+                // "%time %alloc   %time %alloc  ticks  bytes"
+                let source = match self.opt.source {
+                    Source::PercentTime => l.find("%time").unwrap(),
+                    // ticks and bytes columns are weirdly aligned so find the end of the col before
+                    Source::Ticks => l.rfind("%alloc").unwrap() + 6,
+                    Source::Bytes => l.rfind("ticks").unwrap() + 5,
```
jonhoo, line 89:
```suggestion
                    Source::Ticks => l.rfind("%alloc").unwrap() + "%alloc".len(),,
                    Source::Bytes => l.rfind("ticks").unwrap() + "ticks".len(),,
```

jonhoo: I'd like to do better than `.unwrap()` here. If the user has asked to source from a column that doesn't exist in the input file, we should return an `Err`, not panic.

src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,222 @@
...
+        let mut occurrences = Occurrences::new(1);
+        loop {
+            line.clear();
+            if reader.read_until(b'\n', &mut line)? == 0 {
+                break;
+            }
+            let l = String::from_utf8_lossy(&line);
+            let line = l.trim_end();
+            if line.is_empty() {
+                break;
```
jonhoo, line 111: Is it impossible in this format for there to be empty lines in the data? As in, between rows of profiling information.

src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,222 @@
...
+                self.stack.truncate(depth);
+            } else if depth != prev_len {
+                return invalid_data_error!("Skipped indentation level at line:\n{}", line);
+            }
+            // There can be non-ascii names so take care to char offset not byte offset
+            let string_range = |col_start: usize| {
+                line.chars()
+                    .skip(col_start)
+                    .skip_while(|c| c.is_whitespace())
+                    .take_while(|c| !c.is_whitespace())
```
jonhoo, line 199: This won't work for fields that have spaces _in_ them. To do that you'd need to also store the _end_ index of a column, read to the end, and then `trim_right`. Probably not worth it, but worth flagging in a comment in case someone every thinks this _might_ be appropriate for fields with "spaced" values.

src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,222 @@
...
+                    .collect::<String>()
+            };
+            let cost = string_range(cols.source);
+            if let Ok(cost) = cost.trim().parse::<f64>() {
+                let func = string_range(cols.cost_centre);
+                let file = string_range(cols.module);
+                // Costs include self + calls (at least in what we parse)
```
jonhoo, line 206: I don't think this is true for `%time`. At least the _first_ `%time` column is listed as "individual", which makes me think it doesn't include child calls. May be true for ticks/bytes though?

src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,222 @@
...
+                line.chars()
+                    .skip(col_start)
+                    .skip_while(|c| c.is_whitespace())
+                    .take_while(|c| !c.is_whitespace())
+                    .collect::<String>()
+            };
+            let cost = string_range(cols.source);
+            if let Ok(cost) = cost.trim().parse::<f64>() {
+                let func = string_range(cols.cost_centre);
+                let file = string_range(cols.module);
```
jonhoo, line 205: I probably wouldn't call this `file`, given there's a separate field called `src`, which really _is_ file. Call it module maybe?

src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,222 @@
...
+            if depth < prev_len {
+                // If the line is not a child, pop stack to the stack before the new depth
+                self.stack.truncate(depth);
+            } else if depth != prev_len {
+                return invalid_data_error!("Skipped indentation level at line:\n{}", line);
+            }
+            // There can be non-ascii names so take care to char offset not byte offset
+            let string_range = |col_start: usize| {
+                line.chars()
+                    .skip(col_start)
```
jonhoo, line 197: This assumes that GHC's "thing that produces aligned output" aligns by Unicode code point, but that's not _necessarily_ the case. It could be aligning by glyph or grapheme or grapheme cluster instead, in which case this would not give us the right result. Someone mentioned that GHC has a `-pj` flag to produce JSON output — maybe it's worth digging into whether we should just use that instead?

jonhoo: Not to mention, this would make us resistant to if GHC changes this (seemingly human-facing) format.

dten: this is the code that decides the length of function/module names etc.

https://github.com/ghc/ghc/blob/master/rts/ProfilerReport.c#L40

not sure what the implications are..

jonhoo: It looks like they're trying to essentially detect what "kind" of UTF-8 byte they're looking at (based on the [codepage layout](https://en.wikipedia.org/wiki/UTF-8#Codepage_layout)). However I don't _think_ that'll actually come the same thing as `chars()` for code points with, say, 4 bytes in them :eyes: Based on the name of that function though, it certainly feels like what they're _going_ for is the utf-8 encoded string length (which is what `.chars()` will give you), so you may be okay :+1:

src/bin/collapse-ghcprof.rs
```diff
@@ -0,0 +1,76 @@
...
+)]
+struct Opt {
+    // ************* //
+    // *** FLAGS *** //
+    // ************* //
+    /// Count bytes
+    #[clap(long = "bytes")]
+    bytes: bool,
+    /// Count ticks
+    #[clap(long = "ticks")]
```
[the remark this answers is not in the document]
dten: Done via clap ArgGroup

src/bin/collapse-ghcprof.rs
```diff
@@ -0,0 +1,76 @@
...
+    about,
+    after_help = "\
+[1] This processes the .prof output of GHC (Glorious Haskell Compiler)
+    "
+)]
+struct Opt {
+    // ************* //
+    // *** FLAGS *** //
+    // ************* //
+    /// Count bytes
```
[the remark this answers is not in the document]
dten: I've added a few more words around the references to ticks and bytes to help clairfy

src/bin/collapse-ghcprof.rs
```diff
@@ -0,0 +1,76 @@
...
+impl Opt {
+    fn into_parts(self) -> (Option<PathBuf>, Options) {
+        let mut options = Options::default();
+        options.source =
+            if self.ticks {
+                Source::Ticks
+            } else if self.bytes {
+                Source::Bytes
+            } else {
+                Source::PercentTime
```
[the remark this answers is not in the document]
dten: I've added an explicit `--time` arg to the group and documented it is the default if no other is specified

src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,222 @@
+use std::io::{self, BufRead};
+
+use log::warn;
+
+use crate::collapse::common::Occurrences;
+use crate::collapse::Collapse;
+
+// If source is included traces start after this line (ignoring spaces)
+static START_LINE: &[&str] = &[
+    "COST", "CENTRE", "MODULE", "SRC", "no.", "entries", "%time", "%alloc", "%time", "%alloc",
```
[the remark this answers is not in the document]
dten: The ticks and bytes columns are optional which is noted when parsing them but i've added a comment here too

src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,222 @@
...
+#[derive(Clone, Debug, Default)]
+#[non_exhaustive]
+pub enum Source {
+    #[default]
+    /// The %time column
+    PercentTime,
+    /// The ticks column
+    Ticks,
+    /// The bytes column
+    Bytes,
```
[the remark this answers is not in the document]
dten: I've added some more comments where ticks and bytes is documented hopefully clarifying what they represent

src/collapse/ghcprof.rs
dten: added a comment to say it's not expected to contain blank lines

src/collapse/ghcprof.rs
dten: added a comment to say the expected targets do not contain whitespace

src/collapse/ghcprof.rs
dten: You're right. I've corrected the comment

src/collapse/ghcprof.rs
dten: renamed it to module

dten:
I think I've addressed all the comments with updates except for the utf8 one.
I will probably have a look at the json file option at a later date

dten:
these are the svgs of the test outputs, would you like them committing?

percent
ticks
ticks_bytes
ticks_ticks
utf8
utf8_bytes
utf8_ticks

src/bin/collapse-ghcprof.rs
```diff
@@ -0,0 +1,86 @@
...
+use env_logger::Env;
+use inferno::collapse::ghcprof::{Folder, Options, Source};
+use inferno::collapse::Collapse;
+
+#[derive(Debug, Parser)]
+#[clap(
+    name = "inferno-collapse-ghcprof",
+    about,
+    after_help = "\
+[1] This processes the .prof output of GHC (Glorious Haskell Compiler)
```
konsumlamm, line 14:
```suggestion
[1] This processes the .prof output of GHC (Glasgow Haskell Compiler)
```
The G stands for Glasgow, not Glorious.

dten: Oh they seem to have dropped the glorious when I wasn't looking. Thanks 👍

dten: fixed

konsumlamm: It was never called "Glorious Haskell Compiler". The reason you were confused is probably because `ghc --version` says `The Glorious Glasgow Haskell Compilation System`.

src/collapse/ghcprof.rs
```diff
@@ -0,0 +1,240 @@
...
+                let module = l.find("MODULE").unwrap_or(0);
+                // Pick out these fixed columns, first two are individual only
+                // "%time %alloc   %time %alloc"
+                // `ticks` and `bytes` columns are optional and might appear on the end
+                // ticks header is right aligned
+                // bytes header is right aligned
+                //   - BUT it has a max width of 9 whilst its values can exceed (but are always space separted)
+                // "%time %alloc   %time %alloc  ticks  bytes"
+                let source = match self.opt.source {
+                    Source::PercentTime => l.find("%time").unwrap(),
```
jonhoo, line 87: This still panics, but should return an error instead (like the other two now do).

jonhoo (changes requested):
Nice! We're almost there. Just one more bit.

No need to commit the SVGs — just wanted a sanity-check that things looked reasonable.

src/collapse/ghcprof.rs
dten: it's one of those cases where we "know" it's there since otherwise we couldn't be in the block because we just matched it as part of `START_LINE`. I've added a nice return error anyway and made clippy happy

jonhoo: Ah, in that case you can use `expect` instead and give the rationale!

dten: okay swapped it

jonhoo (dismissed):
🎉

jonhoo:
Oh, looks like you may need to merge from main too

dten:
rebased

jonhoo:
Oh, sorry, one very last thing — can you add an entry to `CHANGELOG.md` too? That way I can do a release right after merging!

jonhoo (approved):
Thank you!

jonhoo:
Released in 0.11.17 :tada:

PR #301, opened 2023-08-25
Bump MSRV to match clap dependency

rukai:
Since [clap 4.4 bumped its MSRV to 1.70](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#440---2023-08-24) inferno no longer builds on 1.64 and needs minimum 1.70
This PR:
* bumps inferno's MSRV to 1.70
* makes use of a new stdlib API in 1.70 to drop the is-terminal and rustix dependencies.
    + This was also cargo's motivation for bumping MSRV which is how I noticed we could do the same here.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -75,7 +75,7 @@ jobs:
     # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
     strategy:
       matrix:
-        msrv: [1.64.0] # clap 4 (via clap_lex)
+        msrv: [1.70.0] # clap 4
     name: ubuntu / ${{ matrix.msrv }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


jonhoo (commented):
Thanks! While I appreciate the PR, I don't think I want to bump inferno's MSRV to 1.70 yet — it is too new for the time being. And I don't think dropping the `is-terminal` dependency is _that_ valuable in reality. The fact that `clap 4.4` no longer supports 1.64 doesn't actually preclude anyone from building `inferno` with 1.64, as long as they come up with the right lockfile (which, while annoying, is at least possible). If we land these changes, then users of `inferno` _have_ to use a newer Rust, or pin an older version of inferno.

I haven't formally set an MSRV policy for this crate, but in general I'd be looking at 12-18 months probably.

rukai:
Understandable, I'll mark this as draft so it's around when it's needed.

jonhoo:
Given I'm about to do a breaking release anyway in https://github.com/jonhoo/inferno/pull/335, I'll land this as part of it! May have to do some cherry-picking, but hey, it did end up useful in the end :D

jonhoo:
Merged in https://github.com/jonhoo/inferno/pull/336

PR #302, opened 2023-08-25
Fix `cargo clippy --all-targets` lints

rukai:
`clippy::disallowed_names` was complaining about variables named `foo` in a test, it seemed a silly thing to complain about so I just disabled it instead.
Happy to instead rename those variables if desired.

The rest all seem uncontroversial to me.

### DIFF
--- a/src/flamegraph/color/palettes.rs
+++ b/src/flamegraph/color/palettes.rs
@@ -106,9 +106,9 @@ pub(super) mod js {
         } else if name.contains(':') {
             return BasicPalette::Aqua;
         } else if let Some(ai) = name.find('/') {
-            if (&name[ai..]).contains("node_modules/") {
+            if name[ai..].contains("node_modules/") {
                 return BasicPalette::Purple;
-            } else if (&name[ai..]).contains(".js") {
+            } else if name[ai..].contains(".js") {
                 return BasicPalette::Green;
             }
         }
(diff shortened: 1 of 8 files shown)


src/flamegraph/color/palettes.rs
```diff
@@ -106,9 +106,9 @@ pub(super) mod js {
         } else if name.contains(':') {
             return BasicPalette::Aqua;
         } else if let Some(ai) = name.find('/') {
-            if (&name[ai..]).contains("node_modules/") {
+            if (name[ai..]).contains("node_modules/") {
```
jonhoo, line 109: I think this now doesn't need the surrounding `()` either, no?

src/flamegraph/color/palettes.rs
```diff
@@ -106,9 +106,9 @@ pub(super) mod js {
...
                 return BasicPalette::Purple;
-            } else if (&name[ai..]).contains(".js") {
+            } else if (name[ai..]).contains(".js") {
```
jonhoo, line 111: Same here ref `()`

jonhoo (dismissed):
Thank you for taking the time to do this — I've been wanting to sort those warnings out for a while! Two small bits that I think we could change further, but otherwise this looks great!

jonhoo:
Also, yes, definitely the right call to disable that lint. Feels like a bug in the lint really to be disallowing these names in test code...

rukai:
I addressed those two issues

jonhoo (approved):
Thanks!

PR #304, opened 2023-09-06
consider sample count when parsing folded line when using `base` option

conectado:
When filtering lines given the `base` option, the line was parsed using `;` but given that a space-separated sample count could be included the first iterated value was always ignored to consider whether or not it's part of the base. This caused all the base samples that are leaves to be ignored, this PR fixed that by first splitting using a whitespace.


jonhoo:
Ah, good catch! Would you mind including a test case for this too so we make sure this doesn't regress in the future?

PR #307, opened 2023-11-04
Bump ahash dependency

julianbraha:
The currently-used version of ahash, 0.8.3 was yanked due to this vulnerability: tkaitchuck/aHash#163

This PR bumps it to 0.8.6, the latest version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,14 +10,15 @@ checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
 name = "ahash"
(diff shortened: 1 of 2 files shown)


jonhoo:
Thanks for the heads up! It shouldn't be necessary to update `Cargo.toml`, since for binary consumers the latest version (or what's in `Cargo.lock`) will be used, and for library consumers they should be permitted to control the version they consume through their `Cargo.toml`/`Cargo.lock`. I'll push a commit + release that just `cargo update -p ahash` for binary consumers :+1:

jonhoo:
Published fix in 0.11.18 :tada:

PR #321, opened 2024-07-14
Bump dashmap

jonhoo:
Replaces https://github.com/jonhoo/inferno/pull/317


jonhoo:
Not a breaking change because `Occurences` is only used in a sealed trait.

PR #326, opened 2024-07-17
Fix warnings

Levminer:
As requested on stream :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -141,6 +141,9 @@
 //!   [sample]: https://gist.github.com/loderunner/36724cc9ee8db66db305#profiling-with-sample
 //!   [VTune]: https://software.intel.com/en-us/vtune-amplifier-help-command-line-interface
 
+// Remove when https://github.com/rust-lang/rust/issues/101730 lands
+#![allow(unknown_lints)]
+
 #![cfg_attr(doc, warn(rustdoc::all))]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -141,6 +141,9 @@
 //!   [sample]: https://gist.github.com/loderunner/36724cc9ee8db66db305#profiling-with-sample
 //!   [VTune]: https://software.intel.com/en-us/vtune-amplifier-help-command-line-interface
 
+// Remove when https://github.com/rust-lang/rust/issues/101730 lands
+#![allow(unknown_lints)]
```
jonhoo, line 145: Something seems weird here. After all, `rustdoc::all` shouldn't be enabling `rustdoc::missing_doc_code_examples` if that's not stable yet (if it does, that feels like a bug). Can we instead just remove the `#[allow]`?

Levminer: Its enabled on this line: https://github.com/Levminer/inferno/blob/3419a93cbb12c3a867011862f7d43f0e398f09b3/src/lib.rs#L148

jonhoo: That's an `allow()` line, so it's just making us _not_ get warnings about that particular lint.

Levminer: Ah, okay, but it DOES give warnings 😂. Nevermind, removed that line.

jonhoo (approved):
Thanks!

PR #331, opened 2024-08-14
Fix remaining cargo warnings

korrat:
This PR fixes all warnings reported by `cargo clippy` without further settings.

There is one more warning on this line, due to the experimental lint, which is not supported on stable Rust. This occurs in `cargo test` and `cargo doc`.
https://github.com/jonhoo/inferno/blob/726e7f3631a9e986c60417cc09d53a90dced5e91/src/lib.rs#L145

The fix in the second commit changes behaviour, though arguably, that was the intended behaviour. Alternatively, using `.truncate(false)` would explicitly preserve the current behaviour.


jonhoo (approved):
Nice, thank you!

Issue #325, opened 2024-07-15
Upgrade quick-xml


jonhoo:
We're _way_ behind. At the time of writing, the [latest is 0.36.0](https://github.com/tafia/quick-xml/releases/tag/v0.36.0), and [we're on 0.26.0](https://github.com/jonhoo/inferno/blob/3e788062b9c4a80fe6254983832c06e18f1305bd/Cargo.toml#L40)!

A while back I opened a PR moving us to 0.27.0 (https://github.com/jonhoo/inferno/pull/282), but it languished since I was hesitant to do a major (read: breaking) release. As part of doing this upgrade, we should make quick-xml be a _private_ dependency and not expose its error type directly, so that we don't run into this again in the future.

tomerab1:
Hi @jonhoo,

I'd like to work on upgrading `quick-xml` to version `0.36.0` and making it a private dependency.
Could you please assign this issue to me?

Thanks!

kraktus:
closed by https://github.com/jonhoo/inferno/pull/332 I think

jonhoo:
Yes indeed, thank you!

PR #332, opened 2024-09-28
Bump quick-xml to 0.37.0 and remove it from public APIs

RedPhoenixQ:
This addresses #325.

The `quick_xml::Writer` used in inferno can only ever return `std::io::Error` but wrapped in `quick_xml::Error`. To fix this a small helper function has been added to "unwrap" this io error and return that as the public API.

I'm going to submit a PR to quick-xml to only give out `std::io::Error` from their `Writer` to make this behavior clear. There might be a good reason that I don't see.

I'm happy to make any other changes, like making a custom Error type if that's nicer for the public API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Remove `quick_xml::Error` from the public API. [#332](https://github.com/jonhoo/inferno/pull/332)
+
 ### Removed
 
 ## [0.11.21] - 2024-08-03
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -713,9 +713,9 @@ dependencies = [
 
 [[package]]
 name = "quick-xml"
-version = "0.26.0"
+version = "0.37.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f50b1c63b38611e7d4d7f68b82d3ad0cc71a2ad2e7f61fc10f1328d917c93cd"
+checksum = "ffbfb3ddf5364c9cfcd65549a1e7b801d0e8d1b14c1a1590a6408aa93cfbfa84"
 dependencies = [
  "memchr",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ indexmap = { version = "2.0", optional = true }
 itoa = "1"
 log = "0.4"
 num-format = { version = "0.4.3", default-features = false }
-quick-xml = { version = "0.26", default-features = false }
+quick-xml = { version = "0.37", default-features = false }
 rgb = "0.8.13"
 str_stack = "0.1"
 clap = { version = "4.0.1", optional = true, features = ["derive"] }
--- a/src/bin/flamegraph.rs
+++ b/src/bin/flamegraph.rs
@@ -309,7 +309,7 @@ impl<'a> Opt {
 
 const PALETTE_MAP_FILE: &str = "palette.map"; // default name for the palette map file
 
-fn main() -> quick_xml::Result<()> {
+fn main() -> io::Result<()> {
     let opt = Opt::parse();
 
     // Initialize logger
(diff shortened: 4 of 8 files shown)


jonhoo (approved):
This is great, thank you, and thanks for also filing a change to `quick-xml`! I wonder whether we should wait for 0.37 for when your `io::Error`-only change is present? I suppose with the type now gone from the public API, we should be able to transparently upgrade later anyway, so there's no huge reason to delay.

One question I have is whether we should be exposing `io::Error` directly too, or whether we should introduce an opaque error type (say, `inferno::Error`) that internally (for now) is just `io::Error`), but that we could extend in the future if needed to contain other kinds of errors if we pick up dependencies that _can't_ as easily be turned into `io::Error` :thinking: What do you think?

Also, would you mind updating `CHANGELOG.md`?

jonhoo:
Don't worry about the coverage job failing btw, it's a codecov problem, not a problem with your change.

jonhoo:
Looks like the `features` CI job failure is legitimate though.

RedPhoenixQ:
> This is great, thank you, and thanks for also filing a change to `quick-xml`! I wonder whether we should wait for 0.37 for when your `io::Error`-only change is present? I suppose with the type now gone from the public API, we should be able to transparently upgrade later anyway, so there's no huge reason to delay.

It should be fine as the PR in `quick_xml` has already merged and it unlikely that it would change again before a new release.

> One question I have is whether we should be exposing `io::Error` directly too, or whether we should introduce an opaque error type (say, `inferno::Error`) that internally (for now) is just `io::Error`), but that we could extend in the future if needed to contain other kinds of errors if we pick up dependencies that _can't_ as easily be turned into `io::Error` 🤔 What do you think?

If we do, would that type have `#[non_exhaustive]` to prevent breaking changes in the case where more errors are added? If not it would be a breaking change either way. I don't know what the best shape of the custom error would be. `io::Error` isn't `Clone` so the custom error can't be without `Arc`, which is unfortunate. If that case comes io::Error does have an `ErrorKind::Other` that would allow for passing any error as `io::Error`, but not as nice or explicit. This would at least temporarily allow for the new error without a breaking change and maybe that enough.

I'm willing to create a custom error if that's nicer. I would probably do it without `Clone` och `PartialEq` as `io::Error` doesn't implement it and thus neither do we, or should we in that case?

jonhoo:
0.37 is out already it seems! :tada: 
https://github.com/tafia/quick-xml/releases/tag/v0.37.0

For an opaque error, I was thinking it would be a `struct`, not an `enum`, and indeed with `#[non_exhaustive]`. No public fields (or variants) so that we can change what's inside the error at will. Sort of like [`reqwest::Error`](https://docs.rs/reqwest/0.12/reqwest/struct.Error.html) for example. And yeah, completely agree it probably shouldn't be `Clone` or `PartialEq`.

What we're balancing here is how likely it is that we'd ever want custom errors in inferno, possibly from dependencies, that we don't want to coerce into `io::Error`. Going to seek some input here from @djc who maintains `cargo-flamegraph`, as they are one of the biggest consumer of `fantoccini` — how much of a pain would it be if we introduced our own error variant instead of just producing `io::Error`s?

RedPhoenixQ:
> 0.37 is out already it seems! 🎉 https://github.com/tafia/quick-xml/releases/tag/v0.37.0

Must have happened today! I updated the title and bumped to 0.37. 

I'm still okay with creating a custom error if a decision gets made on it.

djc:
(flamegraph has an inferno dependency, not fantoccini -- I think you got some crates confused? 😄)

flamegraph immediately converts any inferno errors into `anyhow::Error`, so it really doesn't care about anything other than that the `std::error::Error` trait is implemented.

(Note, I have been working for a few years on instant-xml as an alternative to quick-xml. Might be interesting for inferno, although it doesn't support everything inferno needs -- so far noticed that there's no support for pretty printing and its `Serializer` interface wraps `fmt::Write` rather than `io::Write`. Feels weird that `File`/`BufWriter` don't implement `fmt::Write`...)

jonhoo:
(that's what I get for bouncing through a bunch of issues in a short amount of time :sweat_smile: )

Okay, that sounds good. In that case, let's just stick with `io::Error` like you have it now @RedPhoenixQ!

On `instant-xml`, that does sound interesting. We don't particularly care about the pretty printing I don't think :+1: 

Otherwise, just a heads up that I'll be on holiday for the next three weeks, so won't have a chance to land this until I get back :) But I'll approve the PR to make it clear that the intent is to land it + release as a new major version when I return!

jonhoo (approved):

Issue #299, opened 2023-07-24
Document cargo features in readme


rukai:
Users who use this crate as library will very likely want the cli feature disabled and its also not clear to me what the `nameattr` feature is for even after looking at the code.
So a section in the readme or maybe the root of the crate docs that lists what every feature is and which are enabled by default would be really valuable.

jonhoo:
I completely agree! I really wish there was a standard way to document features in Cargo/the Rust ecosystem. There's some discussion in https://github.com/rust-lang/api-guidelines/discussions/43, but no hard guidance as of yet. @itamarst maybe you could take a stab at this after you land #294?

PR #333, opened 2024-09-28
Add docs for feature flags

RedPhoenixQ:
Addresses #299.

I've added docs for the feature available feature flags. I believe these are correct but as I haven't used this crate myself I'm only going of the code and documentation. 

The style is inspired by [tracing's feature docs](https://docs.rs/tracing/latest/tracing/#crate-feature-flags) which I consider to be clear and readable. As I understand it there is no standardized format for feature docs yet.

I've not release any crates myself so I'm open to making any changes that would make the docs better or clearer (or correct if I've misunderstood some features).

I picked this up after watching the backlog of streams on the Open Source maintenance and saw that this was yet to be done.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,6 +116,12 @@
 //! $ inferno-diff-folded folded2 folded1 | inferno-flamegraph --negate > diff1.svg
 //! ```
 //!
+//! # Feature flags
+//! All features below are enabled by default
+//! - `cli`: Enabled when inferno compiled to a standalone binary and is used from the command line
+//! - `multithreaded`: Enables multithreading aggregating values for [`collapse`]
+//! - `nameattr`: Allows for adding customizing and adding attributes to the svg of [`flamegraph`]. See the `--nameattr` option for the flamegraph cli
+//!
 //! # Development
 //!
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -116,6 +116,12 @@
 //! $ inferno-diff-folded folded2 folded1 | inferno-flamegraph --negate > diff1.svg
 //! ```
 //!
+//! # Feature flags
+//! All features below are enabled by default
+//! - `cli`: Enabled when inferno compiled to a standalone binary and is used from the command line
```
jonhoo, line 121:
```suggestion
//! - `cli`: Also builds the `inferno` command-line tools
```

src/lib.rs
```diff
@@ -116,6 +116,12 @@
...
+//! - `multithreaded`: Enables multithreading aggregating values for [`collapse`]
```
jonhoo, line 122:
```suggestion
//! - `multithreaded`: Enables multithreaded stack-collapsing
```

jonhoo (changes requested):
Thanks for taking this on! Just a few tweaks to the text and then I'm happy :)

jonhoo (approved):

PR #334, opened 2024-10-14
Recommend an alternative addr2line tool for speed reasons

marxin:
The `perf script` can easily run for a few minutes for applications like Firefox. For speed comparison, please see https://github.com/gimli-rs/addr2line?tab=readme-ov-file#performance.

Note: perf 6.12 hasn't been released yet, but will be available in a couple of weeks from now.


jonhoo (approved):
Oh nice, that's a good call-out!

PR #338, opened 2024-12-21
Trim reversed stacks to prevent sort check from failing

bobrik:
If the last line of the stack starts with a space, then the space is used in sorting. However, the space is removed when the line is then fed into the flamegraph, which makes the sorting invalid. This commit adds a trim to the reverse process to fix this.

### DIFF
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -429,6 +429,7 @@ where
             }
             stack.push(' ');
             stack.push_str(&line[samples_idx..]);
+            let stack = stack.trim();
             reversed.push(&stack);
         }
         let mut reversed: Vec<&str> = reversed.iter().collect();
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -429,6 +429,9 @@ where
             }
             stack.push(' ');
             stack.push_str(&line[samples_idx..]);
+            // Trim to handle the case where functions names internally contain `;`.
+            // This can happen, for example, with types like `[u8; 8]` in Rust.
+            // See https://github.com/jonhoo/inferno/pull/338.
             let stack = stack.trim();
             reversed.push(&stack);
         }
--- a/tests/data/flamegraph/fractional-samples/with-space-reversed.svg
+++ b/tests/data/flamegraph/fractional-samples/with-space-reversed.svg
@@ -1,0 +1,167 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" width="1200" height="246" onload="init(evt)" viewBox="0 0 1200 246" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:fg="http://github.com/jonhoo/inferno">
+    <!--Flame graph stack visualization. See https://github.com/brendangregg/FlameGraph for latest version, and http://www.brendangregg.com/flamegraphs.html for examples.-->
+    <!--NOTES: -->
+    <defs>
+        <linearGradient id="background" y1="0" y2="1" x1="0" x2="0">
+            <stop stop-color="#eeeeee" offset="5%"/>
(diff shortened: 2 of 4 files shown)


jonhoo:
Thank you! I'm curious though — how can this happen in practice? Like, what causes a function to have a leading space in its name? How do you end up with the file that you included in this PR?

tests/data/flamegraph/fractional-samples/fractional-with-space.txt
```diff
@@ -0,0 +1,6 @@
+cksum;_start;__libc_start_main;main;cksum 31.23
+cksum;cksum 6.1
+cksum;cksum;__GI___fread_unlocked;_IO_file_xsgetn;_IO_file_read;entry_SYSCALL_64_fastpath_[k];sys_read_[k];vfs_read_[k];__vfs_read_[k];ext4_file_read_iter_[k] 1.4
+cksum;main; cksum 19.0
```
jonhoo, line 4: This line specifically

bobrik:
I think some symbols can have `;`:

```
0000000000a759a0 l     F .text	0000000000000008              core::array::<impl core::default::Default for [T; _]>::default::h67c9877e6f2a615c
0000000000a759b0 l     F .text	0000000000000008              core::array::<impl core::default::Default for [T; _]>::default::h8cc48dfe0d9bff78
0000000000a75800 l     F .text	0000000000000097              core::array::<impl core::fmt::Debug for [T; N]>::fmt::h24e996bfa7154ca1
0000000000a759d0 l     F .text	0000000000000008              core::array::<impl core::default::Default for [T; _]>::default::he4723b7c57f6b881
0000000000a00af0 g     F .text	0000000000000074              <[u8; 19] as phf_shared::FmtConst>::fmt_const::hb6b803208e5640c0
0000000000a002f0 g     F .text	0000000000000074              <[u8; 3] as phf_shared::FmtConst>::fmt_const::h524dac4335e1334b
0000000000a00370 g     F .text	0000000000000074              <[u8; 4] as phf_shared::FmtConst>::fmt_const::h117038a50e8f7ca3
0000000000a009f0 g     F .text	0000000000000074              <[u8; 17] as phf_shared::FmtConst>::fmt_const::h4e993ac5b3e99e7c
```

Since the stack frame separator in collapsed files is `;`, they split into two, with the second one having a space before the symbol name.

In our script we have:

```
perf script --no-demangle -i perf.data | ... | c++filt | /usr/bin/inferno-collapse-perf --all | ...  > perf.stacks.collapsed
```

Arguably, these collapsed stacks are somewhat incorrect, but having a somewhat incorrect flamegraph beats having none.

jonhoo:
Oh iiiinteresting. That is indeed pretty broken. I wonder if @brendangregg has thought about how the collapse format for flamegraph more broadly should handle symbols with `;` in them?

In the meantime, would you mind making the test collapse hold lines more along the lines of what you'd realistically get with a Rust type like `[u8; 3]`? The current one feels hand-edited and isn't intuitively obvious how it can happen (which is what prompted this discussion).

bobrik:
> Oh iiiinteresting. That is indeed pretty broken. I wonder if @brendangregg has thought about how the collapse format for flamegraph more broadly should handle symbols with `;` in them?

Internally we use `→` when we do some pre-processing to prepend cgroup hierarchy to the stack trace. There's little chance of it being in the raw data.

> In the meantime, would you mind making the test collapse hold lines more along the lines of what you'd realistically get with a Rust type like `[u8; 3]`? The current one feels hand-edited and isn't intuitively obvious how it can happen (which is what prompted this discussion).

Good idea, updated with a realistic name.

src/flamegraph/mod.rs
```diff
@@ -429,6 +429,7 @@ where
             }
             stack.push(' ');
             stack.push_str(&line[samples_idx..]);
+            let stack = stack.trim();
```
jonhoo, line 432:
```suggestion
            // Trim to handle the case where functions names internally contain `;`.
            // This can happen, for example, with types like `[u8; 8]` in Rust.
            // See https://github.com/jonhoo/inferno/pull/338.
            let stack = stack.trim();
```

jonhoo (dismissed):
Thanks!

jonhoo (approved):

jonhoo:
Releasing in https://github.com/jonhoo/inferno/pull/339

PR #343, opened 2025-05-05
Bump ahash to 0.8.7 for -Zminimal-versions

ppentchev:
Hi,

Thanks a lot for writing and maintaining inferno!

What do you think about this trivial patch that will fix the build with nightly Cargo's `-Zminimal-versions` mode? I routinely try to do this for my own projects, and an `ahash` -> `inferno` -> `tracing-subscriber` dependency chain led me here :) Of course, it is entirely up to you whether you want to support such whims.

Now, in theory one might say that a nightly-toolchain `-Zminimal-versions` build might even be added to the GitHub workflows or some such; let me know if you want me to try to do that, although I have to admit I am not overly familiar with GitHub workflows.

Thanks in advance, and keep up the great work!

G'luck,
Peter

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ multithreaded = ["dashmap", "crossbeam-utils", "crossbeam-channel"]
 nameattr = ["indexmap"]
 
 [dependencies]
-ahash = "0.8"
+ahash = "0.8.7"
 crossbeam-utils = { version = "0.8", optional = true }
 crossbeam-channel = { version = "0.5", optional = true }
 dashmap = { version = "6.0.1", optional = true }


jonhoo (approved):
Absolutely! I'm horrified that I'm one of the offenders here. Happy to add this, although I'm somewhat curious about how we're relying on ahash 0.8.7? What is the feature we're using that isn't present in 0.8.0?

jonhoo:
Releasing in https://github.com/jonhoo/inferno/pull/346

ppentchev:
> Also worth pointing out that we _do_ actually have a CI job that checks minimal versions: https://github.com/jonhoo/inferno/actions/runs/16089142824/job/45403920395?pr=343

Hm, ICBW, but it seems to me that that job only checks that the crate builds with the minimum supported Rust version; it will still pull the latest (compatible) versions of the crates it depends on, and so it will use all the bugfixes in those latest versions. What I'm talking about is using the `-Zminimal-versions` option to `cargo` that is only available in, well, Rust nightly :) It looks at the version requirements in `Cargo.toml` and it picks the earliest compatible versions, not the latest ones. Its main goal is, as you surmised, to make sure you don't accidentally start depending on a feature that was introduced in ahash 0.8.7 while declaring a dependency on ahash 0.8.0, but sometimes it may also catch incompatibilities caused by behavior changes in Rust itself, or, as in this case, it may uncover dependency specifications that were too lax to begin with: the commit that bumped the ahash dependency to 0.8 - https://github.com/jonhoo/inferno/commit/7f14ad7390e53f9214648daa9da5fa508c7a617b - does not build with `cargo build -Zminimal-versions` either, since ahash 0.8.0 itself fails to build.

jonhoo:
The job I linked specifically builds with `-Zminimal-versions`. We have a separate job that does MSRV checking :) And yes, I know well both how painful maintaining minimal versions is as I have that check in pretty much all my projects.

PR #345, opened 2025-07-03
Visualization options for easier viewing

willhansen:
Add four new options:  

- `--include-children`
    - When present, each frame of a differential flamegraph will be coloured according to the sum of all of that frame's children
    - Very useful when many small functions collectively impact performance

- `--frame-width-source`
    - Chooses the frame widths of differential frame graphs, removing need to generate two different graphs every time
    - Possible values (taking A and B as the first and second sample counts for each stack):
       - `before`:     A
       - `after`:      B
       - `difference`: abs(A-B)
       - `common`:     min(A, B)
       - `allsamples`: A+B
       - `max`:        max(A, B)

- `--detailed-tooltips`
    - Provides more information in the tooltips of differential flamegraphs

- `--normalize`
    - Normalizes each stack sample count so that the total sample count of the two measurements does not interfere with comparison
    - Unlike the stack diff tool, uses floating point numbers to better handle single sample stacks

Add nix flake for dev shell


src/flamegraph/mod.rs
```diff
@@ -590,54 +710,165 @@ where
...
+                            )
+                        } else {
+                            (
+                                frame_self_and_total_counts.total_count.before,
+                                overall_total_count.before,
+                            )
+                        };
+                    if opt.negate_differentials {
+                        delta_pct_pt = -delta_pct_pt;
+                        // std::mem::swap(&mut frame_count_before, &mut frame_count_after);
```
jonhoo, line 842: This seems to be leftover from debugging?

jonhoo (changes requested):
Thank you!

I'm not opposed to any of these changes necessarily, though it's very hard to review this with four different features (five when you count nix) all intermingled in the same PR. Could you instead please open a dedicated PR for each PR?

Issue #164, opened 2020-03-07
API for creating flamegraphs that isn't tied to strings


itamarst:
Hi,

I am attempting to embed source code in my flamegraphs, basically to have inline tracebacks/callstacks, and... it's a little tricky. The requirement for text formatting means I need to escape semicolons, white space apparently needs to be trimmed, numbers at the end of a line confuse it I think...

There's also the waste of time memory of taking in-memory objects, turning them into strings, and then parsing them again.

It would be good to have a structured API for creating flamegraphs, in addition to the existing line-of-text parsing API.

jonhoo:
Yup, I completely agree with you! We started working out some of the API over in https://github.com/jonhoo/inferno/issues/30, but no great need ever arose (since the bottleneck is usually in the collection tool anyway) so it never got pushed very far. If you want to pick it up though, I'd be happy to review a PR!

itamarst:
Ah, cool. I already have my own collapsing code, so I would probably just focus on splitting up flamegraph generation.

PR #358, opened 2026-03-02
feat: add typed API for flamegraph and collapse

jonhoo:
Inferno's library API was purely text-based: collapsers produce `"stack;frames count\n"` lines, and flamegraph parses them back. Downstream tools with structured stack data had to round-trip through text, causing escaping issues, wasted work, and loss of type safety (issues #30, #164, #270).

This commit adds a typed API alongside the existing text path. The core type is `Sample<S>`, generic over the stack iterator, which replaces the text line `"main;foo;bar 42"`. Three new entry points accept iterators of `Sample`: `from_sorted_stacks` (requires pre-sorted input), `from_stacks` (sorts internally), and `flame_chart_from_stacks` (preserves time ordering). On the collapse side,
`Collapse::collapse_to_stacks` returns a `FoldedStacks` wrapper whose `samples()` method yields typed `Sample` values for direct rendering without text serialization. `CollapsePrivate` implementors get an efficient override that skips text entirely; other `Collapse` implementors get a default that round-trips through text that they can override if able.

`Options` is now `Clone` and has no lifetime parameter. The mutable `palette_map` reference moves to a separate parameter on all public functions, and text-processing fields (`no_sort`, `reverse_stack_order`, `flame_chart`, `base`) move to a new `PreProcessingOptions` struct that is only needed by the text-based entry points.

This is a breaking API change, hence the version bump to `1.0.0-rc.1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Overhauled public crate API to be more useful for non-text inputs.
+
 ### Deprecated
 
 ### Removed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -541,7 +541,7 @@ dependencies = [
 
 [[package]]
 name = "inferno"
-version = "0.12.6"
+version = "1.0.0-rc.1"
 dependencies = [
  "ahash",
  "assert_cmd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "inferno"
-version = "0.12.6"
+version = "1.0.0-rc.1"
 edition = "2021"
 authors = ["Jon Gjengset <jon@thesquareplanet.com>"]
 rust-version = "1.71.0"
--- a/benches/flamegraph.rs
+++ b/benches/flamegraph.rs
@@ -2,9 +2,15 @@ use std::fs::File;
 use std::io::{self, BufReader, Read};
 
 use criterion::*;
-use inferno::flamegraph::{self, Options};
+use inferno::flamegraph::{self, Options, PreProcessingOptions};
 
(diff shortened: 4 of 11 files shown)


jonhoo:
This took me forever to get around to, but here's finally a first draft of a more reasonable crate API for inferno.

Would love to get input from the various people interested in #30, #164, and #270, like @koute, @Licenser, @Kixiron, @itamarst, and @stepancheg.

I also figure that once we land this, we're probably ready to mark inferno as 1.0 :eyes:

Metbcy (changes requested):
Found a real regression, sorry for the delay in flagging it properly.

**Sort check disagrees with the sort it's checking against.** `frames_from_stacks` now compares split `Vec<&str>` frames instead of whole lines, but `from_lines` still sorts input as raw byte strings (`lines.sort_unstable()`), and the collapse side (`Occurrences::drain_sorted_entries`, default `collapse_to_stacks`) also sorts by `String`. Those two orderings disagree whenever a frame is a prefix of another frame followed by a byte < `;` (0x3B) - e.g. `:` in Rust/C++ symbols. Repro:

```
$ printf 'main::foo;x 1\nmain;y 1\n' | inferno-flamegraph -
main (base):  exit 0, renders fine
this PR:      Error: "unsorted input samples detected"
```

So it sorts the input itself, then rejects its own sort. Only `from_stacks` is internally consistent since it sorts by the same `Vec<&str>` it checks against.

Also: zero test coverage anywhere for `collapse_to_stacks`/`FoldedStacks` (grepped tests/benches/bin - all new tests only exercise the flamegraph-side fns), which is presumably why this slipped through. And the error string changed (`"unsorted input lines"` -> `"unsorted input samples"`) without a CHANGELOG mention - minor but worth a line since anyone matching on it breaks.

Otherwise the design is solid and the mechanical Options/PreProcessingOptions split is clean and consistent everywhere I checked. Once the sort-order mismatch is fixed and collapse_to_stacks has coverage, happy to take another pass.

PR #361, opened 2026-04-15
Switch from libflate to flate2

mkrasnitski:
The `flate2` crate is more popular and as far as I can tell, `inferno` makes no attempt to be `no_std`. The `GzDecoder::new` constructor is infallible because `libflate` attempts to read and validate the gzip header on construction, whereas `flate2` does not and waits until the decoder is actually used to perform validation. Additionally, `flate2` uses a `BufReader` internally, though `GzDecoder` only exposes a `Read` interface.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,10 @@
 version = 3
 
 [[package]]
-name = "adler32"
-version = "1.2.0"
+name = "adler2"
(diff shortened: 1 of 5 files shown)


jonhoo (approved):
Looks reasonable to me, thanks!

PR #364, opened 2026-05-01
Fix DTrace link in README.md

snpefk:
While going through README.md I noticed that DTrace links leads to 404 page on joyent.com. Replace the original link with https://dtrace.org to fix that.


jonhoo (approved):
Thanks!

PR #365, opened 2026-05-05
Generate "ignore case" button on flamegraphs

jputlock:
This feature is already implemented in the original perl script.

Changes:
- Adds an "ignore case" button to the top right of flamegraphs, shifting the search button left slightly
  - Added a tooltip for this, which forced direct use of the svg library instead of calling `write_str`
- `Ctrl+i` can be used to toggle this dynamically

"ignore case" is additionally bound into the URI parameters.

To verify, I ran `cargo run --bin inferno-flamegraph < tests/data/flamegraph/base/flames.txt > test.svg` and got the following flamegraph:
image

Hovering:
image

Searching `final` (standard behavior):
image

Searching `final` with ignore case:
image

### DIFF
--- a/src/flamegraph/flamegraph.css
+++ b/src/flamegraph/flamegraph.css
@@ -1,6 +1,8 @@
 #matched { text-anchor:end; }
 #search { text-anchor:end; opacity:0.1; cursor:pointer; }
 #search:hover, #search.show { opacity:1; }
+#ignorecase { text-anchor:end; opacity:0.1; cursor:pointer; }
+#ignorecase:hover, #ignorecase.show { opacity:1; }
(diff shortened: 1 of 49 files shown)


jonhoo:
Oh, nice! Could you please link to the original PR for this change in the upstream Perl project?

jputlock:
> Oh, nice! Could you please link to the original PR for this change in the upstream Perl project?

https://github.com/brendangregg/FlameGraph/pull/211

jputlock:
@jonhoo sorry for the ping; let me know if there's anything else I need to / can do here

jputlock:
Any news here?

PR #367, opened 2026-05-24
fix: avoid integer overflow when computing percentage for large sample counts

### LINKED ISSUE #170: "offwaketime" output causes 'attempt to multiply with overflow'
Worth noting btw, initially I have built it in release, and it has shown just invalid top-half of stacktraces. [Related report on vanilla flamegraph](https://github.com/brendangregg/FlameGraph/issues/233)

## Steps to reproduce

1. Make sure `bcc-tools` [are installed](https://github.com/iovisor/bcc/blob/master/INSTALL.md) *(or alternatively use the attached `out.txt`)*.
2. `cargo build` to compile the project in debug mode
3. Create a 15 seconds trace `sudo /usr/share/bcc/tools/offwaketime -f 15 > /tmp/out.txt`, *(or alternatively use the attached `out.txt`)*.
4. `target/debug/inferno-flamegraph /tmp/out.txt > /tmp/out.svg`

An `out.txt` from my system: [out.txt](https://github.com/brendangregg/FlameGraph/files/4508961/out.txt)

### Expected

It creates an svg file

### Actual

It crashes with

```
thread 'main' panicked at 'attempt to multiply with overflow', /home/constantine/Projects/inferno/src/flamegraph/mod.rs:539:23
```

## Version

latest git as of commit adc83b9

SAY-5:
Fixes #170.

When a stack has a very large sample count (e.g. `offwaketime` output with values around 2.4e17), `100 * samples` overflowed the integer type and panicked with "attempt to multiply with overflow" at `src/flamegraph/mod.rs`. The same applied to `100 * delta` in the differential path.

Performing the `* 100` in floating point (`100.0 * x as f64`) avoids the overflow. Output is unchanged for normal-magnitude inputs.

Added a regression test that feeds large sample counts through `from_lines` and asserts it does not panic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Avoid integer overflow when computing percentages for very large sample counts. [#367](https://github.com/jonhoo/inferno/pull/367)
+
 ### Security
 
 ## [0.12.7] - 2026-07-05
--- a/src/flamegraph/mod.rs
+++ b/src/flamegraph/mod.rs
@@ -612,7 +612,7 @@ where
         let info = if frame.location.function.is_empty() && frame.location.depth == 0 {
             write!(buffer, "all ({} {}, 100%)", samples_txt, opt.count_name)
         } else {
-            let pct = (100 * samples) as f64 / (timemax as f64 * opt.factor);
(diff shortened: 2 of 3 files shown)


jonhoo (dismissed):
Nice, thanks! Would you mind also adding an entry to CHANGELOG.md?

jonhoo (approved):
Looks great, will release shortly, thanks!

